PR #4, opened 2025-07-12
Fix: Remove exec arguments like %U and %F from desktop files

xdenotte:


Ly-sec:
Sorry for the confusion, I won't be merging this.
https://quickshell.org/docs/types/Quickshell/DesktopAction/ < as you can see we can just use the inbuilt execute() function. I will push that in a bit. Thank you so much though <3

xdenotte:
Np that's what I thought, too.

PR #13, opened 2025-07-15
Fix buffer for Cava

xdenotte:
Cava is no longer stuck for me


ferrreo:
This breaks the visualizations, they are no longer doing what they are supposed to with these changes, I have a fix and some misc improvements in another PR so closing this one.

PR #20, opened 2025-07-16
Fix: Corners don't block mouse clicks

MrDowntempo:
Will allow mouse clicks through corners, so UI on the top and bottom of windows can still be clicked.


ferrreo:
Thanks! Closing this as I borrowed these changes for https://github.com/Ly-sec/Noctalia/pull/21 and then fixed the overview issue introduced by this PR.

PR #21, opened 2025-07-16
fix: fixes more stuff

ferrreo:


MrDowntempo:
Was the addition of the
```WlrLayershell.layer: WlrLayer.Background``` lines
necessary? I think that it might be redundant with ```aboveWindows: false``` since the docs about aboveWindows says ["Note: On Wayland this property corrosponds to
WlrLayershell.layer."](https://quickshell.org/docs/types/Quickshell/PanelWindow/) or conversely, perhaps the aboveWindows lines could be removed now.

ferrreo:
> Was the addition of the `WlrLayershell.layer: WlrLayer.Background` lines necessary? I think that it might be redundant with `aboveWindows: false` since the docs about aboveWindows says ["Note: On Wayland this property corrosponds to WlrLayershell.layer."](https://quickshell.org/docs/types/Quickshell/PanelWindow/) or conversely, perhaps the aboveWindows lines could be removed now.

Above windows is needed for when not under Wayland. Sure some features won't work outside of Wayland but technically the rest should xD

And it puts it in the bottom layer not background layer so we need the other line too in order to fix all possible runtime cases.

PR #31, opened 2025-07-19
Fix starting applications

xdenotte:


Ly-sec:
> Applications not starting anymore

Can't reproduce but I'll merge it, should be fine.

PR #33, opened 2025-07-20
Quick fixs

anasgets111:
For battery percentage being as 0 to 1 float (needed to x100)
and shellRoot vs shellDir deperication edits
fix audio device selector signal name: Duplicate signal name: invalid override of property change signal or superclass signal ?

### DIFF
--- a/Bar/Modules/AudioDeviceSelector.qml
+++ b/Bar/Modules/AudioDeviceSelector.qml
@@ -7,7 +7,7 @@ import qs.Settings
 
 PanelWithOverlay {
     id: ioSelector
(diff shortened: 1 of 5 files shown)


Ly-sec:
> For battery percentage being as 0 to 1 float (needed to x100) and shellRoot vs shellDir deperication edits fix audio device selector signal name: Duplicate signal name: invalid override of property change signal or superclass signal ?

` ERROR:   caused by @Bar/Modules/Volume.qml[87:9]: Cannot assign to non-existent property "onClosed"` I think you missed something there.

PR #36, opened 2025-07-20
fix: non-compliant to XDG Base directory specification

Aiz0:
Checks if `XDG_CONFIG_HOME` is unset or empty before defaulting to `~/.config`

https://specifications.freedesktop.org/basedir-spec/latest/#variables


Ly-sec:
Going to merge it tomorrow, totally didn't think about that.

PR #44, opened 2025-07-27
feat: Environment variables, taskbar, and IPC enhancements

JPratama7:
Configuration & Environment Support:
- Environment variable support: All settings can now be overridden using NOCTALIA_* environment variables
- Theme file path configuration: Customizable theme and settings file locations

Taskbar & Window Management
- Idle inhibitor: Prevent system sleep via IPC when needed

Notification Enhancements
- Popup visibility toggle: Show/hide notification popups via IPC while maintaining silent mode
- Right-click silence mode: Toggle notification silence with visual icon state indicators


README.md
```diff
@@ -153,14 +153,23 @@ To make the weather widget, wallpaper manager and record button work you will ha
...
+### Toggle Idle Inhibitor:
 
+```
+qs ipc call globalIPC toggleIdleInhibitor
+```
+
+### Toggle Fullscreen:
+
+```
+qs ipc call globalIPC toggleFullscreen
```
ferrreo, line 171: This doesn't exist atm?

Bar/Modules/Taskbar.qml
```diff
@@ -0,0 +1,185 @@
...
+import Quickshell.Wayland
+import Quickshell.Widgets
+import qs.Settings
+import qs.Components
+
+Item {
+    id: taskbar
+    width: runningAppsRow.width
+    height: Settings.settings.taskbarIconSize
+
```
ferrreo, line 13: Can we make the visibility of this a setting as well please? For people that want to keep it minimal

Ly-sec: > Can we make the visibility of this a setting as well please? For people that want to keep it minimal

Totally agree with that, also then we will have to put the ColumnLayout inside a ScrollView too else it will break the layout.

README.md
JPratama7: it already exist in wm

PR #46, opened 2025-07-28
feat: configurable binary path, change icon when charging

JPratama7:
Centralize binary path, also change battery icon for charging

### DIFF
--- a/Bar/Modules/Battery.qml
+++ b/Bar/Modules/Battery.qml
@@ -16,6 +16,19 @@ Item {
     // Choose icon based on charge and charging state
     function batteryIcon() {
         if (!show) return "";
+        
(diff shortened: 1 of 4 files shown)


Ly-sec:
> Centralize binary path, also change battery icon for charging

I don't really see a reason to centralize the binary path if I'm being honest.
The battery part should 100% be added however

Ly-sec:
After thinking about it some more and talking to @ferrreo I don't think we need to have a centralized path for the binaries given how niche that would be. If people already write their own stuff I don't think we need to provide a part for it in the settings given how niche that is.

JPratama7:
alright, lemme revert that

Ly-sec:
Merging it, will be in the next release (or just git pull/clone).

PR #47, opened 2025-07-29
Wallpaper improvements

quadbyte:
I'm not using SWWW as I don't really change my wallpaper that often.
This PR addresses a couple issues with the basic wallpaper/overview implementation when not using SWWW daemon.

- Fix: The wallpaper was only applied to the primary screen.
- Improve the look of the blurred backdrop by applying a slight tint over the image (based on theme) which makes things look better imo.
- Also tinted the lockscreen in the same way.

image

### DIFF
--- a/Widgets/Background.qml
+++ b/Widgets/Background.qml
@@ -6,28 +6,38 @@ import qs.Settings
 
 ShellRoot {
     property string wallpaperSource: WallpaperManager.currentWallpaper !== "" && !Settings.settings.useSWWW ? WallpaperManager.currentWallpaper : ""
-    PanelWindow {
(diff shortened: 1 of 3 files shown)


Ly-sec:
> I'm not using SWWW as I don't really change my wallpaper that often.
> This PR addresses a couple issues with the basic wallpaper/overview implementation when not using SWWW daemon.
> 
> - Fix: The wallpaper was only applied to the primary screen.
> - Improve the look of the blurred backdrop by applying a slight tint over the image (based on theme) which makes things look better imo.
> - Also tinted the lockscreen in the same way.
> 
> image
> 

Awesome work, looks really good ♥️

PR #49, opened 2025-07-29
feat: add suspend option in power menu

msdevpt:
Just add the suspend option in the power menu.


Ly-sec:
Going to merge it, if it creates issues for people, I'll revert it (given how annoying linux can be with suspend).

PR #53, opened 2025-07-30
Lockscreen: fix initial weather fetching 

### LINKED ISSUE #51: weather lockscreen doesn't follow weather settings
I change location in weather settings but lockscreen doesn't follow that settings

quadbyte:
This PR fixes #51 

Noticed the weather data shown on the lockscreen was wrong  (temperature was off by some degrees).
After investigation I saw that the initial fetching of weather data was always done with the default city name: "Dinslaken"

It seems the JSON settings are not loaded yet when the lockscreen component is initialized.
Added a Qt.callLater() so that the weather fetching is done once everything is properly loaded.

### DIFF
--- a/Widgets/LockScreen.qml
+++ b/Widgets/LockScreen.qml
@@ -27,9 +27,12 @@ WlSessionLock {
     property double currentTemp: 0
     locked: false
 
-    // On component completed, fetch weather data
+    // On component completed, request to fetch weather data, with a little delay.
+    // Without this delay the city name is not loaded yet.
     Component.onCompleted: {
-        fetchWeatherData();
+        Qt.callLater(function () {
+            fetchWeatherData();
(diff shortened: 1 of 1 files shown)


Ly-sec:
> This PR fixes #51
> 
> Noticed the weather data shown on the lockscreen was wrong (temperature was off by some degrees). After investigation I saw that the initial fetching of weather data was always done with the default city name: "Dinslaken"
> 
> It seems the JSON settings are not loaded yet when the lockscreen component is initialized. Added a Qt.callLater() so that the weather fetching is done once everything is properly loaded.

That was unlucky timing, I just pushed an update while you were creating the PR - could you pull the latest changes from main and resolve the conflicts?
Or do you want me to resolve the conflict real quick? Up to you

Issue #54, opened 2025-07-30
Volume:  pill no longer auto activates after using the mousewheel


ItsLemmy:
Steps to reproduce : 
- Hover the volume icon and use the mousewheel up/down to alter the volume : The pill opens and closes nicely
- Then use the media keys on you keyboard or rotary encoder : The volume is changed but the pill no longer opens and show the volume.

A restart of quickshell fixes it.

I tried to look into it, but did not find a fix yet.

Ly-sec:
Can't replicate, just scrolled on the Volume pill and after that used my volume knob. Displays just fine

PR #55, opened 2025-07-30
Fix: Volume pill no longer auto activates #54

ItsLemmy:
Not sure why it works for you Ly-sec, but this fixed it for me.


Ly-sec:
> Not sure why it works for you Ly-sec, but this fixed it for me.
No idea why it fixes it either but it's merged.

ItsLemmy:
Awesome, thanks!

PR #58, opened 2025-07-30
Overview: using MultiEffect blur instead of the FastBlur

quadbyte:


Ly-sec:
> _ No description provided. _

Oh thought I did it for overview too, guess I only did it for the Lockscreen, my bad.

Issue #66, opened 2025-08-02
The avator and thumbnails of wallpaper seem to be in low res.


Bzzz-fan:
The avator and wallpaper thumbnails seem to be low res in setting panel, while the resolution of the avator image shown on lock screen appears to be normal.

<img width="923" height="199" alt="Image" src="https://github.com/user-attachments/assets/dec52b8a-bc7c-4965-8e95-e2ef63b01835" />

<img width="917" height="627" alt="Image" src="https://github.com/user-attachments/assets/efaa3e1f-fbf7-455e-849a-3651905e64c0" />

Ly-sec:
I'll take a look at it later today :) didn't even notice it before, thanks for pointing it out!

ItsLemmy:
This Avatar should look much better now, can you confirm @Bzzz-fan

The wallpapers should be a bit better but they are still severely downscaled to avoid wasting too much RAM.

Bzzz-fan:
> This Avatar should look much better now, can you confirm [@Bzzz-fan](https://github.com/Bzzz-fan)
> 
> The wallpapers should be a bit better but they are still severely downscaled to avoid wasting too much RAM.

<img width="913" height="184" alt="Image" src="https://github.com/user-attachments/assets/697fdf66-dbf6-4eed-b1e0-d79c7c17accf" />
<img width="933" height="406" alt="Image" src="https://github.com/user-attachments/assets/9143aa5c-ed78-4c0c-af2a-f80110397854" />

It seems to be still kind of blurry for my avatar, and I notice that the avatar image even get a little worse on lockscreen..

Bzzz-fan:
I can confirm that https://github.com/Ly-sec/Noctalia/pull/82/commits/8bb7bbda1462a94a82a0ab304eca74b3d64d0390 fixes the blurriness issue.

PR #73, opened 2025-08-03
Improve images resizing quality accross the board

quadbyte:
- Using mipmap: true and smooth: true for images that are downscaled (avatar, icons, ...)
- Using mipmap: false and smooth: true for images that are potentially upscaled (background, overview, ...)
- Removed source.width and height override which leads to poor quality


ItsLemmy:
This should fix/improve issue #66

ferrreo:
We should maybe swap Image with IconImage for the ones that are actually app icons as is the "quickshell" way - still need the mipmap flags for those tho

PR #77, opened 2025-08-03
fix: exisiting support network

JPratama7:
detect existing early when shell startup, add refresh button on wifi panel, i also change button from connect into a icon

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 .qmlls.ini
 notification_history.json
+Programs
--- a/Widgets/Sidebar/Panel/WifiPanel.qml
+++ b/Widgets/Sidebar/Panel/WifiPanel.qml
@@ -30,7 +30,10 @@ Item {
         running: false
         command: ["nmcli", "-t", "-f", "SSID,SECURITY,SIGNAL,IN-USE", "device", "wifi", "list"]
         onRunningChanged: {
-            // Removed debug log
+            console.log("scanProcess.running changed: " + running);
+            // if (!running) {
+            //     console.log("scanProcess finished.");
+            // }
         }
         stdout: StdioCollector {
             onStreamFinished: {
@@ -399,10 +402,22 @@ Item {
                         Layout.fillWidth: true
                     }
                     Item { Layout.fillWidth: true }
+                    Spinner {
+                        id: refreshIndicator
+                        Layout.preferredWidth: 24
+                        Layout.preferredHeight: 24
+                        Layout.alignment: Qt.AlignVCenter
+                        visible: scanProcess.running
+                        running: scanProcess.running
+                        color: Theme.accentPrimary // Assuming Spinner supports color property
+                        size: 22 // Based on the existing Spinner usage
+                    }
                     IconButton {
+                        id: refreshButton
                         icon: "refresh"
                         onClicked: wifiLogic.refreshNetworks()
                     }
+
                     Rectangle {
                         implicitWidth: 36; implicitHeight: 36; radius: 18
                         color: closeButtonArea.containsMouse ? Theme.accentPrimary : "transparent"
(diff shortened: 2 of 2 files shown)


Ly-sec:
> detect existing early when shell startup, add refresh button on wifi panel, i also change button from connect into a icon

Just tested it, it seems like the disconnect button doesn't work...

JPratama7:
what quickshell version you using? currently i'm using this version
`quickshell 0.2.0, revision a5431dd02dc23d9ef1680e67777fed00fe5f7cda, distributed by: AUR (package: quickshell-git)`

Ly-sec:
> what quickshell version you using? currently i'm using this version `quickshell 0.2.0, revision a5431dd02dc23d9ef1680e67777fed00fe5f7cda, distributed by: AUR (package: quickshell-git)`

`quickshell 0.1.0, revision 4dad44757085a42423f758bf0177cebcd07b4a4a, distributed by: AUR (package: quickshell-git)`

Why am I on 0.1.0 wtf lemme update


//edit:
Updated and it still behaves the same

Ly-sec:
@JPratama7 I found the issue.
image
Look at this, the old setup had quickshell-... in their name, however the new one doesn't, that means it wont disconnect the old connection or delete it... uhm any neat ideas on how we could fix that?

JPratama7:
ahh i see, i remove quickshell prefix

Ly-sec:
> ahh i see, i remove quickshell prefix

Yeah, do we wanna add it back or do you have any better idea so we don't run into breaking changes?

JPratama7:
i remove quickshell prefix because it will get conflict if i'm using nm-applet and it will make duplicate when inspecting using nmcli

Ly-sec:
Ah I see... uhm I guess we could disconnect the quickshell- one and also the one without and then delete the quickshell-entry perhaps? Would force the people to enter the password once again though.

JPratama7:
that a good idea, lemme try

Ly-sec:
> that a good idea, lemme try

Yeah please do so, we can then later remove the deletion of the quickshell- one, gonna update the readme so people know that they should reconnect once.

JPratama7:
@Ly-sec try this one, i change nmcli command to use ssid rather than profile id, so i don't need to modify nor delete anything

Ly-sec:
> @Ly-sec try this one, i change nmcli command to use ssid rather than profile id, so i don't need to modify nor delete anything

Just tested, sadly results in the same.

Perhaps we can just create a process:
```
Process {
...
    command: ["sh", "-c", "nmcli -t -f NAME connection show | grep '^quickshell-' | while read -r conn; do nmcli connection delete \"$conn\"; done"]
...
}
```

**Keep in mind the process was not tested yet, about to do it**

And then just run it in the `showAt` function, this would forcefully delete the connection which should be fine given we would just have to update the readme to let people know that this will happen once.
Then in a week or two we can just remove that process. Think that's the better call?

JPratama7:
i want to do that, but it will do delete without using user approval. should we use setting for that?

Ly-sec:
> i want to do that, but it will do delete without using user approval. should we use setting for that?

That's a good call actually... uhm maybe a popup or an info + button in the panel ooor something when you open the wifi panel and it detects that a `quickshell-` entry is present? I'm not sure

ferrreo:
Maybe add a check for quickshell- and pop a modal up asking if the user gives permission to move over to the new naming?

JPratama7:
hmmm idk how to do that, for now i just remove prefix and replace it, wdyt?

Ly-sec:
@ferrreo @JPratama7 I think I found a sleek fix:
`nmcli connection modify "quickshell-FRITZ_Box_7530_RG" connection.id "FRITZ_Box_7530_RG"
`

This literally just renames it, we could do that on the showAt function no?

JPratama7:
yup, try new commit @Ly-sec

Ly-sec:
> yup, try new commit @Ly-sec

Tried it, no wifi connections are loading. Also I get `WARN scene: @Widgets/Sidebar/Panel/WifiPanel.qml[158:-1]: TypeError: Type error`.
Also it doesn't rename it.


//edit:

```
        function replaceQuickshell(ssid: string): string {
            const newName = ssid.replace("quickshell-", "");
            
            if (!ssid.startsWith("quickshell-")) {
                return newName;
            }

            if (wifiLogic.networks && newName in wifiLogic.networks) {
                console.log(`Quickshell ${newName} already exists, deleting old profile`)
                deleteProfileProcess.connName = ssid;
                deleteProfileProcess.running = true;
            }

            console.log(`Changing from ${ssid} to ${newName}`)
            renameConnectionProcess.oldName = ssid;
            renameConnectionProcess.newName = newName;
            renameConnectionProcess.running = true;

            return newName;
        }
``` 

This works however we get an odd output in the terminal:
```
 DEBUG qml: Changing from quickshell-FRITZ_Box_7530_RG to FRITZ_Box_7530_RG
 DEBUG qml: Renamed connection 'quickshell-FRITZ_Box_7530_RG' to 'FRITZ_Box_7530_RG'
 ERROR qml: Error renaming connection 'quickshell-FRITZ_Box_7530_RG': 
```

PR #80, opened 2025-08-03
Remove qt5.comp dep

leiserfg:
- **Remove qt5.comp dependency**

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 .qmlls.ini
 notification_history.json
+Programs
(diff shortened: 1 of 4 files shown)


leiserfg:
The mask was incorrect cause now you can't use a Rectangle directly, now should be fine.

Ly-sec:
> The mask was incorrect cause now you can't use a Rectangle directly, now should be fine.

Okay that works, please pull the latest changes and fix the conflict, then I'll merge it :).

PR #81, opened 2025-08-03
Wallpaper refining

quadbyte:
- Swww: allow all images formats supported by Swww but only show the one we can render in the panel
- Fix: timer should be restarted when selecting a wallpaper
- Fix: avoid recomputing selected thumbnail by altering the border width
- Removed LazyLoader as it was not helping


leiserfg:
Isn't adding  https://github.com/KDE/kimageformats to the system enough to support extra formats?

PR #85, opened 2025-08-05
Added a way to choose on which screen the bar should be attached.

quadbyte:
- It's working fine and could save a lot of memory for users with multiple screen setup.
- There is no UI yet to edit this settings yet, but one can directly edit the list in ~/.config/Noctalia/settings.json to restrict the bar on the selected screen(s)
- Backward compatible for users without this settings (the bar will show on every screen)

I deciced to go with the QS naming convention by using "screen" instead of monitours, output (from Quickshell.screens)

### DIFF
--- a/Bar/Bar.qml
+++ b/Bar/Bar.qml
@@ -26,224 +26,234 @@ Scope {
         Variants {
             model: Quickshell.screens
 
-            Item {
+            Loader {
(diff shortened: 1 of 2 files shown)


ItsLemmy:
Closing this as we have a similar system in place on the dev branch.

PR #88, opened 2025-08-05
Improve Visibility

xdenotte:
I've adjusted some parameters to improve the visibility of the text on the battery icon on the lock screen, especially on bright wallpapers.
image
image


Ly-sec:
Pull the latest main branch and I'll merge it. Was just thinking about it though. What if we just create a rectangle with topRightRadius with backgroundPrimary, so that it would be always easily visible :>?

xdenotte:
like clock are made?

Ly-sec:
> like clock are made?

Yeah something like that perhaps? That should help with visibility

xdenotte:
Good idea, if it looks according to the design

ItsLemmy:
Thanks for the PR it was perfectly fine and justified however the lockscreen was fully redesigned, so this should no longer be an issue.

PR #91, opened 2025-08-06
Feat: Allow PgUp/PgDown in app menu

MrDowntempo:
Should be pretty basic. But I wouldn't be suprised if I hosed something up. The PageUp and PageDown were eaten by the text box, so I had to move the key detection stuff up a bracket.

### DIFF
--- a/Bar/Modules/Applauncher.qml
+++ b/Bar/Modules/Applauncher.qml
@@ -171,6 +171,14 @@ PanelWithOverlay {
                 if (filteredApps.length > 0)
                     selectedIndex = Math.max(selectedIndex - 1, 0);
             }
+            function selectNextPage() {
+                if (filteredApps.length > 0)
+                    selectedIndex = Math.min(selectedIndex + 10, filteredApps.length - 1);
+            }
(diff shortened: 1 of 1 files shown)


Bzzz-fan:
I suggest also include vim mode shortcuts <kbd>Ctrl</kbd>+<kbd>u</kbd> and <kbd>Ctrl</kbd>+<kbd>d</kbd> for pageup and pagedown for someone like me are used to vim keys. For example, I'm using <kbd>Ctrl</kbd>+<kbd>j</kbd> and <kbd>Ctrl</kbd>+<kbd>k</kbd> to select entries
```
Keys.onPressed: {
                                if (event.modifiers & Qt.ControlModifier) {
                                    switch (event.key) {
                                    case Qt.Key_J:
                                        root.selectNext()
                                        event.accepted = true
                                        break
                                    case Qt.Key_K:
                                        root.selectPrev()
                                        event.accepted = true
                                        break
                                    }
                                }
                            }
```

ItsLemmy:
merged this manually, since the code has changed a lot. Thanks!

PR #96, opened 2025-08-09
Fixed album image not appearing on bar 

wer-zen:
Removing layer stuff and replacing with ClippingRectangle


Ly-sec:
Well done :>!

PR #100, opened 2025-08-09
feat: idle inhibitor button

JPratama7:
add button to control idle inhibitor

### DIFF
--- a/Bar/Modules/IdleToggle.qml
+++ b/Bar/Modules/IdleToggle.qml
@@ -1,0 +1,67 @@
+import QtQuick
+import qs.Settings
+import qs.Components
+import Quickshell.Io
+
+Item {
+    id: idleToggle
+
+    // Provided by Bar
+    property var shell
+    property var screen: (typeof modelData !== 'undefined' ? modelData : null)
+
+    property bool inhibited: true
+
+    width: 22
+    height: 22
+
+    // Icon
+    Text {
+        id: iconText
+        anchors.centerIn: parent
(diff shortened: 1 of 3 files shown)


Bar/Modules/IdleToggle.qml
```diff
@@ -0,0 +1,67 @@
...
+        text: idleToggle.inhibited ? "bedtime" : "wb_sunny"
+        // Use filled vs outlined to indicate state
+        font.family: (mouseArea.containsMouse || idleToggle.inhibited) ? "Material Symbols Rounded" : "Material Symbols Outlined"
+        font.pixelSize: 16 * Theme.scale(screen)
+        color: idleToggle.inhibited ? Theme.accentPrimary : (mouseArea.containsMouse ? Theme.accentPrimary : Theme.textPrimary)
+    }
+
+    Process {
+        id: executor
+        command: ["qs", "ipc", "call", "globalIPC", "toggleIdleInhibitor"]
```
ferrreo: We shouldn't need to call out to a global IPC within the shell itself, just import and call the function

Bar/Modules/IdleToggle.qml
```diff
@@ -0,0 +1,67 @@
...
+import Quickshell.Io
+
+Item {
+    id: idleToggle
+
+    // Provided by Bar
+    property var shell
+    property var screen: (typeof modelData !== 'undefined' ? modelData : null)
+
+    property bool inhibited: true
```
ferrreo: defaulting to true? This should reflect that actual idle inhibit status at all times

PR #110, opened 2025-08-17
Added option for transparent background

MichaelThomas0721:
Currently there is no way of making the bar background transparent, this pr adds a new toggle option in the bar settings for a transparent background. When enabled, it adds `#66` to the beginning of the hex code which is 40% opacity.

### DIFF
--- a/Commons/Settings.qml
+++ b/Commons/Settings.qml
@@ -78,6 +78,7 @@ Singleton {
         property bool showBrightness: true
         property bool showNotificationsHistory: true
         property bool showTray: true
+        property bool transparentBackground: false 
         property list<string> monitors: []
       }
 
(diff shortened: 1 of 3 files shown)


ItsLemmy:
Good idea!

Could you please rework the PR : 
- Use a NSlider instead of NToggle.
- bool transparentBackground: false  => real backgroundOpacity: 1.0
- Apply the opacity on the color with Qt.rgba(Color.mSurface.r,  Color.mSurface.g,  Color.mSurface.b, Settings.data.bar.backgroundOpacity)

Thank you!

MichaelThomas0721:
Great suggestion!
I've changed it to a slider and real.
Let me know what you think!

PR #117, opened 2025-08-20
app icon fetching adjustments

ThatOneCalculator:
- Adds app icon to active window in menu bar
- Fixes the `getAppIcon()` method in the dock to be far more reliable

### DIFF
--- a/Commons/Settings.qml
+++ b/Commons/Settings.qml
@@ -111,6 +111,7 @@ Singleton {
       bar: JsonObject {
         property string position: "top" // Possible values: "top", "bottom"
         property bool showActiveWindow: true
+        property bool showActiveWindowIcon: true
         property bool showSystemInfo: false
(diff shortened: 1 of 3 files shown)


Ly-sec:
Hey, I just checked the commit. You're right the fix you did for the getAppIcon() is much more reliable. However looking at the active window icon, they're really blurry see the image.
image

If you can find a way somehow to make them less blurry I'd love to merge it (if you base it on the dev branch that is).
Perhaps you could also add a setting to disable/enable the active window icon.

ThatOneCalculator:
Gotcha. Please consider making the default branch the development branch, Github only forks the default branch unless manually specified otherwise.

ThatOneCalculator:
@Ly-sec added a toggle for showing the window icon! The icons don't look blurry on my end, so I'm not sure how to debug that.

PR #120, opened 2025-08-20
Change shebangs to the more portable /usr/bin/env bash

ErinvanderVeen:
This resolves issues on systems where bash is not in /bin/bash (e.g. NixOS).

This PR also adds the missing newlines at the end of these files.


ItsLemmy:
Thanks!

PR #125, opened 2025-08-22
Archlinux updater widget

anasgets111:
# Arch Updater Feature

- Adds an **Arch Linux updater button** to the system bar  
- **Requires**: `pacman-contrib` package  
- **Uses**: `xdg-terminal-exec` (currently hardcoded; can be customized manually by editing)  
- A **Title** is set for the update window (can be used to float it and such)  

## Functionality
- Checks/polls updates via `checkupdates` at intervals  
- Handles notifications and errors  
- Initiates system update  
- Widget displays:
  - Update status  
  - Tooltips with package details 
  - Trigger update when availble or `checkupdates` manually  

## Settings
- New `showArchUpdater` Toggle available in the **bar settings panel** to enable/disable feature 

image

### DIFF
--- a/Commons/Settings.qml
+++ b/Commons/Settings.qml
@@ -123,7 +123,6 @@ Singleton {
         property bool showActiveWindowIcon: true
         property bool alwaysShowBatteryPercentage: false
         property real backgroundOpacity: 1.0
-        property bool showArchUpdater: true
         property list<string> monitors: []
 
         // Widget configuration for modular bar system
--- a/Commons/Settings.qml
+++ b/Commons/Settings.qml
@@ -123,6 +123,7 @@ Singleton {
         property bool showActiveWindowIcon: true
         property bool alwaysShowBatteryPercentage: false
         property real backgroundOpacity: 1.0
+        property bool showArchUpdater: true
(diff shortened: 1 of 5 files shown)


Ly-sec:
As you can see we updated our way of adding Widgets to the bar. If you want, you can gladly check the Components/WidgetsLoader.qml to see how we do it now. I'd be glad to merge it when you update your PR.

anasgets111:
I think it's ready to be reviewed again, not sure how I can make load in the widgets list conditionally if it's arch based, if that's a must 
but I handled it in the UI for now

Ly-sec:
Hey, just tested it and it seems like there is a small issue.
As you can see in the image, `checkupdates` does show available updates, however the Widget does state that "No updates available".
image

anasgets111:
ya working in 2 quickshell repos and copying things from one to another , trying it myself and it's working let me know if there any more issues 
image

Ly-sec:
Yup seems to work now, probably an issue with the repo still being down sometimes
image

Ly-sec:
Wait shouldn't it also update or am I missing something @anasgets111 ?

PR #129, opened 2025-08-22
fix: use heuristicLookup for desktop entries if available

ThatOneCalculator:
ref: https://github.com/quickshell-mirror/quickshell/issues/192

Before:

image

After:

image


ThatOneCalculator:
Hmm... gonna draft this for now

image

ThatOneCalculator:
@Lemmy-Cook @Ly-sec seems to work as expected now! I think that was just a one-off gltich

PR #136, opened 2025-08-23
Added setting for workspace names

MichaelThomas0721:
New option in settings for adding the workspace names onto the workspace indicator on the bar.
I use Hyprland but I installed Niri to see how workspaces are handled there and as far as I can tell, they don't have names, correct me if I am wrong though.

### DIFF
--- a/Commons/Settings.qml
+++ b/Commons/Settings.qml
@@ -122,6 +122,7 @@ Singleton {
         property bool showActiveWindowIcon: true
         property bool alwaysShowBatteryPercentage: false
         property real backgroundOpacity: 1.0
+        property bool showWorkspaceNames: false
         property list<string> monitors: []
 
         // Widget configuration for modular bar system
--- a/Modules/Bar/Widgets/Workspace.qml
+++ b/Modules/Bar/Widgets/Workspace.qml
@@ -161,6 +161,21 @@ Item {
         Rectangle {
           id: workspacePill
           anchors.fill: parent
+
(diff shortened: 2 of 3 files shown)


anasgets111:
Niri has named workspaces, they are the ones that presist, other wise they are dynamically created

for example here's mine
```
 niri msg workspaces
Output "DP-2":
   1 "Telegram"
 * 2 "Discord"
   3 "Slack"
   4

Output "DP-3":
 * 1 "Zen"
   2 "Chromium"
   3 "Code"
   4 "Zed"
   5 "Files"
   6 "Thunderbird"
   7 "Games"
   8 "Terminal"
   9
   ```

MichaelThomas0721:
@anasgets111 Thank you for that comment, I was able to use your niri dots to test this and the names are working.
They are limited to two characters since more than that won't fit.
I did two characters instead of one to make it less likely that the workspace names overlap.

ItsLemmy:
Could you add a few screenshots of how it looks, we are kindda doubtful about this one.   We tried before having numbered workspaces and even with one character is was not super practical.

MichaelThomas0721:
Ya for sure. With two characters it's not the prettiest thing but I think it's very practical, plus it doesn't have to be turned on.
Let me know what you think!
image
image
image
image
image
image

ItsLemmy:
Thanks, I'll probably tweak it a little bit more to see if I can improve the look.

PR #141, opened 2025-08-25
adding a flake.nix file to build this shell with nix

NotWinder:
Hello i have wrote a flake file for building this shell as a binary.
this would be useful for nixos users so they can add the flake of the repo into their configs and add the package of the flake input to their system environment

it would also be useful for the people on other distros who use the nix package manager(with flakes enabled)

i have taken the flake of the caelestia-shell project as a base line and changed a few things (like using the systems flake rather then what that was using)

i have tested the binary and it seems that everything is working as expected
here is a list of what i have tested
- the workspace indicator
- the media player (cava isn't working but i think thats a problem from my system because when i run cava in a terminal it doesn't work there either)
- wallpaper changing (both with sww enabled and disabled)
- video recording
- notifications 
- tray
- system monitor indicators (cpu, gpu, ram, storege)

the only other thing that isn't working for me (other then cava)
is bluetooth  and tab does not show the paird devises but it does scan for new ones but even then it only shows the mac address

which also was happening when i installed the dependencies and cloned the repo from github and ran it directly
(i do have a usb bluetooth dongle so that might be the thing that is causing that)

i don't have wifi so i couldn't try that

but overall i couldn't find a diffrence between running if via flake or cloning the repo and running quickshell directly

something to take note of is that since the binary is just a wrapped version of quickshell
the ipc calls DO work BUT the should use the ```noctolia-shell``` command instead of the ```qs``` command 
(so to run the shell it would be ```noctalia-shell``` and to open the launcher it would be ```noctalia-shell ipc call launcher toggle```)

### DIFF
--- a/nix/default.nix
+++ b/nix/default.nix
@@ -4,29 +4,23 @@
   stdenv,
   makeWrapper,
   makeFontsConf,
-  fish,
   ddcutil,
   brightnessctl,
   cava,
   networkmanager,
-  lm_sensors,
   wl-clipboard,
-  libqalculate,
-  inotify-tools,
+  libnotify,
   bluez,
   bash,
-  hyprland,
   coreutils,
   findutils,
   file,
   material-symbols,
-  roboto-serif,
-  roboto-slab,
-  roboto-mono,
-  roboto-flex,
+  roboto,
   inter-nerdfont,
   matugen,
   cliphist,
+  swww,
   gpu-screen-recorder,
   gcc,
   qt6,
@@ -36,34 +30,28 @@
 }: let
   runtimeDeps =
     [
-      fish
-      ddcutil
+      bash
+      bluez
       brightnessctl
       cava
-      networkmanager
-      lm_sensors
-      wl-clipboard
-      libqalculate
-      matugen
       cliphist
-      gpu-screen-recorder
-      inotify-tools
-      bluez
-      bash
-      hyprland
       coreutils
-      findutils
+      ddcutil
       file
+      findutils
+      gpu-screen-recorder
+      libnotify
+      matugen
+      networkmanager
+      swww
+      wl-clipboard
     ]
     ++ extraRuntimeDeps;
 
   fontconfig = makeFontsConf {
     fontDirectories = [
       material-symbols
-      roboto-serif
-      roboto-slab
-      roboto-mono
-      roboto-flex
+      roboto
       inter-nerdfont
     ];
   };


NotWinder:
since i wouldn't call my self even an intermediate in nix, if anyone has any improvements or suggestions i would be happy to know.
and if i have missed any dependencies please let me know.

Ly-sec:
> since i wouldn't call my self even an intermediate in nix, if anyone has any improvements or suggestions i would be happy to know. and if i have missed any dependencies please let me know.

There is a lot of dependencies in the `default.nix` that aren't needed, same for some fonts etc. Also I don't think we should be installing `hyprland` at all because people presumably either have hyprland or niri installed when looking at Noctalia.

NotWinder:
thank you for the comment
i did remove the unnecessary dependencies 

i also check in a vm to see if there are any missing fonts and there were none
i also forgot to add sww in the dependencies and why it was working was because i had swww installed but when i tried it on the vm i figured that it was not in the dependencies 
which i have added now

Ly-sec:
Just making sure before I merge it, you fully tested it. No weird edge cases or anything that you encountered? Sadly can't test it currently since I'm not on Nix right now.

NotWinder:
i made a test nixos vm with just hyprland and kitty and this flake as the packages 
i'll put some photos and media in the vm to check the media management and make sure that the wallpapers will work and i'll report back

NotWinder:
ok so i have check in a vm
- the media worked along (with the visualizer) but the cover art wasn't showing up (but that might be because i was using cmus to play the song, since it is working on my main system which is using mpd)
- the wallpaper without swww worked without issues 
- swww did work when i enabled i but it did NOT work when i re-ran the shell and showed an error that swww-daemon is not running(it did work again when i disable swww and re-enable it) it might be because it's in a vm?
- the weather stuff worked without issues
- the power selector was showing a warning about the service which was gone after i enabled the reletive services in the nixos config (which i don't think can be enabled by the flake, at least in the current form which is only a package) but it didn't let me select the profile for some reason (maybe a vm issue)

- the only thing that didn't work on the vm (with no errors) was the screen recorder. the logs show that it is recording and the buttons light up correctly but it did not save the video (even though in the logs it says that it did save it there) and it only makes a directory (which again i think it's a vm issue because i does save on my main system)

to summarize i think the flake is fine because the things that was not working on my main system are working in the vm (like the audio visualizer)
and the things that are not working on the vm are working on my main system(swww, the screen recording and the power-profile issues are only in the vm)
and i didn't find any missing fonts or any issues that are present both in the vm and the main system

NotWinder:
besides the bluetooth and wifi list not showing up 
but i don't think that's related to the flake 
since there are no errors in the logs indicating any missing dependencies or wrong settings

NotWinder:
and also there is an error saying that ```Command: QList("checkupdates")``` is not found 
which i'm guessing that is related to the arch update module in the shell and if it is that i don't think it is needed for nixos
unless someone uses nix package-manager on arch

Ly-sec:
Okay I'll merge it for now and do some more testing the coming days and see if everything properly works.

PR #147, opened 2025-08-26
Nix Part for README

wer-zen:
Added nix part and fixed keybindings part


wer-zen:
can somebody test it and see if it works on their machine? I know nix doesn't have the _it  works on my machine_ problem but I still want to make sure

Ly-sec:
Thank you very much :heart:

wer-zen:
No problem <3

PR #152, opened 2025-08-26
small flake.nix fix - providing a proper 'default' and fixing formatter errors

Drazzy9295:
There seems to be a small discrepancy between the NixOS install guide and what the flake.nix actually provides. Nothing major, but there are two solutions:

1. Change the "quickshell.packages.${system}.default" on step 2 to "inputs.noctalia.defaultPackage.${pkgs.system}"
2. Add a proper 'default' to the Flake.nix

Both options work just as well, but I think the latter is the 'proper' convention. I've gone ahead with doing the second solution on this pull request instead of changing the install guide.

I was also getting errors with how the 'formatter' was setup, so I sorted that too. 

**Side note:** I'm not sure if you have a preferred formatting for the code, so sorry ahead of time on that.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -11,57 +11,83 @@
     };
   };
 
-  outputs = { self, nixpkgs, systems, quickshell, ... }:
-    let
-      eachSystem = nixpkgs.lib.genAttrs (import systems);
-    in {
-      formatter = eachSystem (pkgs: pkgs.alejandra);
+  outputs = {
+    self,
+    nixpkgs,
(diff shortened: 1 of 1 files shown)


NotWinder:
it seems good to me 
i have initially made a nix directory for the build module to live in 
cuz i was planning on making a home-manager module if everything was working as expected 

but now all the build steps live in the flake.nix 
which is fine but it probebly will be subject to change if there is planes on making a home-manger module.

PR #158, opened 2025-08-27
Added kitty matugen template

MichaelThomas0721:
Based off the previous version but I changed some that I found had too low contrast compared to the background. I also changed it so existing terminals will reload the color theme using the post_hook.
I've attached some screenshots with different color wallpapers. Let me know what you think.
image
image
image
image


Ly-sec:
I'll try it later with a more basic shell just to make sure everything is fine and then I'll merge it :). Very much appreciate it ♥️

Ly-sec:
Looks awesome! Merging it, thank you very much :heart:

PR #161, opened 2025-08-28
Added ghostty matugen template

MichaelThomas0721:
Tried to match the kitty terminal theme as closely as possible.
For this comment in discussion https://github.com/noctalia-dev/noctalia-shell/discussions/149#discussioncomment-14237449

Ghostty doesn't really support live reloading so that isn't included. See this discussion on ghostty's github https://github.com/ghostty-org/ghostty/discussions/3643#discussion-7751140


Ly-sec:
Instead of creating `config-noctalia` you should be doing something like:
`lines.push('output_path = "~/.config/ghostty/themes/noctalia"')` that way people can find it with `ghostty +list-themes
` also perhaps you could find a way to add a post_hook that replaces `theme = theme-name` inside the `config` with `theme = noctalia`, this way nobody would have to do anything but restart ghostty.

MichaelThomas0721:
I've changed the output location and added a post_hook that will add it to the ghostty conf file.
The command takes into account the theme line already existing, not already existing, formated "theme =" or formatted "theme=".

Ly-sec:
Works perfectly fine :) merging it. Thank you very much!

PR #164, opened 2025-08-28
brightness: avoid DDC on internal panels, add timeouts, auto-blacklis…

lesi-nedo:
Title: brightness: avoid DDC on internal panels, add timeouts, auto-blacklist bad DDC buses

Summary
- Prevent DDC on internal panels (eDP/LVDS/DSI).
- Add timeouts to ddcutil detect/get/set to avoid UI hangs.
- Auto-blacklist DDC bus when VCP 10 returns ERR and fall back to internal.

Rationale
ddcutil on certain I2C buses (e.g., /dev/i2c-6) retries and blocks, causing Quickshell to become unresponsive. Skipping DDC for internal panels and timing out ddcutil calls prevents this. Auto-blacklist avoids repeated stalls.

Changes
- New root.ddcBlacklist.
- method selection prefers internal for eDP/LVDS/DSI or blacklisted buses.
- timeout wrappers: detect (3s), get/set (1s), safe fallbacks.
- Re-init after blacklisting to switch method.

Logs
- Example:
  - Raw: `VCP 10 ERR` on eDP-1
  - ddcutil journal shows repeated EDID checks on /dev/i2c-6.

Testing
- Arch Linux, Quickshell current.
- eDP-1 now uses internal backlight; no UI stalls.
- External monitors with valid DDC unaffected.
- Buses with ERR auto-blacklisted once, then stable.

Risk
- Minimal; only affects DDC path and internal-panel detection via name prefix.

### DIFF
--- a/Services/BrightnessService.qml
+++ b/Services/BrightnessService.qml
@@ -6,214 +6,233 @@ import Quickshell.Io
 import qs.Commons
 
 Singleton {
-  id: root
+    id: root
 
-  property list<var> ddcMonitors: []
-  readonly property list<Monitor> monitors: variants.instances
-  property bool appleDisplayPresent: false
+    property list<var> ddcMonitors: []
+    readonly property list<Monitor> monitors: variants.instances
+    property bool appleDisplayPresent: false
+    // Blacklist DDC buses that error or hang
+    property var ddcBlacklist: []
 
-  function getMonitorForScreen(screen: ShellScreen): var {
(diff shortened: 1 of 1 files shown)


Ly-sec:
Hey there, I just tested it with my external display (DDC). It seems to still work in terms of changing the brightness however it shows 0% on every quickshell restart (on the Widget and also inside the BrightnessTab).

lesi-nedo:
Understood, I will try to figure out how to fix

PR #174, opened 2025-08-30
feat: app taskbar

JPratama7:


ItsLemmy:
Thanks for your PR, I have a few comments to keep the code base more homogenous.

In _Modules/Bar/Widgets/Taskbar.qml_

Line 12-13 should be exactly as other bar widgets. ie: 
```   
  property ShellScreen screen
  property real scaling: ScalingService.scale(screen)
```

Line 14 should use Style.baseWidgetSize (which equals to 32) 

Line 47-53 could be shortened in a single line:
```
taskbarItem.isActive ? Color.mPrimary : root.color
```

When logging an error (in your Try / Catch) you should user Logger.error instead of Logger.log

Tooltip should use NTooltip.

Thanks again!

JPratama7:
Alrighttt

ItsLemmy:
Thanks!

PR #176, opened 2025-08-31
Prevent notebook panel freezes when DDC/CI is not supported

lesi-nedo:
Problem
- Issue: “ddcutil freezes the notebook monitor if is not supports ddc/ci”.
- Root cause: We could end up issuing DDC VCP commands for displays that don’t actually support DDC/CI, which can hang some eDP laptop panels.

Solution
- Only ever run ddcutil get/set VCP on monitors positively identified as DDC/CI-capable.
- Use full ddcutil detect output to determine support and bus, then filter out non‑DDC panels.

What changed
- DDC detection:
  - Switched to ddcutil detect --sleep-multiplier=0.5.
  - Parse “Model:” and “I2C bus: /dev/i2c-N”.
  - Detect “This monitor does not support DDC/CI” and set isDdc accordingly.
  - root.ddcMonitors now contains only isDdc === true entries (shape: { model, busNum, isDdc }).
  - Added per‑display Logger.log line to verify detection.
- Monitor logic:
  - isDdc/busNum are derived from the filtered list; initBrightness/setBrightness now call ddcutil only for confirmed DDC monitors.
- Performance: Reduced detect latency with a lower sleep multiplier.

Why it fixes the freeze
- Non‑DDC panels (typical laptop eDP) are excluded from root.ddcMonitors, so no ddcutil getvcp/setvcp calls are made against them, avoiding the hang.

Compatibility notes
- The stored monitor entries now include isDdc (non‑breaking).
- Model matching uses the “Model:” field from ddcutil; validate that it matches ShellScreen.model on your hardware.

Testing
- On a notebook with a non‑DDC internal panel:
  - Verify logs show the panel as “is DDC: false”.
  - Confirm no ddcutil getvcp/setvcp invocations occur for the internal panel and no freeze happens.
- With external DDC monitors:
  - Ensure they are logged as “is DDC: true” and brightness get/set works as before.
- With Apple Studio Display:
  - appleDisplayPresent should be true; asdbctl get/set should continue to work.

### DIFF
--- a/Services/BrightnessService.qml
+++ b/Services/BrightnessService.qml
@@ -6,214 +6,233 @@ import Quickshell.Io
 import qs.Commons
 
 Singleton {
-  id: root
+    id: root
 
-  property list<var> ddcMonitors: []
-  readonly property list<Monitor> monitors: variants.instances
-  property bool appleDisplayPresent: false
+    property list<var> ddcMonitors: []
+    readonly property list<Monitor> monitors: variants.instances
+    property bool appleDisplayPresent: false
+    // Blacklist DDC buses that error or hang
+    property var ddcBlacklist: []
 
-  function getMonitorForScreen(screen: ShellScreen): var {
-    return monitors.find(m => m.modelData === screen)
-  }
-
-  function getAvailableMethods(): list<string> {
-    var methods = []
-    if (monitors.some(m => m.isDdc))
-      methods.push("ddcutil")
-    if (monitors.some(m => !m.isDdc))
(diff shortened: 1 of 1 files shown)


Ly-sec:
Just tested it, seems to work fine. Is there anything you still want to add/change or shall I merge it?

lesi-nedo:
For now nothing, however in the future I might have some more code. This change is important, at least for my case. You can see the problem here: https://nedo.im/media/videos/freezing.webm

Issue #165, opened 2025-08-29
[Bug]: Disconnect devices in bluetooth panel not working, no remove device option for paired devices


Bzzz-fan:
### Description
All `Disconnect` buttons in bluetooth panel cannot be pressed, and the operation for paired but not connecting devices should be `Remove` or `Forget`.

### Steps to Reproduce
1. Pair some bluetooth device
2. Toggle bluetooth panel
3. Connect or disconnet this device from terminal, try to press the `Disconnect` button

### Expected Behavior
The `Disconnect` button should work for currently connecting devices and maybe `Remove` or `Forget` option for paired but not connecting devices.

### Screenshots

<img width="784" height="1011" alt="Image" src="https://github.com/user-attachments/assets/ff837a52-d84f-4548-ada8-05f0e508531a" />

### Environment
- Distro: Arch
- Compositor: Niri
- Version: `main`

ItsLemmy:
Can you try the latest changes ?  thank you

PR #179, opened 2025-08-31
Fix to issue #165

lesi-nedo:
Here’s why this should be merged and what was implemented to fix the issue.

Problem and impact
- Symptoms: “Disconnect” could not be triggered in the Bluetooth panel; paired-but-not-connected devices had no clear Remove/Forget action.
- Root causes:
  - Action eligibility was tied to the wrong state (paired vs connected), so UI states didn’t match what the device could actually do.

Key fixes in the service
- Corrected connection logic:
  - canConnect now uses !device.connected (not !device.paired).
  - canDisconnect now uses device.connected.
- Busy-state guard:
  - isDeviceBusy returns true for pairing/connecting/disconnecting to prevent conflicting actions.
- Added actionable helpers:
  - forgetDevice: clears trusted and calls forget.

Key fixes in the UI
- Unified, reliable click handling:
  - Whole device row is clickable (MouseArea anchors.fill: parent).
  - Left-click: connect or disconnect; right-click: forget.
- Clear device grouping and affordance:
  - Connected devices, Known devices (paired/trusted, not connected), Available devices.
  - Tooltip on Known devices: “Left click to connect, right click to forget.”
- Accurate visual states:
  - Dynamic label (“Connect”/“Disconnect”/“Pairing…”/“Blocked”).
  - Busy cursor while pairing/connecting/disconnecting.
  - Safer actions hidden/disabled while busy.

Outcome
- “Disconnect” works consistently, even when connection state changes outside the panel (e.g., via terminal).
- Paired-but-not-connected devices have an explicit Remove/Forget action via right-click.
- UX is clearer, and accidental/invalid actions are prevented during busy states.

Why merge
- Resolves a user-facing regression with minimal, well-scoped changes.
- Aligns behavior with expected semantics (connected vs paired).
- Improves reliability and discoverability without altering core Bluetooth APIs.

If you want, I can add a short note to the changelog describing the behavior changes.

### DIFF
--- a/Assets/Matugen/Matugen.qml
+++ b/Assets/Matugen/Matugen.qml
@@ -67,6 +67,11 @@ Singleton {
       lines.push(
             'post_hook = "sed -i /themes/d ~/.config/fuzzel/fuzzel.ini && echo include=~/.config/fuzzel/themes/noctalia >> ~/.config/fuzzel/fuzzel.ini"')
     }
+    if (Settings.data.matugen.vesktop) {
+      lines.push("\n[templates.vesktop]")
+      lines.push('input_path = "' + Quickshell.shellDir + '/Assets/Matugen/templates/vesktop.css"')
+      lines.push('output_path = "~/.config/vesktop/themes/noctalia.theme.css"')
+    }
 
     return lines.join("\n") + "\n"
   }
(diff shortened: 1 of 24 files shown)


ItsLemmy:
Good job!
I did not get any decent BT device to test all this,  so I was coding blindly!
I ordered a pair of cheap BT earbuds a couples day ago but I'm still waiting for the delivery.

I'm gonna merge this even if I can't really test it properly.
Thanks!

PR #182, opened 2025-09-01
disable test mode for battery widget.

nollidnosnhoj:
Assuming test mode is used to test out the battery widget when the machine has not battery; however, it was left enabled :D (I was confused why my laptop battery was low when it has been changing xD)


ItsLemmy:
My bad, thanks for the fix!

PR #189, opened 2025-09-01
Added a toggle for Noctalia-shell wallpaper management

lonerOrz:
- Added a toggle for Noctalia-shell wallpaper management
- fix shader error


ItsLemmy:
Thank you for you contribution!

PR #224, opened 2025-09-05
make default notification action text "Open"

ThatOneCalculator:
99% of times an action on a notification will be opening the related content


Ly-sec:
Sounds good to me, merged :).

PR #225, opened 2025-09-06
fix: align KeyboardLayout widget with other bar components

lonerOrz:
图片


ItsLemmy:
Good catch, did not see the issue at scale 1.0.  Thanks for your contribution

PR #229, opened 2025-09-06
Fix for fingerprint flow on lock screen

matejc:
Better fingerprint integration by removing the check for empty password.

With this change, fingerprint works the same way as everywhere else for me on NixOS:

In NixOS configuration:
```
services.fprintd.enable = true;
```

And in the lock screen:
1. Press Enter
2. "Authenticating..." is shown
3. In this moment scan your finger on fingerprint scanner
4. Screen unlocks

Note:
Without this change, it will ask for the password first, and since the password for `login` in PAM is always enabled on NixOS, you need to put something in, and obviously it has to be a correct password.

### DIFF
--- a/Modules/LockScreen/LockContext.qml
+++ b/Modules/LockScreen/LockContext.qml
@@ -28,12 +28,6 @@ Scope {
       return
     }
 
-    if (currentText === "") {
-      errorMessage = "Password required"
-      showFailure = true
-      return
-    }
-
     root.unlockInProgress = true
     errorMessage = ""
     showFailure = false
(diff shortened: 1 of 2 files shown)


ItsLemmy:
I'm willing to try this, but for some reason my finger print reader on my old Arch laptop is acting up.  I can enroll a finger succesfully, but can never actual unlock with my finger.   I'll persist!

matejc:
> I'm willing to try this, but for some reason my finger print reader on my old Arch laptop is acting up.  I can enroll a finger succesfully, but can never actual unlock with my finger.   I'll persist!
> 

Hm, from the top of my head I can think of two reasons:
1. Did you use sudo or something like that to enroll? As you have to use the same user to enroll as for unlock
2. Does your /etc/pam.d/login has pam fprint auth module line?

But for both of these options will always ask for password imediately and not initialize or wait for fingerprint input

Edit:
You could also check if your fprintd systemd service is running and what is in its logs.

ItsLemmy:
Thanks!

PR #231, opened 2025-09-07
Side Panel Bottom was a lil too small

MrDowntempo:
I run with no scaling. My SidePanel's margins at the top is 15px and between cards is 15px or 16px depending. Kinda odd, inconsistency, but I didn't fool with that. The the bottom was only 11px. So I increased the size of the panelHeight by 4 pixels so it matches the spacing on the top.

Side by Side comparison

image

### DIFF
--- a/Modules/SidePanel/SidePanel.qml
+++ b/Modules/SidePanel/SidePanel.qml
@@ -11,7 +11,7 @@ NPanel {
   id: root
 
   panelWidth: 460 * scaling
-  panelHeight: 708 * scaling
+  panelHeight: 712 * scaling
 
   panelContent: Item {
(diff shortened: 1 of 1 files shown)


ItsLemmy:
I believe I fixed that yesterday with a better approach which auto size (even tho we have a forced initial size).  Can you test the main branch ?

MrDowntempo:
> I believe I fixed that yesterday with a better approach which auto size (even tho we have a forced initial size). Can you test the main branch ?

Yeah, you pushed your fix literally minutes after mine. I commented about it jokingly on Discord but I guess you weren't around. Your fix is far superior and has resolved my issue.

PR #241, opened 2025-09-09
Fix: Incorrect 0% battery warning at startup

lonerOrz:
At application startup, the low battery warning is incorrectly shown as 0%, even when the actual battery level is higher


Ly-sec:
Looks good to me, I'll merge it. Thank you very much :heart:

PR #243, opened 2025-09-10
fix for PowerPanel Shortcut invalid

juvevood:
fixed Shortcut Escape for PowerPanel invalid.


Ly-sec:
I'm a bit puzzled, given `... ipc call powerPanel toggle` works just fine?

Ly-sec:
Merged, I didn't properly read through it.

PR #244, opened 2025-09-10
Screen Corners use general radius ratio of settings

juvevood:
Screen Corners use general radius ratio from settings

### DIFF
--- a/Commons/Settings.qml
+++ b/Commons/Settings.qml
@@ -317,6 +317,7 @@ Singleton {
         property bool dimDesktop: false
         property bool showScreenCorners: false
         property real radiusRatio: 1.0
+        property real screenRadiusRatio: 1.0
         // Animation speed multiplier (0.1x - 2.0x)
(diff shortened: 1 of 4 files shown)


ThatOneCalculator:
I think they should be separate, I believe the setting is in place to account for the rounding of laptop displays.

juvevood:
> I think they should be separate, I believe the setting is in place to account for the rounding of laptop displays.

border of UI elements should be with screen inner corner

ThatOneCalculator:
That's your personal opinion and doesn't fit for every screen or setup

juvevood:
> That's your personal opinion and doesn't fit for every screen or setup

how about add a separate configuration item for edge of screen?
2025-09-11T09:31:15,581766867+08:00

ThatOneCalculator:
I think that's good

ItsLemmy:
Thank you for your contribution!

PR #250, opened 2025-09-11
fix(bluetooth): rename wifiSwitch to bluetoothSwitch

matejc:
rename wifiSwitch to bluetoothSwitch to fix the toggle switch

not sure how renaming the id fixes the toggle switch, but it works now - I am very new to qml and qs


ItsLemmy:
Weirdly thoses ids are not used anywhere, in the code.   Merging anyway to keep the codebase clean. Thank you!

PR #252, opened 2025-09-11
Update nix readme instructions

BinaryQuantumSoul:
- Quickshell was stated as input for quickshell instead of noctalia
- Add emphasis on usage of nixpkgs unstable (won't compile otherwise)


ItsLemmy:
Thanks a lot!

PR #253, opened 2025-09-11
import gpu-screen-recorder only on x86_64 in nix flake

### LINKED ISSUE #251: [Bug]:  nix(aarch64): Noctalia fails to build due to hard dependency on gpu-screen-recorder
### Description

Building Noctalia via Nix on aarch64 fails at evaluation because gpu-screen-recorder is hard-required and is restricted to x86_64-linux. This breaks the Nix install on ARM (e.g., Asahi Linux).

### Steps to Reproduce

1. On an aarch64 machine (e.g., Asahi Linux), run:
2. nix build github:noctalia-dev/noctalia-shell
    - or add as a flake input and run: sudo nixos-rebuild switch --flake .#<host>
3. See the error:
    - error: Package ‘gpu-screen-recorder-5.6.5’ … is not available on the
requested hostPlatform:
     hostPlatform.config = "aarch64-unknown-linux-gnu"
     package.meta.platforms = [ "x86_64-linux" ]
     … refusing to evaluate.

### Expected Behavior

Noctalia should at least evaluate on aarch64-linux and ideally build/run.
gpu-screen-recorder should be optional or platform-gated so ARM users can install and use Noctalia without it.

### Screenshots

N/A (build-time evaluation failure).

### Environment

- Distro: NixOS (Asahi Linux) on aarch64
- Compositor: Niri
- Noctalia-shell Version: flake HEAD (nixos-unstable input)

### Additional Context

- Root cause: runtimeDeps in the flake unconditionally includes
gpu-screen-recorder, which is x86_64-only in nixpkgs. Because this list is
evaluated before the derivation is constructed, downstream overlays cannot remove
it in time.
- Proposed fixes:
    - Conditionally include gpu-screen-recorder only on x86_64-linux (e.g.,
lib.optionals stdenv.isx86_64 [ gpu-screen-recorder ]).
    - Or expose a parameter (e.g., withGpuScreenRecorder ? stdenv.isx86_64) and
gate the dependency.
    - Alternatively, provide an ARM/minimal package variant that omits
gpu-screen-recorder.

SailorSnoW:
This PR now conditionally import the `gpu-screen-recorder` for the nix flake if the system architecture is x86_64 as the package is not available currently for aarch64 on nixpkgs.

Fix #251

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -11,81 +11,96 @@
     };
   };
 
-  outputs = {
-    self,
-    nixpkgs,
-    systems,
-    quickshell,
-    ...
-  }: let
-    eachSystem = nixpkgs.lib.genAttrs (import systems);
-  in {
-    formatter = eachSystem (
-      system:
-        nixpkgs.legacyPackages.${system}.alejandra
-    );
+  outputs =
+    {
+      self,
+      nixpkgs,
+      systems,
+      quickshell,
+      ...
+    }:
+    let
+      eachSystem = nixpkgs.lib.genAttrs (import systems);
+    in
+    {
+      formatter = eachSystem (system: nixpkgs.legacyPackages.${system}.alejandra);
 
-    packages = eachSystem (
-      system: let
-        pkgs = nixpkgs.legacyPackages.${system};
-        qs = quickshell.packages.${system}.default.override {
-          withX11 = false;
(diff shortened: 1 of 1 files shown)


ItsLemmy:
Thanks a lot!

PR #254, opened 2025-09-12
Fix a missing semicolon in Nix instructions.

RdrSeraphim:
Good ol' missing semicolon. :)


ItsLemmy:
Thanks, good catch :P

PR #255, opened 2025-09-12
Add Monochrome (Black & White) color scheme

MrDowntempo:
[I saw this comment](https://old.reddit.com/r/unixporn/comments/1ndsdq0/niri_noctalia/ndogzrv/) in the reddit thread and said "HEY! I CAN DO THAT!" I did run into an issue where the check in the Wallpaper selector and ColorScheme selector are assigned to different colors. I think they should be consistent, but I don't know which of the two would be the best to keep. In any event, I've tested this in both Dark and Light mode, and it seems to be pretty usable everywhere, nothing besides that check stood out to me, but after swapping one set of blacks and whites around, that nitpik doesn't affect this color scheme. There are a few instances when alpha transparency is used which results in some grays. I doubt that's a big enough issue to need special attention for this one niche theme, but thought it worth mentioning.

### DIFF
--- a/Assets/ColorScheme/Monochrome.json
+++ b/Assets/ColorScheme/Monochrome.json
@@ -1,0 +1,34 @@
+{
+  "dark": {
+    "mPrimary": "#ffffff",
+    "mOnPrimary": "#000000",
+    "mSecondary": "#ffffff",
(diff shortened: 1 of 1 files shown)


ItsLemmy:
Thank you!

PR #257, opened 2025-09-12
Add toggle to hide unoccupied workspaces

mkuritsu:
# Description

First of all I don't know if I should have open a Feature Request Issue before making this PR but since it was not a big change and I think I could make it myself I decided to just make it.

This adds an option to the workspace widget to not display workspaces that aren't currently occupided by windows. I also added some qmlls files to allow to auto-formatting (though I only added the indent width maybe more options could be added to better follow the project style)

# Why?
When using niri, I have some static workspaces to make my Super+number go always to the same workspace, but I don't want my bar to be filled with workspaces that are not in use, so I made this change

### DIFF
--- a/Modules/Bar/Widgets/Workspace.qml
+++ b/Modules/Bar/Widgets/Workspace.qml
@@ -202,8 +202,6 @@ Item {
                     return Color.mOnError
                   if (model.isActive || model.isOccupied)
                     return Color.mOnSecondary
-                  if (model.isUrgent)
-                    return Color.mOnError
 
                   return Color.mOnSurface
                 }
@@ -219,8 +217,6 @@ Item {
               return Color.mError
             if (model.isActive || model.isOccupied)
               return Color.mSecondary
-            if (model.isUrgent)
-              return Color.mError
 
             return Color.mOutline
           }


ItsLemmy:
Cool, I've been wanting to add this feature.   Let us know when it is ready.

mkuritsu:
Yes mostly is working just having some problems in bar not refreshing after toggling the option in the settings, have to check if I need to manually call a refresh or if some data from widget is not being correctly set, that's why I changed this to a draft while I try to learn about the codebase and fix the issue.

ItsLemmy:
I think this should do the trick, as there is no easy other way to monitor a setting change via Connections.

``` 
property var widgetSettings: {
    var section = barSection.replace("Section", "").toLowerCase()
    if (section && sectionWidgetIndex >= 0) {
      var widgets = Settings.data.bar.widgets[section]
      if (widgets && sectionWidgetIndex < widgets.length) {
        return widgets[sectionWidgetIndex]
      }
    }

    // Refresh wokspaces if a setting has changed
    refreshWorkspaces()
    return {}
  }
```

I may optimize it later.

ItsLemmy:
Scratch that, I think you can simply add 

onHideUnoccupiedChanged:  refreshWorkspaces()

mkuritsu:
> Scratch that, I think you can simply add
> 
> onHideUnoccupiedChanged: refreshWorkspaces()

I had tried that but it didn't, the issues was because my dumbass copied the labelMode declaration to be consistent but forgot to change the variable to bool so it was always passing conditions when it shouldn't. Now everything seems to be working properly so I marked this ready.

I also added the value to the registry to have metadata, I seems to be needed looking at other options, also removed a redundant comparison that I found.

ItsLemmy:
Thank you for your contribution!

PR #258, opened 2025-09-12
feat(Modules/Notification): auto-close history panel on clear history

matejc:
I got annoyed by having to always click close after clearing history... I think it makes sense since after clearing the history you do not need to do anything else on the panel other than closing it.


ItsLemmy:
Yep, makes sense. Thank you!

PR #266, opened 2025-09-13
Fix steam games not launching with app2unit

povvke:
With these changes launching steam games works fine now. Everything else launches as well. 

Terminal applications like fish or btop don't launch, so if an application is specified as runInTerminal I just use the previous command.


Ly-sec:
Thank you very much! Merging it :)

PR #267, opened 2025-09-13
docs: mention `gpu-screen-recorder` Flatpak

ThatOneCalculator:


ThatOneCalculator:
Actually, should `gpu-screen-recorder` be in optional rather than mandatory dependencies?

Ly-sec:
> Actually, should `gpu-screen-recorder` be in optional rather than mandatory dependencies?

In theory we could move it to optional dependencies now since we have proper checks in place indeed. I'll think about it and probably change it tomorrow :).

Ly-sec:
I'll merge it for now and keep thinking about the dependency, I'm rather unsure about what the ideal thing is.

PR #272, opened 2025-09-14
fix: power toggle icon alignment

ThatOneCalculator:
Top is before, bottom is after. It's *pretty much* unnoticeable but unless I'm going crazy this is a very minor alignment fix

image


ItsLemmy:
Fixed via Layout. Thanks!

PR #273, opened 2025-09-14
fix(consistency): use unfilled pause icon for suspend in power menu

ThatOneCalculator:
Before:

image

After:

image


ItsLemmy:
We also use this pause for the media player. It's way smaller tho.

 I kept it filled in both places for consistency.   Would you recommend going unfilled for those two ?  (play, pause, skip next, skip prev)

image

ThatOneCalculator:
I think in the media player it looks better filled

ItsLemmy:
Thanks!

Issue #275, opened 2025-09-14
[Bug] Battery display looks strange when set to always show


ThatOneCalculator:
As of latest main (19312d94c3e6492468969aab258bb3c470b457fc):

<img width="113" height="64" alt="Image" src="https://github.com/user-attachments/assets/72d83bc9-00ce-4baf-bf15-43dbfe87c69c" />

ThatOneCalculator:
This is my fix in my personal fork, but this feels a bit hacky: https://github.com/ThatOneCalculator/noctalia-shell/commit/975ea216343ee083de474560689e38763aa3727b

ThatOneCalculator:
Fixed properly by 7e965262f5b36e773902df7c05beadf16e3a7474

ThatOneCalculator:
Technically closed but #276 is an improvement imo

PR #276, opened 2025-09-14
fix: npillhorizontal offset/margin/textcolor

ThatOneCalculator:
See the battery widget, follow-up to #275

Before:

image

After:

image

### DIFF
--- a/Widgets/NPillHorizontal.qml
+++ b/Widgets/NPillHorizontal.qml
@@ -37,7 +37,7 @@ Item {
   // Exposed width logic
   readonly property int iconSize: Math.round(Style.baseWidgetSize * sizeRatio * scaling)
   readonly property int pillHeight: iconSize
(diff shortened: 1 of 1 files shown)


ItsLemmy:
This is not looking good when the Pill is not always open (no breathing room).
It also change the text color to mOnSurface which is inconsistent  with the vertical pill and clock, system monitor, etc... 
I dont know why it was changed to mOnSurface as it was mPrimary for the few last releases.

ThatOneCalculator:
I feel like the text color of percentages has always been the regular foreground color, not the primary color? Unless I'm severely misremembering.

ItsLemmy:
Let's talk on discord :P

ItsLemmy:
Thanks for the contribution, will merge manually with what we discussed.

ThatOneCalculator:
Closed after discussion :)

PR #295, opened 2025-09-16
Kanagawa colorscheme

knuesel:
This adds the Kanagawa colorscheme (only available in dark).

### DIFF
--- a/Assets/ColorScheme/Kanagawa.json
+++ b/Assets/ColorScheme/Kanagawa.json
@@ -1,0 +1,34 @@
+{
+    "dark": {
+        "mPrimary": "#76946a",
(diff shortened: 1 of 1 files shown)


ItsLemmy:
I don't know much about Kanagawa, but looking at git they seem to have a bright/light version (the yellowish one)
We don't support colorscheme without a light version, so we gonna have to come up  with something even if it's not perfect. 

Can you try to implement the light one ?

sijow:
Sure! I don't know how I missed it, I did check or so I thought.

sijow:
There's now also a light variant.

ItsLemmy:
Thanks a lot!

PR #312, opened 2025-09-18
Make password prompt look like a terminal/tty

nalakawula:
This small PR is to adjust the  password prompt on lock screen to look like a password prompt on tty.

Thanks!


Ly-sec:
Very nice, thank you very much :heart: !

PR #313, opened 2025-09-18
Add nix dev shell

matejc:
Add nix dev shell to flake.nix

Usage:

```shell
$ cd ./noctalia-shell
$ nix develop
```

it should open shell where you can launch the editor from, in my case I have neovim with qmlls language server enabled (using nvim-lspconfig)

Note: nvim-lspconfig uses `qmlls -E` (-E means that it reads imports from environment variable QML_IMPORTS_PATH, and not through the `.qmlls.ini`, which I think is better to read from env var since in that config you need to put static paths and not everyone have packages on the same path)

But I have some import issues with everything that is imported with `qs.` prefix

image

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -79,6 +79,17 @@
         });
 
       defaultPackage = eachSystem (system: self.packages.${system}.default);
+
+      devShells = eachSystem (system: let
+        pkgs = nixpkgs.legacyPackages.${system};
+      in {
+        default = pkgs.mkShell {
+          packages = [ pkgs.qt6Packages.qtdeclarative ];  # for qmlls executable
+          shellHook = ''
+            export QML_IMPORT_PATH="$PWD:${pkgs.qt6Packages.qtdeclarative}/lib/qt-6/qml:$QML_IMPORT_PATH"
+          '';
+        };
+      });
     };
 }
 


Ly-sec:
Any news about this?

matejc:
I am sorry, I did not have time to continue since then.
But there is an open question: How to resolve `qs.*` kinda imports by using qmlls... we might need to add some qmldir files in certain directories, but haven't found any easily usable way to do that. I also discovered if I move ALL directories with source code into repo/qs/* place then there is some progress... But yeah, I am first time working on QML or QT related project. If you have any ideas, feel free to open up here.

Ly-sec:
> I am sorry, I did not have time to continue since then. But there is an open question: How to resolve `qs.*` kinda imports by using qmlls... we might need to add some qmldir files in certain directories, but haven't found any easily usable way to do that. I also discovered if I move ALL directories with source code into repo/qs/* place then there is some progress... But yeah, I am first time working on QML or QT related project. If you have any ideas, feel free to open up here.

No worries!
Also no you shouldn't need qmldir's since `import qs.ABC` basically acts as qmldir from what I understood.

ixxie:
@matejc this has been hanging for a whole, so I'm closing this in favor of https://github.com/noctalia-dev/noctalia-shell/pull/530

If you have some additions to that work which you would like to propose, feel free to open a new PR!

matejc:
Sure

PR #314, opened 2025-09-19
fix(clock): respect monthBeforeDay setting in vertical clock date dis…

kevindiaz314:
- Fix vertical bar clock to show month before day when `monthBeforeDay` setting is enabled
- Ensures consistent date ordering across all clock display modes


ItsLemmy:
Thanks!

PR #322, opened 2025-09-20
nix flake: systemd service + home manager settings

ixxie:
Additions to the `flake.nix`:
- `nixosModule`: a systemd service which you activate with `services.noctalia-shell.enable = true`
- `homeModule`: an option to configure settings an attribute set, string or filepath
- `homeModule`: an option to configure colors an attribute set, string or filepath
- Formats the flake using `nixfmt-rfc-style`

### Nix Flake Installation
        
For Nix users, Noctalia provides both NixOS and home-manager modules:
        
#### NixOS Module (systemd service)

This will automatically start after Niri. 

>[!NOTE]
> For **Hyprland**, it depends on how you configure your systemd service. Many options like UWSM will setup `graphical-session.target` for you an _should_ work by default. This has not been tested, let us know if you find bugs.
```nix
{
  imports = [ inputs.noctalia-shell.nixosModules.default ];
        
  services.noctalia-shell = {
    enable = true;
    target = "graphical-session.target"; # this is the default; use "hyprland-session.target" if you have something like that
  };
}
```
        
#### Settings

>[!IMPORTANT]
> You **must** set all configuration options, since this overrides the *full* settings file.
> Its good to backup your current config - found in `~/.config/settings.json` before trying,
> and to use this file as a value here. We hope to fix this in a future PR; see https://github.com/noctalia-dev/noctalia-shell/issues/323

```nix
{
  home-manager.users.mrfoobar = {
    imports = [ inputs.noctalia-shell.homeModules.default ];
        
    programs.noctalia-shell = {
      enable = true;
      settings = {
        # put *all* settings here, see warning above
     };
    };
  };
}
```

#### Colors

You can configure your color theme like this:

>[!WARNING]
>This is experimental. We are not entirely sure how Mutagen behaves when you write this. In my own system so far, this has been ok.

```nix
{
   # ...        
    programs.noctalia-shell = {
      enable = true;
      colors = {
        # put *all* colors here, see warning above
     };
    };
  };
}
```

The home-manager module automatically restarts the systemd service when settings change. I've noticed a bug where rebuilding the system can hang, but after modifying the `onChange` hook and the systemd service restart settings, it seems fine.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -146,26 +146,77 @@
                 or filepath, to be written to ~/.config/noctalia/settings.json.
               '';
             };
+
+            colors = lib.mkOption {
+              type =
+                with lib.types;
+                nullOr (oneOf [
+                  attrs
+                  str
+                  path
+                ]);
+              default = null;
+              example = lib.literalExpression ''
+                 {
+                   mError = "#dddddd";
+                   mOnError = "#111111";
+                   mOnPrimary = "#111111";
+                   mOnSecondary = "#111111";
+                   mOnSurface = "#828282";
+                   mOnSurfaceVariant = "#5d5d5d";
+                   mOnTertiary = "#111111";
+                   mOutline = "#3c3c3c";
+                   mPrimary = "#aaaaaa";
+                   mSecondary = "#a7a7a7";
+                   mShadow = "#000000";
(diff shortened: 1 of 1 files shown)


ItsLemmy:
Thank you!

PR #330, opened 2025-09-21
DisplayTab fixes

MarkusVolk:


ItsLemmy:
Thanks for the PR. I can't merge it in its current form, but I appreciate you flagging the layout problem.

We need to avoid using space characters for spacing, as the RowLayout has its own spacing properties. Using a manual space can hide the actual bug, which we need to fix properly.

I agree that the Manual Scheduling view is cramped, and I will rework it to improve the layout. I'll be keeping the labels, but will adjust the layout.

So I can investigate further, could you tell me what resolution and display scaling you are using? I didn't notice any issues during my tests on QHD and 1080p.

ItsLemmy:
I have committed some changes to the display tab, let me know if that helps. Thanks

MarkusVolk:
On Sun, Sep 21 2025 at 13:52:43 -07:00:00, Lemmy 
***@***.***> wrote:
> So I can investigate further, could you tell me what resolution and 
> display scaling you are using?

Sorry for being late. My resolution currently is 3840x2160 with scale 
1.5
But I can confirm that with the latest version both issues are gone. 
Thanks for the quick fix!

PR #336, opened 2025-09-22
fix(bar): Ensure SystemMonitor temperature is fully visible

lonerOrz:
In the vertical bar layout, the temperature text in the SystemMonitor widget (e.g., "55°C") could be truncated due to the widget's fixed width.
  This commit resolves the issue by applying a dynamic scale
transformation to the text component.

图片


Ly-sec:
Very smart way of fixing that issue. Well done!

Issue #329, opened 2025-09-20
[Bug] Notifications don't scale with display scaling


Drazzy9295:
### Description
Notifications don't scale with the display scaling in the settings menu

### Steps to Reproduce
1. Go to 'Settings''
2. Click on 'Display'
3. Increase monitor scale
4. Run "notify-send hello"

### Expected Behavior
For the notification window to scale

### Environment
- Distro:  NixOS
- Compositor: Hyprland
- Noctalia-shell Version: Flake (Latest commit)

Ly-sec:
Fixed in [879d9ec](https://github.com/noctalia-dev/noctalia-shell/commit/879d9ec879f8091658581985b10c9aeda5b3e939).

PR #339, opened 2025-09-22
fix: Notifications don't scale with display scaling

lonerOrz:
fixed: #329

This ensures that notification popups now correctly adapt their size and elements based on the current display scaling factor.


Ly-sec:
Fixed in [879d9ec](https://github.com/noctalia-dev/noctalia-shell/commit/879d9ec879f8091658581985b10c9aeda5b3e939). Sorry, didn't see the PR early enough!

PR #345, opened 2025-09-23
Clock Widget: Add option to change font size of the clock widget

DuckySoLucky:
What is done:
- Added option to change font size of the clock widget
> Wasn't really a huge fan of default font size of the clock, so I added option to scale it, I'm not QML dev so i'm not entirely sure if this is the best way to do it or anything, I did this more for myself-

https://jumpshare.com/s/YS4JR0AWK99SAXNyGiho


NeeRaj-2401:
> Wasn't really a huge fan of default font size of the clock, 

Sometimes i feel the same for entire shell's of font family, size and weight combination.

DankMaterialShell looks better in this term, but noctalia is more feature rich

Ly-sec:
The font "size" is mostly depending on the mono font you set. Which is something we can't always account for given the amount of options.

ItsLemmy:
Thanks for the PR! While we won't be merging it, we appreciate the effort.

We're avoiding fixed pixel sizes in Noctalia to ensure everything works with our fractional scaling system. However, we agree that font sizing needs to be adjustable, and we'll be introducing a font scaling feature soon to give users that control.

Thanks again!

PR #351, opened 2025-09-24
Audio changes

FUFSoB:
What is done:
- Allow volume overdrive (going above 100%; configurable, default false)
- Fix inconsistent volume values
- Fix shown volume after sink/source switch

### DIFF
--- a/Assets/settings-default.json
+++ b/Assets/settings-default.json
@@ -137,6 +137,7 @@
   },
   "audio": {
     "volumeStep": 5,
+    "volumeOverdrive": false,
     "cavaFrameRate": 60,
     "visualizerType": "linear",
     "mprisBlacklist": [],
--- a/Commons/Settings.qml
+++ b/Commons/Settings.qml
@@ -261,6 +261,7 @@ Singleton {
     // audio
     property JsonObject audio: JsonObject {
       property int volumeStep: 5
+      property bool volumeOverdrive: false
       property int cavaFrameRate: 60
       property string visualizerType: "linear"
       property list<string> mprisBlacklist: []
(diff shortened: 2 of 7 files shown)


ItsLemmy:
Thanks for the PR! 

I'm having some doubt about the volumeOverdrive tho.
For some reason it does not work on my system (we had that feature in an older version of noctalia and it did not work either.)    

Basically, if i go over 100% 
- the sound does not get louder
- the slider in the settings / audio tab still works visually but it stop changing the volume, it get fixed if I use mediakeys to alter the volume

 I'm on arch, might be some config difference on my side. I don't think I have tweaked my pw stack at all, but if you have any ideas I would be interested to hear them.

ItsLemmy:
Probably a hardware limitation of my Audeze Maxwell...

FUFSoB:
I have pretty much default PW setup, as in, I didn't modify any settings specifically for this to work. First time I used this feature in KDE, but I don't think it modifies the PW stack. I noticed that slider volume can't go above 100% because one of my headsets is on 105%.

In my testings sliders work perfectly, I can hear the difference between 100% and 150%. Using Arch. Works on both usb and 3.5-jack headsets.

Not sure about your case :(

Try `wpctl set-volume @DEFAULT_AUDIO_SINK@ 1.5` to change the volume. Then `wpctl get-volume @DEFAULT_AUDIO_SINK@` will show 1.5 as it should. Does it make any difference in sound?

ItsLemmy:
Yeah I tried the wpctl way half an hour ago. It's definitely my hardware.  Volume will never go above 100%.   I  Will probably just live with it as I don't like distortion anyways. Maybe we can improve the description of the NToggle to make it clear that it may work or not.

FUFSoB:
> Maybe we can improve the description of the NToggle to make it clear that it may work or not.

Reflected this in #352. Also did a bugfix for this PR I noticed

PR #356, opened 2025-09-24
Initialize volume silently

lonerOrz:
Initialize volume silently


ItsLemmy:
Have not tried it but it feels like we dont need the extra variable InSilentPeriod.
Why not just have the timer set both MuteInitialized and VolumeInitialized to true when it triggers ?
I might be wrong, but the simpler the better.

ItsLemmy:
Excellent, thanks

PR #358, opened 2025-09-25
Fix brightness sync after external command changes

lonerOrz:
Fix brightness sync after external command changes, improve brightness module compatibility

### DIFF
--- a/Services/BrightnessService.qml
+++ b/Services/BrightnessService.qml
@@ -119,6 +119,51 @@ Singleton {
     // Signal for brightness changes
     signal brightnessUpdated(real newBrightness)
 
+    // Execute a system command to get the current brightness value directly
+    readonly property Process refreshProc: Process {
+      stdout: StdioCollector {
+        onStreamFinished: {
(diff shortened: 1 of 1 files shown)


ItsLemmy:
Thanks for the PR, have you tried this with both Internal and External displays(DDC) ?
BrightnessService is kinda fragile.

lonerOrz:
> Thanks for the PR, have you tried this with both Internal and External displays(DDC) ? BrightnessService is kinda fragile.

I tested it on my laptop and everything worked fine, but I didn't test it on an external monitor.

lonerOrz:
The built-in display and external display have been tested and both functions are normal.

ItsLemmy:
Looking good !

But this "split" hurts my soul.
image

Can you make it more robust with something along those lines.
```
var lines = dataText.split("\n")
```

Thanks!

lonerOrz:
> var lines = dataText.split("\n")

done

lonerOrz:
thanks！

PR #370, opened 2025-09-26
Rounds the ends of NSliders to be more consistent with the look

MrDowntempo:
Admittedly a very very minor change. But just look at it!

image

I noticed Height/2 is used in the same file previously, and there might be some optimization possible by calculating it once and reusing it. But I wasn't certain if that was worth it since it would also (i guess) mean bytes of memory just being wasted all willy nilly.

~~If you want me to attempt that optimization, lemme know and I'll tweak the PR on my end.~~

Actually I think it wouldn't help after all, so this is likely fine as is.

### DIFF
--- a/Widgets/NSlider.qml
+++ b/Widgets/NSlider.qml
@@ -15,6 +15,8 @@ Slider {
   readonly property real trackHeight: knobDiameter * 0.4
   readonly property real cutoutExtra: Math.round(Style.baseWidgetSize * 0.1 * scaling)
 
+  padding: cutoutExtra / 2
+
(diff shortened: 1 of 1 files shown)


MrDowntempo:
Lemmy broke my poor little newbie heart by telling me he had purposely removed this feature in the past because the left side got squished when you brought the slider close to 0. I confirmed that behavior. He also told me he worked on it for hours, pouring over it only to admit defeat. But I just fixed it sooo damn hard. Be amazed. Old tricks are the best tricks. I'm using a semicircle and rectangle instead of rounding the rectangle. Or more accurately a full circle with the rectangle sitting on top of it hiding half of it.

image

PR #371, opened 2025-09-26
 allow zh-CN like language code

pugaizai:


pugaizai:
hi, please review my code @ItsLemmy

PR #372, opened 2025-09-26
Centered circles

MrDowntempo:
Sometimes the circles of Toggles and the slider were off centered ever so slightly. I believe I've resolved this with a bit of rounding, and occasionally ditching ImplicitHeight for Height. I don't ***THINK*** that I broke anything.

before:
image

after:
image

### DIFF
--- a/Widgets/NRadioButton.qml
+++ b/Widgets/NRadioButton.qml
@@ -19,9 +19,8 @@ RadioButton {
     anchors.verticalCenter: parent.verticalCenter
 
     Rectangle {
-      anchors.centerIn: parent
-      implicitWidth: Style.marginS * scaling
-      implicitHeight: Style.marginS * scaling
+      anchors.fill: parent
+      anchors.margins: parent.width * 0.3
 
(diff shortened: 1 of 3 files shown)


MrDowntempo:
Oh, I need to try and fix radio buttons too

ItsLemmy:
Have you tried with different scaling?

MrDowntempo:
> Have you tried with different scaling?

I tested with 2x scaling in niri

image

ItsLemmy:
what about Noctalia's scaling from the Display/ tab

MrDowntempo:
> what about Noctalia's scaling from the Display/ tab

It works... uh.. mostly. ;_; There's a range of values where it don't though. ~130-160ish. Only the slider seems to be affected. I'll squish this with a nother commit. Probably just needs another Math.round() somewhere.

MrDowntempo:
I think we're good to go!

ItsLemmy:
Good job Bobby!

PR #380, opened 2025-09-27
feat(flake): deep merge settings with defaults

ixxie:
This uses the new `settings-default.json` to set default values in the flake.

It means that users configuring with the flake using attribute sets can now set a partial configuration.


ixxie:
I realized I didn't actually properly test this... so consider it still WIP

ixxie:
Now verified to work as expected

PR #388, opened 2025-09-28
Nicer SpinBox with better mTertiary hover

MrDowntempo:
The Widget's outline used to only change color when you hovered over the middle, now it changes wherever you hover. The circles on the end were changed to semicircles that fade into the widget. And the hover color is now mTertiary to be consistent with other widgets that have a hover color. I've tested at all scale levels and that was the most frustrating thing to fix. I had to get rid of the inner border (the .8 * parent.width) but with this effect I think it looks better without it. And **NO**, I'm not just rationalizing that in light of how impossible it was to consistently vertically center the semicircles when there was an inner border. :)

### DIFF
--- a/Widgets/NSpinBox.qml
+++ b/Widgets/NSpinBox.qml
@@ -5,6 +5,7 @@ import qs.Commons
 import qs.Services
 import qs.Widgets
 
+
 RowLayout {
   id: root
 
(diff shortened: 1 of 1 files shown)


MrDowntempo:
Tested my merges (i think). Still works across all scaling values, and retains the text entry.

ItsLemmy:
Thanks!

PR #393, opened 2025-09-29
Sort windows in Taskbar by their scrolling position on Niri

luleyleo:
Also fixes stale focus state when opening a new window.

### DIFF
--- a/Services/NiriService.qml
+++ b/Services/NiriService.qml
@@ -227,6 +227,7 @@ Item {
 
         // Only emit activeWindowChanged if the focused window actually changed
         if (oldFocusedIndex !== focusedWindowIndex) {
+          windows[oldFocusedIndex].isFocused = false
           activeWindowChanged()
         }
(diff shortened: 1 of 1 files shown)


ItsLemmy:
Thanks!

ItsLemmy:
It's working pretty well, but I think there is still an edge case when using the mouse drag and drop.
image
If I move the window via keybinds, its always good.
If I move the window by drag&drop with mouse it works ok, unless I drop the window at the very right. In that case the icon (nautilus) stay at the original position.  This is fixed by changing workspace, focusing another screen, opening an new window.

ItsLemmy:
While QML supports ES6, we can't use all of its syntax in Noctalia.
Specifically, array destructuring like const [windowId, layout] = myArray; is not supported.

Our project's formatting tool, qmlfmt, fails to parse this syntax, which corrupts the file and makes it unusable.

PR #394, opened 2025-09-29
[NixOS] feat/temp settings

ixxie:
If you use the new flake option to declaratively configure Noctalia, `~/.config/settings.json` becomes a symlink to the Nix store, and is managed by Nix.

Therefore, edits in the GUI settings panel succeed in memory, but silently fail to write the `settings.json` file, with a `PERMISSION DENIED` error.

This isn't really a problem for using Noctalia as-such; the settings _should_ befined by the user's Nix dotfiles in this case. However, NixOS users using the settings option have a challenge: how to figure out the available configuration options?

I added the _defaults_ to the documentation; but this obviously doesn't reveal _all_ the options. In an ideal world, all options would be meticulously documented somehow. But I think doing this would be too much work to maintain, at least at the moment.

This adds a fallback write path, configurable with `NOCTALIA_SETTINGS_FALLBACK`. When provided, it writes the config there in addition to the default location. The flake then sets it in the systemd service.

### DIFF
--- a/Commons/Settings.qml
+++ b/Commons/Settings.qml
@@ -71,7 +71,13 @@ Singleton {
     id: saveTimer
     running: false
     interval: 1000
-    onTriggered: settingsFileView.writeAdapter()
+    onTriggered: {
+      settingsFileView.writeAdapter()
+      // Write to fallback location if set
+      if (Quickshell.env("NOCTALIA_SETTINGS_FALLBACK")) {
+        settingsFallbackFileView.writeAdapter()
+      }
+    }
   }
 
   FileView {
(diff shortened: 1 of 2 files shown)


ppenguin:
@ixxie Thanks for this feature! Sorry for replying to this closed issue, but since there's no discussions, I figured it's less "invasive" than opening an issue for some questions:

- From quickly perusing the code I figured that `gui-settings.json` is write-only, so its settings wouldn't take effect (i.e. be merged) after restart, correct? 
- That means the intended usage is to manually add the settings from `gui-settings.json` to the declarative config?

TIA!

ixxie:
> @ixxie Thanks for this feature! Sorry for replying to this closed issue, but since there's no discussions, I figured it's less "invasive" than opening an issue for some questions:
> 
> - From quickly perusing the code I figured that `gui-settings.json` is write-only, so its settings wouldn't take effect (i.e. be merged) after restart, correct? 
> - That means the intended usage is to manually add the settings from `gui-settings.json` to the declarative config?
> 
> TIA!
> 

I'm out of the loop and things have moved fast. Ask on Discord!

ixxie:
> @ixxie Thanks for this feature! Sorry for replying to this closed issue, but since there's no discussions, I figured it's less "invasive" than opening an issue for some questions:
> 
>     * From quickly perusing the code I figured that `gui-settings.json` is write-only, so its settings wouldn't take effect (i.e. be merged) after restart, correct?
> 
>     * That means the intended usage is to manually add the settings from `gui-settings.json` to the declarative config?
> 
> 
> TIA!

@ppenguin my response was a bit quick and reflexive since I was super stressed.

The motivation for this PR was to make settings visible. The fallback path allows Noctalia to write settings somewhere when you use the settings GUI mutably. However, your NixOS declarative configuration is the main source of truth. The reason for having a fallback path is that you can check it to see what happens when you toggle a setting, and then copy it to your nixos config to make it permanent.

Since this PR, we show the configuration defaults in the documentation site, so this feature is less important.

ppenguin:
> @ppenguin my response was a bit quick and reflexive since I was super stressed.

No worries, it got me what I needed. FYI, about yesterday this option was removed upstream in favour of the documentation.

PR #396, opened 2025-09-30
NiriService: Keep windows sorted when moving them with the mouse

luleyleo:
This fixes the issue observed in https://github.com/noctalia-dev/noctalia-shell/pull/393#issuecomment-3349237165

### DIFF
--- a/Services/NiriService.qml
+++ b/Services/NiriService.qml
@@ -217,8 +217,8 @@ Item {
       } else {
         // Add new window
         windows.push(newWindow)
-        windows.sort(compareWindows)
       }
+      windows.sort(compareWindows)
 
       // Update focused window index if this window is focused
       if (newWindow.isFocused) {


ItsLemmy:
Hey, thanks for the fix.
We are still experimenting a lot of ot the warning below, any idea ?
Only happens when reordering a window with the mouse on Niri.

```  WARN: <Unknown File>: workspaceId is null. Adding an object with a null member does not create a role for it.
  WARN: <Unknown File>: workspaceId is null. Adding an object with a null member does not create a role for it.
  WARN: <Unknown File>: workspaceId is null. Adding an object with a null member does not create a role for it.```

**Comment by luleyleo** [comment]
I saw those too, but I haven't managed to reproduce them deterministically yet. I have another idea to try once I'm back though.

**Comment by ItsLemmy** [comment]
This seems to do the trick... but meh  :P  

using -1 instead of null

```
  function getWindowData(win) {
    return {
      "id": win.id,
      "title": win.title || "",
      "appId": win.app_id || "",
      "workspaceId": win.workspace_id || -1,
      "isFocused": win.is_focused === true,
      "output": getWindowOutput(win) || "",
      "position": getWindowPosition(win.layout)
    }
  }
```

luleyleo:
Actually, that kind of makes sense. While dragging, the window isn't attached to any workspace. I think setting it to -1 would be reasonable.

ItsLemmy:
Great - commited!

ItsLemmy:
also getting this one, but did not find a repro step yet.
 ```ERROR qml: [20250930-084738]    NiriService Error handling WindowOpenedOrChanged: TypeError: Value is undefined and could not be converted to an object```

PR #398, opened 2025-09-30
Add clock font setting for customizable clock displays

DiscoCevapi:
This commit adds a new 'Clock Font' setting that allows users to customize the font used specifically for clock displays in the bar and widgets, independent of the default UI font.

Features:
- New clockFont property in Settings.data.ui (defaults to 'Roboto')
- Updated Bar Clock widget to use the custom font with fallback support
- Added searchable font dropdown in General Settings tab
- Backward compatible - uses default font if clockFont is not set
- Real-time updates - changes apply immediately

The font selection uses FontService.availableFonts and includes proper fallback logic that respects the existing monospaced font setting. Currently has only been tested on Niri, i'm still new to this all so i hope this is something atleast useable. It works for me currently...

### DIFF
--- a/Commons/Settings.qml
+++ b/Commons/Settings.qml
@@ -272,6 +272,7 @@ Singleton {
     property JsonObject ui: JsonObject {
       property string fontDefault: "Roboto"
       property string fontFixed: "DejaVu Sans Mono"
+      property string clockFont: "Roboto"
       property real fontDefaultScale: 1.0
       property real fontFixedScale: 1.0
       property list<var> monitorsScaling: []
--- a/Modules/Bar/Widgets/Clock.qml
+++ b/Modules/Bar/Widgets/Clock.qml
@@ -65,7 +65,7 @@ Rectangle {
           NText {
             visible: text !== ""
             text: modelData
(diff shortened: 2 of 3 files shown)


ItsLemmy:
Hey thanks for the PR.
We don't want to add anymore fonts to the general settings.
We also tried to avoid as much as possible to have widget settings in the main settings.
 Can you move the settings to the clock widget settings ? 
Relevant files are /Bar/WidgetSettings/ClockSettings.qml  and  /Services/BarWidgetRegistry.qml

DiscoCevapi:
Hey not a problem sorry about that! i will move them tomorrow afternoon
when I am back on the computer currently about to go to bed then work
tomorrow

ItsLemmy:
you can remove the clock option "use monospaced font" and put yours in its place.

DiscoCevapi:
Was that there before I did what I did? If not that sick now thanks! I just
got a thing for Dseg(lcd) font for clocks 😄

ItsLemmy:
Thanks, yeah the use monospaced font has been there for a week or so :P

PR #405, opened 2025-10-02
Fix clock font dropdown

DiscoCevapi:
1. UI Freezing - Font dropdown no longer freezes when clicking
2. Missing Fonts - Now shows all 3000+ system fonts instead of 21 fallback fonts  
3. Console Warnings - Eliminated all ReferenceError and undefined assignment warnings
4. Property Errors - Added missing valueUseMonospacedFont property

### DIFF
--- a/Modules/Settings/Bar/WidgetSettings/ClockSettings.qml
+++ b/Modules/Settings/Bar/WidgetSettings/ClockSettings.qml
@@ -14,12 +14,66 @@ ColumnLayout {
   property var widgetData: null
   property var widgetMetadata: null
 
-  // Local state
-  property bool valueUsePrimaryColor: widgetData.usePrimaryColor !== undefined ? widgetData.usePrimaryColor : widgetMetadata.usePrimaryColor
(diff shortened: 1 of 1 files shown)


ItsLemmy:
This PR looks very weird to me.

I removed the valueUseMonospacedFont on purprose, I dont want 2 mechanisms to select font.
The way I see it, either you go with the standard non-monospaced font,  or you pick a custom font.  It avoids having multiple options for the same purpose, as the clock settings is already huge.

Also lots of other things are not right imo.

// Local state with safe defaults - use explicit default values to avoid undefined assignments   
We dont need to unfold the code like that. WidgetMetaData is always existing, so we don't want to have extra fallback data as we already have one.   If you are facing issue where WidgetMetaData is undefined then it is sign there is a bigger issue that need to be fixed.

Same thing for FontService, it's a singleton so it's always available.

We don't do console.log, we use the proper Logger class.

I don't see why we would need some hardcoded fallback fonts?  

Let's talk on discord if needed!

PR #411, opened 2025-10-02
Basic sway support

### LINKED ISSUE #403: [Feature] Sway support
### Feature Description
Support for the sway compositor

### Why Is This Needed?
I use sway and I want a more fully featured bar/shell than waybar et al

### Suggested Solutions
Provide needed integration points for using noctalia with sway

### Additional Context
I've already cobbled something together that works. If there's interest on supporting sway upstream I'll try to extract some reasonable patches to send off.

https://github.com/keis/noctalia-shell/tree/my-hacks

keis:
Brings basic sway support

focus/close window is kinda buggy with the current setup. I found the Toplevel object[1] has an convenient `activate()` method that might work but I'm not sure it's a good idea to pass around this object instead of the hash of details like now.

Fixes #403 

1: https://quickshell.org/docs/v0.1.0/types/Quickshell.Wayland/Toplevel/

### DIFF
--- a/Modules/Bar/Widgets/Taskbar.qml
+++ b/Modules/Bar/Widgets/Taskbar.qml
@@ -105,13 +105,13 @@ Rectangle {
 
             if (mouse.button === Qt.LeftButton) {
               try {
-                CompositorService.focusWindow(taskbarItem.modelData.id)
+                CompositorService.focusWindow(taskbarItem.modelData)
               } catch (error) {
                 Logger.error("Taskbar", "Failed to activate toplevel: " + error)
               }
             } else if (mouse.button === Qt.RightButton) {
               try {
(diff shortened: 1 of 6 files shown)


Ly-sec:
Hey there, first of all thank you very much for this PR. I'm wondering if you checked if hyprland/niri still behaves fine after your changes?
For the activate() part, I'm not entirely sure myself right now, perhaps it could be worth a try :)

keis:
I've tested with niri and everything seems to work. Hyprland I can't get running on my system at all so didn't test with that.

Tried the wayland activate() calls and it works!

ItsLemmy:
I need some times to test this properly!

ItsLemmy:
Thanks!

PR #412, opened 2025-10-02
fix: never scroll media mini on pause

ThatOneCalculator:
Media players almost never scroll the title when the media is paused


ItsLemmy:
Thanks

Issue #407, opened 2025-10-02
[Feature] Idle Inhibitor inhibits suspend on lid close


matejc:
### Feature Description

I would like to discuss if there is a chance to remove lid closing idle inhibiting (systemd), or at least optionally disable it in settings panel.

### Why Is This Needed?

I believe (due to my personal workflow) that idle inhibiting on lid close should not be a thing.
When working I frequently enable Keep Alive and that in turn also disables suspend on lid close, and it happens that when I go home from office I am always forgetting to disable Keep Alive and the laptop never suspends, and since my laptop does not have any lights that indicate its still running... The laptop gets quite warm in bag and eventually dies from battery depletion.

What do you think?

### Suggested Solutions

Like I said before, either:
- remove lid closing from idle inhibiting, or
- make an option to disable it

### Additional Context

I can implement either first or second solution, but for the second, I would like to know where to add that disabling option (on which panel, I would hate to make its own panel for one option)?

ItsLemmy:
Feel free to remove lid closing from IdleInhibitorService and create a PR.
Anyway this service will be retired once Quickshell updates with  a builtin inhibitor.
Thanks!

matejc:
Any ETA on the Quickshell update of a builtin inhibitor?

ItsLemmy:
Nothing official, gotta ask upstream.   Heard it might come after Qt 6.10 is released.

PR #423, opened 2025-10-05
systemd-inhibit: remove handle-lid-switch

matejc:
As discussed in https://github.com/noctalia-dev/noctalia-shell/issues/407


ItsLemmy:
Thanks!

PR #425, opened 2025-10-05
update lastSeenTs in notifications toggleHistory ipc call

damian-ds7:
Currently only clicking on the notification icon will update timestamp and mark all notifications as read, with this change it will be possible when opening with `qs -c noctalia-shell ipc call notifications toggleHistory`. This enables marking notifications as read programmatically via IPC.

### DIFF
--- a/Modules/Bar/Widgets/NotificationHistory.qml
+++ b/Modules/Bar/Widgets/NotificationHistory.qml
@@ -61,7 +61,6 @@ NIconButton {
   onClicked: {
     var panel = PanelService.getPanel("notificationHistoryPanel")
     panel?.toggle(this)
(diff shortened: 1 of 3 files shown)


ItsLemmy:
Hey, thanks for the PR. Can you please rework it a bit.
- remove the timestamp change from Bar/Widgets/NotificationHistory.qml
- don't add anything in IPCService.qml
- change the timestamp when the panel actually opens, that's in NotificationHistoryPanel.qml
you can add the following signal handler at the top and do the timestamp thing here.
```
onOpened: function() {
}
```

damian-ds7:
Yeah sure I can do that, no problem

ItsLemmy:
Perfect, thanks a lot for your contribution.

PR #427, opened 2025-10-05
Create intermediate folders for theme syncing

ecoskey:
Creates intermediate directories for placing synced theme files, avoiding an error if the directory is missing.

### DIFF
--- a/Services/MatugenService.qml
+++ b/Services/MatugenService.qml
@@ -422,21 +422,24 @@ Singleton {
     if (Settings.data.templates.foot) {
       if (copyCmd !== "")
         copyCmd += " ; "
-      copyCmd += `cp -f ${getTerminalColorsTemplate('foot')} ~/.config/foot/themes/noctalia`
+      copyCmd += `mkdir -p ~/.config/foot/themes/`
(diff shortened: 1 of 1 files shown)


Ly-sec:
I was thinking about that, wouldn't it be easier to have a more generalized solution? Something like grab last folder and mkdir based on that instead of having to add it per config/template?

ecoskey:
> I was thinking about that, wouldn't it be easier to have a more generalized solution? Something like grab last folder and mkdir based on that instead of having to add it per config/template?

yep, I just did the simplest thing for starters. I'll do that + some extra cleanup so there isn't so much repeated code here

ecoskey:
Should be pretty much done :)

Another thing I found: logs show `matugen image` on `matugen` 2.4.1 doesn't recognize the argument `--import-json`. Seems like an update broke this maybe?

Ly-sec:
> Should be pretty much done :)
> 
> Another thing I found: logs show `matugen image` on `matugen` 2.4.1 doesn't recognize the argument `--import-json`. Seems like an update broke this maybe?

For that you do need matugen-git, gotta update the docs and stuff later.

PR #428, opened 2025-10-06
Add media ipc target

ecoskey:
I missed my media hotkeys switching to noctalia :) 

Lmk if this is/isn't up to y'alls standards, I didn't know if there were any calls I should make other than forwarding to the media service.

### DIFF
--- a/Services/IPCService.qml
+++ b/Services/IPCService.qml
@@ -166,4 +166,45 @@ Item {
       Settings.data.wallpaper.randomEnabled = true
     }
   }
+
+  IpcHandler {
+      target: "media"
+      function playPause() {
+        MediaService.playPause()
+      }
+
+      function play() {
+        MediaService.play()
+      }
+
+      function pause() {
+        MediaService.pause()
+      }
+      
+      function next() {
+        MediaService.next()
+      }
+
+      function previous() {
+        MediaService.previous()
+      }
+
+      function seek(position: string) {
+        var positionVal = parseFloat(position)
+        if (Number.isNaN(positionVal)) {
+            Logger.warn("Media", "Argument to ipc call 'media seek' must be a number")
+            return
+        }
+        MediaService.seek(positionVal)
+      }
+
+      function seekByRatio(position: string) {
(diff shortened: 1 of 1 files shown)


Ly-sec:
Hey! I just tested your PR and I'm a bit unsure what the intention for the seek part is. It does only seem to go to position X (where X is second/%) instead of moving +/- X. Could you please explain that?

ecoskey:
> Hey! I just tested your PR and I'm a bit unsure what the intention for the seek part is. It does only seem to go to position X (where X is second/%) instead of moving +/- X. Could you please explain that?

Mainly I just wanted to expose the underlying `seek` function and keep its behavior. I didn't have a specific use in mind but having it there seems nice. Maybe people will find a use for it with custom music widgets or something? Idk. Would you like me to add something like a `seekForwardBy` function?

ItsLemmy:
I would just replace the absolute seek by relative (-+ 10s)   I dont see any real life use case for the absolute ones.

ItsLemmy:
Thanks!

notiant:
Could you please add an ipc call for a `stop` function?

ItsLemmy:
I did update the documentation, to reflect the most recent keybinds.  We don't have stop, but we got pause.

https://docs.noctalia.dev/getting-started/keybinds/#media-controls

notiant:
It should be as easy as adding this to MediaService.qml
```
  function stop() {
    if (currentPlayer) {
      currentPlayer.stop()
    }
  }
```

and this to IPCService.qml
```
      function stop() {
        MediaService.stop()
      }
```
I have a stop button in my function keys and would like to use it as intended :)

ItsLemmy:
Yeah I know how to do it :)
But I assumed you were not aware of the existing IPC calls as they were not documented, and most people don't use stop, but rather pause / play.

Will add it soon.

ItsLemmy:
the IPC call is in, but be aware that once you call stop, you can no longer call play via IPC on that player (at lease from my testing on Youtube, Spotify and Twitch)

PR #433, opened 2025-10-06
fix(settings): Correct brightness slider layout issues

### LINKED ISSUE #430: [Bug] "internal" text in Display settings overlap w/ monospace font
### Description
The "internal" label under the brightness slider in the display settings gets overlapped w/ the brightness % text when using a monospace font

Image

lonerOrz:
fixed #430 

图片


ItsLemmy:
It might fix the issue, but I find the look very odd / messy.  I'd  like to keep both sliders mostly aligned/symetrical

lonerOrz:
I refactored the layout of these two sliders， they should now be aligned.

ItsLemmy:
thanks!

PR #434, opened 2025-10-06
fix(settings): make color scheme grid layout responsive

lonerOrz:
The color scheme grid previously used a fixed number of columns, which caused cards to be partially clipped when the window width was narrow.

This change makes the number of columns dynamically calculated based on the available width. It ensures the grid layout gracefully adapts from 3 (or more) columns down to 1 column, preventing content overflow and clipping.

This fixes a UI bug where the right-side content of color scheme cards was not visible at certain window widths.

图片


ItsLemmy:
Thanks

PR #438, opened 2025-10-07
feat: Merge duplicate MPRIS players when browser dual report

MrDowntempo:
Librewolf has shoed up as both Firefox and Librewolf at the same time. Both give some metadata, but only firefox has the album art. I've hopefully made a robust way for forked browsers to get combined into one virtual player that gets all the metadata, and uses the forks title. I've tested this with Chrome (which doesn't give album art no matter what) Ungoogled Chromium, Firefox (to make sure it doesn't get confused) and Librewolf running simultaneously. Probably needs to be tested with a few more forks, but I believe all major issues have been eliminated.


MrDowntempo:
I tested with Mullvad-Browser as well. It names it correctly but gets no art (but I think that might just be the way mullvad works)

MrDowntempo:
Did a little clean up

MrDowntempo:
I did some cleanup. I don't think playerStateMonitor was there when I did my first commit. And I may have messed up my chance to merge cleanly with it when I force-pushed (I'm still kinda crappy with github) Lemme know I if I can straigten it up or if that's not an issue.

Ly-sec:
Tested it a bit more just to be safe, seems to work just fine. Thank you very much :)!

PR #444, opened 2025-10-09
feat: show calendar events in calendar

ThatOneCalculator:
Loads calendar events from evolution-data-server! Documentation, nix flake, and AUR package need updates to add optional dependencies of `python3` & `evolution-data-server`.

image


MrDowntempo:
I want this so bad, but would be great if we could reduce dependencies and/or support more calendar services

ThatOneCalculator:
I don't think there's any way to reduce dependencies. Most modern Linux installs already have Python, and no extra python modules have to be installed. We could hypothetically swap Python for Perl (since those are the only two languages that have proper ways to interface with evolution-server) but I have literally never used Perl in my life.

MrDowntempo:
Eeww. I'll take python over Perl any day. :D

shouya:
Looks really nice! Does evolution-data-server requires some kind of daemon running? I sync my calendar with vdirsync and view/edit them offline with ikhal/khal/malakal. I would personally prefer a dependency on khal than evolution-server.

Edit: Looks like [dms's calendar implementation](https://github.com/AvengeMedia/DankMaterialShell/blob/master/Services/CalendarService.qml#L66) is based on khal's json output, which may be relevant to this.

ThatOneCalculator:
> Does evolution-data-server requires some kind of daemon running?

Nope!

As for the khal stuff, I've never used khal and I'm not in the KDE ecosystem. I think down the road (as in ***after*** this initial version is merged) there could be multiple adapters for different calendar backends.

Issue #447, opened 2025-10-09
[Feature] Filter out apps from tray


Irubataru:
### Feature Description
I would like it to be possible to filter out certain apps from the tray so they don't show.

### Why Is This Needed?
Is't mostly because the tray shows wifi and bluetooth status, which noctalia already has prettier icons for, meaning they are duplicated.

### Suggested Solutions
That the "Tray" item has a configuration with a list of excluded apps. Potentially also supporting wildcards.

ItsLemmy:
You can probably simply uninstall the "network-manager-applet" package.

ItsLemmy:
Fixed in the main branch, thanks @lonerOrz

Irubataru:
Thanks a lot ^.^

Irubataru:
Did this stop working? I have a blacklist for `dropbox*` due to #513, and it's still there.

Ly-sec:
> Did this stop working? I have a blacklist for `dropbox*` due to [#513](https://github.com/noctalia-dev/noctalia-shell/issues/513), and it's still there.

It must've been a wildcard issue, can you please try [94d1d9d](https://github.com/noctalia-dev/noctalia-shell/commit/94d1d9dc9cd8026cd65c825180b36c5e319bb08e) this commit (or anything afterwards) and see if it works again?

Irubataru:
That also didn't work. It doesn't seem to be a wildcard issue, it also doesn't filter it if I write the full name of the app (`dropbox-client-1909` in this instance). Btw, this is the name I see when I hover over the icon, so I am not guaranteed that it's the name I am looking for. In ps it's just called dropbox, but adding a blacklist for just `dropbox` also doesn't work.

The `dropbox*` used to work, but I can't remember when it stopped.

PR #451, opened 2025-10-09
feat(tray): Implement core blacklist filtering logic

lonerOrz:
fiexd #447

### DIFF
--- a/Modules/Settings/Bar/WidgetSettings/TraySettings.qml
+++ b/Modules/Settings/Bar/WidgetSettings/TraySettings.qml
@@ -25,37 +25,46 @@ ColumnLayout {
 
   spacing: Style.marginM * scaling
 
-  // Input for new blacklist items
-  RowLayout {
+  ColumnLayout {
     Layout.fillWidth: true
     spacing: Style.marginS * scaling
 
-  NTextInput {
-    id: newRuleInput
-    Layout.fillWidth: true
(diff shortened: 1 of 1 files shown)


ItsLemmy:
Hey, thanks for you contribution. Here is a few changes I'd like to see : 
- even if the blacklist is empty by default, I still like to see it populated as empty in the BarWidgetRegistry.qml, as you can see all others widgets define all their settings. It's easier to figure what widget expose which setting. 
- the Tray.qml widget should also be like the others bar widget with custom settings, so it needs, the missing property illustrated in the screenshot.   Also needs to fetch default from widgetMetaData, for the sake of uniformity.
image
- let's avoid duplicate code, you introduced "saveImmediate()" method, then you may as well use it in the  saveTimer (in settings.qml)

lonerOrz:
> a few changes

Done

ItsLemmy:
What's up with the trayBlacklist in settings.qml  ? as there is no UI to edit it, is that only for NixOS users ? 

Also line 40 of Modules/Bar/Widgets/Tray.qml should rather use widgetMetadata as proper fallback
```property list<string> blacklist: widgetSettings.blacklist || Settings.data.bar.trayBlacklist || widgetMetadata.blacklist ```

lonerOrz:
Sorry, that was the global `trayBlacklist` I set up at the beginning to test the blacklist logic 😭

ItsLemmy:
Thanks for your contribution!

PR #452, opened 2025-10-09
updated foot theme for pre-defined color schemes

Sridou:
deprecated: foot: cursor.color: use colors.cursor instead


ItsLemmy:
Thanks!

PR #454, opened 2025-10-09
Battery charging treshold

damian-ds7:
This PR adds a panel when clicking the battery widget that lets you set different charging thresholds. It includes a polkit rule so you don't have to enter your password every time after the initial setup. When quickshell launches, BatteryService.init() writes the last set threshold to make sure it persists after reboot. 

Current state: Still WIP. Need to improve the UI, add an option to change default thresholds, save the last chosen threshold permanently, and expand the battery paths file since it's pretty limited right now (maybe I'll change it up a bit and pass path to the script using nativePath property of UPowerDevice). Open to suggestions on what else should be changed.


damian-ds7:
This is what the feature currently looks like


https://github.com/user-attachments/assets/bd6eda29-d1b5-4a96-a0d6-b7e2d2290ee6

damian-ds7:
@ItsLemmy I think the feature is mostly done, during the installation process, the script checks if any of the paths provided in the file are valid. If none are, the user is informed that the device is not supported. I’ll be adding more paths in the near future.

Would you change anything about the UI or the functionality itself? I’m also considering adding a toggle in the panel that would reset the charging mode to Disabled and set the limit back to 100%. What do you think about that?

Additionally, what are your thoughts on making the thresholds customizable? If you think that would be useful, would it make more sense to add a separate tab in Settings or to manage them directly from the widget settings?

The videos I sent show how the feature behaves both when everything works correctly and when errors occur.

https://github.com/user-attachments/assets/f0d90339-6370-4569-a8f2-e94c7974fd68


https://github.com/user-attachments/assets/440524d8-9f48-42c9-9ae5-6018b4ab4915

ItsLemmy:
It's looking good. I only reviewed the QML/JS so far.
Gonna need a bit more time to try it properly.

ItsLemmy:
To answer your question, I don't think we need an editable threshold. Keep it simple!

damian-ds7:
Ok, now that I think about it I should also add some ipc calls for changing modes, disabling the limit and toggling the panel. An uninstall script would be useful too probably, to easily remove installed script, policy files and logs

damian-ds7:
Do you think there's a point in making a disable switch that would basically do the same thing as setting charging to full capacity? I could add a check in init to only apply the settings when the mode is set to balanced or lifespan, which would make the Disabled state redundant other than for not showing any selected option in the panel. Initially I thought that the disabled state would be useful with customizable thresholds, but without them it's not needed.

I could also keep the state and make a toggle that would install/uninstall the manager from the system and modify the panel so that it clearly says the manager is disabled, similarly to the bt and wifi panels.

damian-ds7:
I added a toggle that will prompt user for a password and remove all files created during installation using a script created during installation and stored in quickshell directory. The script is also write protected so unless someone deletes it with -f flag it should print a warning and let the user do the cleanup manually. After all files are removed successfully a cleanup process is started that deletes the uninstaller script

damian-ds7:
This is how the feature currently works

https://github.com/user-attachments/assets/f1dcaaf1-9473-46f2-af46-a1121bcd82a0

I also added translations for other languages but I used ChatGPT and I'm not familiar with any of them so I can't really verify how good they are.

I think the feature is more or less done

ItsLemmy:
Cool, if you fix the two conflict, I'll merge it.

ItsLemmy:
Thanks a lot

ItsLemmy:
Merged, but I'm having a hard time using it.

When opening the panel (by clicking on battery icon in the bar) I get the battery panel : margins are wrong and it says "Battery manager disabled."

I tried to run manually the install-battery-manager.sh and restart noctalia but it did not make a difference.

damian-ds7:
That's weird, there should be a toggle, and I dunno why the margins are all messed up. I'll take a look at it later

damian-ds7:
Can you try changing chargingMode to 1 in your settings file if you have already installed manager? It should then show mode selection in panel

damian-ds7:
There must have been some change made after I created this branch, that I wasn't aware of. I get the same issue you do when using version from main. I can make a fix a bit later today.

damian-ds7:
I fixed the issue in #486 it should all be fine now. Later I will also send a PR to update the docs

PR #458, opened 2025-10-10
Add DockMenu actions

xqwzts:
* Adds app-specific actions from the DesktopEntry to the context menu / DockMenu
* Only displays actionable actions (ie: does not show Close or Focus if the app is not running, as those would do nothing)

example extra actions for firefox:
image

example extra actions for alacritty:
image

example reduced actions for an app that is not running:
image


ItsLemmy:
Thanks !

PR #460, opened 2025-10-10
feat(Application Launcher): add option to ignore initial mouse position

### LINKED ISSUE #448: Launcher should default to the first item, ignoring initial mouse position
### **Feature Description**

When the launcher is activated, the selected item should always default to the first one in the results list, regardless of the mouse cursor's position on the screen. The selection should only change due to mouse hover *after* the user intentionally moves the mouse, not at the instant the launcher appears.

### **Why Is This Needed?**

The current behavior disrupts a keyboard-centric workflow and leads to unintended actions.

**Problem Scenario:**
1. I activate the launcher with a hotkey, intending to quickly type a query and press `Enter` to launch the first result. This is a very high-frequency workflow.
2. However, if my mouse cursor happens to be resting within the area where the launcher window will appear, it might be hovering over an item that is not the first one (e.g., the third or fourth item in the list).
3. The launcher then incorrectly identifies the item under the cursor as the "currently selected" one.
4. If I don't notice this and press `Enter` immediately as intended, I launch the wrong application entirely.

**This leads to the following problems:**
*   **Reduced Efficiency:** I am forced to consciously check my mouse position or manually press the `Up Arrow` key to move the selection back to the top every time I use the launcher. This defeats the purpose of using a fast, keyboard-driven utility.
*   **Unpredictable Behavior:** The launcher's action becomes unpredictable, as it depends on a factor I am not currently concerned with—the static location of the mouse cursor.
*   **Operational Errors:** It easily leads to mis-launching applications, causing frustration.

The core interaction logic for an excellent launcher should be "keyboard-first."

### **Suggested Solutions**

I propose addressing this at a logical level.

1.  **Force Initial Focus:** After the launcher window appears and the results list is populated, programmatically force the selection state and focus to the first item (index 0).
2.  **Ignore Initial Mouse Events:** Temporarily ignore or defer listening for mouse `hover` or `mouse-enter` events until after the initial focus has been programmatically set. The launcher should only begin responding to mouse movement after this initial state is established.
3.  **(Optional) Add a Configuration Option:** A boolean option could be added to the configuration file, such as `ignore_initial_mouse_hover: true`, to control this behavior. Setting this to `true` by default would provide the best user experience while preserving choice.

### **Additional Context**

This is a behavioral logic issue, so screenshots may not fully demonstrate the problem.

For reference, this is the standard and expected behavior in most popular application launchers, including **Ulauncher**, **Albert**, **Rofi** (in most configurations), as well as macOS's **Alfred** and **Raycast**. They all adhere to the "default to the first item on launch" design principle to ensure a predictable and stable experience for keyboard-driven users.

DuckySoLucky:
## Description

Closes #448

## Preview

https://canary.discord.com/channels/1401598189823590460/1426257862996394035/1426271819211735132 (noctalia-shell discord support thread)

### DIFF
--- a/Assets/settings-default.json
+++ b/Assets/settings-default.json
@@ -1,5 +1,5 @@
 {
-  "settingsVersion": 16,
+  "settingsVersion": 15,
   "bar": {
     "position": "top",
     "backgroundOpacity": 1,
--- a/Commons/Settings.qml
+++ b/Commons/Settings.qml
@@ -125,7 +125,7 @@ Singleton {
   JsonAdapter {
     id: adapter
 
-    property int settingsVersion: 16
+    property int settingsVersion: 15
 
     // bar
     property JsonObject bar: JsonObject {


ItsLemmy:
Thanks a lot!

PR #466, opened 2025-10-11
fix: Alignment issue on current day vs other days

MrDowntempo:
This solves the issues [reported on Discord](https://discord.com/channels/1401598189823590460/1401607843123036162/1425543967310745710) where objects in the top rectangle of the calendar dropdown fromt he time widget are in different vertical positions when on the current day vs any other day.


MrDowntempo:
~~Ack, I should probably make this respect Animation Speed setting. Hold on~~

MrDowntempo:
I'm going to have to tweak this a little more before merge. There's an edge case of September that won't quite fit. Also I think some of the elements could line up a tad nicer.

Ly-sec:
_touches CalendarPanel..._ just kidding :p merging it.

PR #469, opened 2025-10-11
NixOS: split the flake & add app2unit depdencency

ixxie:
This:
- Refactors the flake, breaking up the package, home module and systemd service into seperate files
- Changes to package to use the default qt wrapper rather than a custom wrapper (best practice)
- Adds app2unit to home packages if the user configures the launcher to use it

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -11,287 +11,47 @@
     };
   };
 
-  outputs =
-    {
-      self,
-      nixpkgs,
-      systems,
-      quickshell,
-      ...
-    }:
-    let
-      eachSystem = nixpkgs.lib.genAttrs (import systems);
-    in
-    {
-      formatter = eachSystem (system: nixpkgs.legacyPackages.${system}.alejandra);
+  outputs = {
+    self,
+    nixpkgs,
+    systems,
+    quickshell,
+    ...
+  }: let
+    eachSystem = nixpkgs.lib.genAttrs (import systems);
+  in {
+    formatter = eachSystem (system: nixpkgs.legacyPackages.${system}.alejandra);
 
-      packages = eachSystem (
-        system:
-        let
-          pkgs = nixpkgs.legacyPackages.${system};
-          qs = quickshell.packages.${system}.default.override {
+    packages = eachSystem (
+      system: let
+        pkgs = nixpkgs.legacyPackages.${system};
+      in {
+        default = pkgs.callPackage ./nix/package.nix {
+          version = self.rev or self.dirtyRev or "dirty";
+          quickshell = quickshell.packages.${system}.default.override {
             withX11 = false;
             withI3 = true;
           };
-
-          runtimeDeps =
-            with pkgs;
-            [
-              bash
-              bluez
-              brightnessctl
-              cava
-              cliphist
-              coreutils
-              ddcutil
-              file
-              findutils
-              libnotify
-              matugen
-              networkmanager
-              wlsunset
-              wl-clipboard
-            ]
-            ++ lib.optionals (pkgs.stdenv.hostPlatform.isx86_64) [ gpu-screen-recorder ];
-
-          fontconfig = pkgs.makeFontsConf {
-            fontDirectories = [
-              pkgs.roboto
-              pkgs.inter-nerdfont
-            ];
-          };
-        in
-        {
-          default = pkgs.stdenv.mkDerivation {
-            pname = "noctalia-shell";
-            version = self.rev or self.dirtyRev or "dirty";
-            src = ./.;
-
-            nativeBuildInputs = [
-              pkgs.gcc
-              pkgs.makeWrapper
-              pkgs.qt6.wrapQtAppsHook
-            ];
-            buildInputs = [
-              qs
-              pkgs.xkeyboard_config
-              pkgs.qt6.qtbase
-            ];
-            propagatedBuildInputs = runtimeDeps;
-
-            installPhase = ''
-              mkdir -p $out/share/noctalia-shell
-              cp -r ./* $out/share/noctalia-shell
-
-              makeWrapper ${qs}/bin/qs $out/bin/noctalia-shell \
-                --prefix PATH : "${pkgs.lib.makeBinPath runtimeDeps}" \
-                --set FONTCONFIG_FILE "${fontconfig}" \
-                --add-flags "-p $out/share/noctalia-shell"
-            '';
-
-            meta = {
-              description = "A sleek and minimal desktop shell thoughtfully crafted for Wayland, built with Quickshell.";
-              homepage = "https://github.com/noctalia-dev/noctalia-shell";
-              license = pkgs.lib.licenses.mit;
-              mainProgram = "noctalia-shell";
-            };
-          };
-        }
-      );
-
-      defaultPackage = eachSystem (system: self.packages.${system}.default);
-
-      homeModules.default =
-        {
-          config,
-          lib,
-          pkgs,
-          ...
-        }:
-        let
-          cfg = config.programs.noctalia-shell;
-          defaultSettings = builtins.fromJSON (builtins.readFile ./Assets/settings-default.json);
-
-          # Deep merge user settings with defaults
-          mergedSettings =
-            if cfg.settings == null then
-              defaultSettings
-            else if builtins.isAttrs cfg.settings then
-              lib.recursiveUpdate defaultSettings cfg.settings
-            else
-              cfg.settings; # Pass through strings/paths as-is
-        in
-        {
-          options.programs.noctalia-shell = {
-            enable = lib.mkEnableOption "Noctalia shell configuration";
-
-            settings = lib.mkOption {
-              type =
-                with lib.types;
-                nullOr (oneOf [
-                  attrs
-                  str
-                  path
-                ]);
-              default = null;
-              example = lib.literalExpression ''
-                {
-                  bar = {
-                    position = "bottom";
-                    floating = true;
-                    backgroundOpacity = 0.95;
-                  };
-                  general = {
-                    animationSpeed = 1.5;
-                    radiusRatio = 1.2;
-                  };
-                  colorSchemes = {
-                    darkMode = true;
-                    useWallpaperColors = true;
-                  };
-                }
-              '';
-              description = ''
-                Noctalia shell configuration settings as an attribute set, string
-                or filepath, to be written to ~/.config/noctalia/settings.json.
-                When provided as an attribute set, it will be deep-merged with
-                the default settings.
-              '';
-            };
-
-            colors = lib.mkOption {
-              type =
-                with lib.types;
-                nullOr (oneOf [
-                  attrs
-                  str
-                  path
-                ]);
-              default = null;
-              example = lib.literalExpression ''
-                 {
-                   mError = "#dddddd";
-                   mOnError = "#111111";
-                   mOnPrimary = "#111111";
-                   mOnSecondary = "#111111";
-                   mOnSurface = "#828282";
-                   mOnSurfaceVariant = "#5d5d5d";
-                   mOnTertiary = "#111111";
-                   mOutline = "#3c3c3c";
-                   mPrimary = "#aaaaaa";
-                   mSecondary = "#a7a7a7";
-                   mShadow = "#000000";
-                   mSurface = "#111111";
-                   mSurfaceVariant = "#191919";
-                   mTertiary = "#cccccc";
-                }
-              '';
-              description = ''
-                Noctalia shell color configuration as an attribute set, string
-                or filepath, to be written to ~/.config/noctalia/colors.json.
-              '';
-            };
-          };
-
-          config =
-            let
-              restart = ''
-                ${pkgs.systemd}/bin/systemctl --user try-restart noctalia-shell.service 2>/dev/null || true
-              '';
-              useApp2Unit = mergedSettings.appLauncher.useApp2Unit or false;
-            in
-            lib.mkIf cfg.enable {
-              home.packages = lib.optional useApp2Unit pkgs.app2unit;
-
-              xdg.configFile = {
-                "noctalia/settings.json" = {
-                  onChange = restart;
-                }
-                // (
-                  if builtins.isAttrs mergedSettings then
-                    { text = builtins.toJSON mergedSettings + "\n"; }
-                  else if builtins.isString mergedSettings then
-                    { text = mergedSettings; }
-                  else
-                    { source = mergedSettings; }
-                );
-                "noctalia/colors.json" = lib.mkIf (cfg.colors != null) (
-                  {
-                    onChange = restart;
-                  }
-                  // (
-                    if builtins.isAttrs cfg.colors then
-                      { text = builtins.toJSON cfg.colors; }
-                    else if builtins.isString cfg.colors then
-                      { text = cfg.colors; }
-                    else
-                      { source = cfg.colors; }
-                  )
-                );
-              };
-            };
         };
+      }
+    );
 
-      nixosModules.default =
-        {
-          config,
-          lib,
-          pkgs,
-          ...
-        }:
-        let
-          cfg = config.services.noctalia-shell;
-        in
-        {
-          options.services.noctalia-shell = {
-            enable = lib.mkEnableOption "Noctalia shell systemd service";
+    defaultPackage = eachSystem (system: self.packages.${system}.default);
 
-            package = lib.mkOption {
-              type = lib.types.package;
-              default = self.packages.${pkgs.system}.default;
-              description = "The noctalia-shell package to use";
-            };
-
-            target = lib.mkOption {
-              type = lib.types.str;
-              default = "graphical-session.target";
-              example = "hyprland-session.target";
-              description = "The systemd target for the noctalia-shell service.";
-            };
-          };
-
-          config = lib.mkIf cfg.enable {
-            systemd.user.services.noctalia-shell = {
-              description = "Noctalia Shell - Wayland desktop shell";
-              documentation = [ "https://github.com/noctalia-dev/noctalia-shell" ];
-              after = [ cfg.target ];
-              partOf = [ cfg.target ];
-              wantedBy = [ cfg.target ];
-              restartTriggers = [ cfg.package ];
-
-              environment = {
-                PATH = lib.mkForce null;
-              };
-
-              unitConfig = {
-                StartLimitIntervalSec = 60;
-                StartLimitBurst = 3;
-              };
-
-              serviceConfig = {
-                ExecStart = "${cfg.package}/bin/noctalia-shell";
-                Restart = "on-failure";
-                RestartSec = 3;
-                TimeoutStartSec = 10;
-                TimeoutStopSec = 5;
-                Environment = [
-                  "NOCTALIA_SETTINGS_FALLBACK=%h/.config/noctalia/gui-settings.json"
-                ];
-              };
-            };
-
-            environment.systemPackages = [ cfg.package ];
-          };
-        };
+    homeModules.default = {
+      pkgs,
+      lib,
+      ...
+    }: {
+      imports = [./nix/home-module.nix];
+      programs.noctalia-shell.app2unit.package =
+        lib.mkDefault
+        nixpkgs.legacyPackages.${pkgs.system}.app2unit;
     };
+
+    nixosModules.default = {pkgs, ...}: {
+      imports = [./nix/nixos-module.nix];
+      services.noctalia-shell.package = self.packages.${pkgs.system}.default;
+    };
+  };
 }
--- a/nix/home-module.nix
+++ b/nix/home-module.nix
@@ -1,0 +1,114 @@
+{
+  config,
+  lib,
+  pkgs,
+  ...
+}: let
+  cfg = config.programs.noctalia-shell;
+  defaultSettings = builtins.fromJSON (builtins.readFile ../Assets/settings-default.json);
+  extractAttrs = x:
+    if builtins.isAttrs x
+    then x
+    else if builtins.isString x
+    then builtins.fromJson x
+    else builtins.fromJson (builtins.readFile x);
+in {
+  options.programs.noctalia-shell = {
+    enable = lib.mkEnableOption "Noctalia shell configuration";
+
+    settings = lib.mkOption {
+      type = with lib.types;
+        nullOr (oneOf [
+          attrs
+          str
+          path
+        ]);
+      default = {};
+      apply = x: lib.recursiveUpdate defaultSettings (extractAttrs x);
+      example = lib.literalExpression ''
+        {
+          bar = {
+            position = "bottom";
+            floating = true;
+            backgroundOpacity = 0.95;
+          };
+          general = {
+            animationSpeed = 1.5;
+            radiusRatio = 1.2;
+          };
+          colorSchemes = {
+            darkMode = true;
+            useWallpaperColors = true;
+          };
+        }
+      '';
+      description = ''
+        Noctalia shell configuration settings as an attribute set, string
+        or filepath, to be written to ~/.config/noctalia/settings.json.
+        When provided as an attribute set, it will be deep-merged with
+        the default settings.
+      '';
+    };
+
+    colors = lib.mkOption {
+      type = with lib.types;
+        nullOr (oneOf [
+          attrs
+          str
+          path
+        ]);
+      default = {};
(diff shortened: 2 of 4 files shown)


nix/package.nix
```diff
@@ -0,0 +1,76 @@
...
+  self,
+  pkgs,
+  quickshell,
+  system,
+}:
+let
+  qs = quickshell.packages.${system}.default.override {
+    withX11 = false;
+    withI3 = true;
+  };
```
Rexcrazy804: pass in the overrided quickshell drv from the flake.
this obstructs end user's ability to override this derrivation

nix/package.nix
```diff
@@ -0,0 +1,76 @@
...
+      ddcutil
+      file
+      findutils
+      libnotify
+      matugen
+      networkmanager
+      wlsunset
+      wl-clipboard
+    ]
+    ++ lib.optionals (pkgs.stdenv.hostPlatform.isx86_64) [ gpu-screen-recorder ];
```
Rexcrazy804: Don't use pkgs.
Individually include each dependency at the top.

Using pkgs directly breaks the everything the `callPackages` function was made to handle, please avoid this pattern.

nix/package.nix
```diff
@@ -0,0 +1,76 @@
...
+
+  fontconfig = pkgs.makeFontsConf {
+    fontDirectories = [
+      pkgs.roboto
+      pkgs.inter-nerdfont
+    ];
+  };
```
Rexcrazy804: same as above

nix/package.nix
```diff
@@ -0,0 +1,76 @@
...
+in
+pkgs.stdenv.mkDerivation {
+  pname = "noctalia-shell";
+  version = self.rev or self.dirtyRev or "dirty";
+  src = ../.;
```
Rexcrazy804: use lib.cleanSource or even better use lib.filesets like [stash/nix/package.nix](https://github.com/NotAShelf/stash/blob/a59e207e76d48c6d016c7b8fe13234d864da6532/nix/package.nix#L11-L18)

nix/package.nix
```diff
@@ -0,0 +1,76 @@
...
+  fontconfig = pkgs.makeFontsConf {
+    fontDirectories = [
+      pkgs.roboto
+      pkgs.inter-nerdfont
+    ];
+  };
+in
+pkgs.stdenv.mkDerivation {
+  pname = "noctalia-shell";
+  version = self.rev or self.dirtyRev or "dirty";
```
Rexcrazy804: have the rev as an attribute at the top and define it on the flake.
Don't pass in self.
use stdenv directly (same as above, I'll call it noPKGS hereon)

nix/package.nix
```diff
@@ -0,0 +1,76 @@
...
+
+  nativeBuildInputs = [
+    pkgs.gcc
+    pkgs.qt6.wrapQtAppsHook
+  ];
+  buildInputs = [
+    qs
+    pkgs.xkeyboard_config
+    pkgs.qt6.qtbase
+  ];
```
Rexcrazy804: noPKGS

nix/package.nix
```diff
@@ -0,0 +1,76 @@
...
+    cp ${qs}/bin/qs $out/bin/noctalia-shell
+  '';
+
+  preFixup = ''
+    qtWrapperArgs+=(
+      --prefix PATH : ${pkgs.lib.makeBinPath runtimeDeps}
+      --set FONTCONFIG_FILE ${fontconfig}
+      --add-flags "-p $out/share/noctalia-shell"
+    )
+  '';
```
Rexcrazy804: you are better off using symlinkJoin instead of stdenv.mkDerrivation. See [zaphkiel/pkgs/kurukurubar.nix](https://github.com/Rexcrazy804/Zaphkiel/blob/bfdcc2eaa3167172bc98eef19bb0b104f13fab48/pkgs/kurukurubar.nix#L69-L89)

nix/package.nix
```diff
@@ -0,0 +1,76 @@
...
+
+  meta = {
+    description = "A sleek and minimal desktop shell thoughtfully crafted for Wayland, built with Quickshell.";
+    homepage = "https://github.com/noctalia-dev/noctalia-shell";
+    license = pkgs.lib.licenses.mit;
```
Rexcrazy804: noPKGS

nix/nixos-module.nix
```diff
@@ -0,0 +1,61 @@
+self:
```
Rexcrazy804: Avoid patterns like these.
Here, all you need self for is the default package.
See how this is handled in [lanzaboote/flake.nix](https://github.com/nix-community/lanzaboote/blob/747b7912f49e2885090c83364d88cf853a020ac1/flake.nix#L44-L50)

In the future if a module depends on an extra argset use lib.importApply to pass such args, this will avoid cryptic error messages (if there are errors) due to unconventional import.

nix/nixos-module.nix
```diff
@@ -0,0 +1,61 @@
...
+      default = self.packages.${pkgs.system}.default;
+      description = "The noctalia-shell package to use";
+    };
+
+    target = lib.mkOption {
+      type = lib.types.str;
+      default = "graphical-session.target";
+      example = "hyprland-session.target";
+      description = "The systemd target for the noctalia-shell service.";
+    };
```
Rexcrazy804, line 21: Redundant.
If the user wants to override this, they can do so directly by using lib.mkForce on the systemd service.

flake.nix
```diff
@@ -29,266 +29,18 @@
...
-            '';
-
-            meta = {
-              description = "A sleek and minimal desktop shell thoughtfully crafted for Wayland, built with Quickshell.";
-              homepage = "https://github.com/noctalia-dev/noctalia-shell";
-              license = pkgs.lib.licenses.mit;
-              mainProgram = "noctalia-shell";
-            };
+          default = pkgs.callPackage ./nix/package.nix {
+            inherit self quickshell system;
```
Rexcrazy804: avoid passing flakes as arguments to a derrivation.
cleanly only pass in the exact attribute/package as required.
see the reveiw in package.nix

flake.nix
```diff
@@ -29,266 +29,18 @@
...
-        {
-          options.services.noctalia-shell = {
-            enable = lib.mkEnableOption "Noctalia shell systemd service";
-
-            package = lib.mkOption {
-              type = lib.types.package;
-              default = self.packages.${pkgs.system}.default;
-              description = "The noctalia-shell package to use";
-            };
+      homeModules.default = import ./nix/home-module.nix;
```
Rexcrazy804: you don't have to use import here, you can directly use the path

flake.nix
```diff
@@ -29,266 +29,18 @@
...
-                Environment = [
-                  "NOCTALIA_SETTINGS_FALLBACK=%h/.config/noctalia/gui-settings.json"
-                ];
-              };
-            };
-
-            environment.systemPackages = [ cfg.package ];
-          };
-        };
+      nixosModules.default = import ./nix/nixos-module.nix self;
```
Rexcrazy804: avoid passing in self, see the review on nixos-module.nix.

nix/home-module.nix
Rexcrazy804, line 1: I haven't looked to closely at this file,
but I can clearly see something is done poorly here due to repeated if and else ifs.
See if you can improve it, if not mention it and I will take a closer look.

nix/package.nix
ixxie: done

nix/package.nix
ixxie: done

nix/package.nix
ixxie: done

nix/package.nix
ixxie: Done, although this is very new to me and I would love it if you checked this bit especially to make sure I did it right

nix/package.nix
ixxie: done

nix/nixos-module.nix
ixxie: This is true, but many of our users don't feel comfortable with doing things like that in their configs. I'm going to opt to leave this in to give them a simpler interface.

flake.nix
ixxie: done

flake.nix
ixxie: done

flake.nix
ixxie: done, but I'm a bit unsure if the solution is idiomatic.

nix/home-module.nix
ixxie: Made a reusable function for the repeated logic. I'm not sure what else I could do to improve it

ixxie:
> Heyya, lysec had asked me to review your nix code / Do go through my reviews and let me know if anything needs further clarification

@Rexcrazy804 thank you for taking the time to review this! To give a bit of context:

- I've used NixOS for years but never got deep enough into the language / ecosystem to grasp a lot of the patterns used; you can say I have been more an end-user rather than a developer.
- I've inherited a lot of this flake code and only recently started contributing (the systemd service and home-module are my main contributions)

I'm grateful for your review; I already learned some interesting things. I think I addressed most of your comments, but I'm unsure of some of the solutions.

nix/home-module.nix
```diff
@@ -0,0 +1,121 @@
...
+      '';
+    };
+  };
+
+  config =
+    let
+      restart = ''
+        ${pkgs.systemd}/bin/systemctl --user try-restart noctalia-shell.service 2>/dev/null || true
+      '';
+      useApp2Unit = mergedSettings.appLauncher.useApp2Unit or false;
```
cdata: My config has `useApp2Unit` set to true, but when I updated it immediately failed to build, complaining:

```
       error: attribute 'app2unit' missing
       at /nix/store/4hs6cv7501yf66jrxfix05chh819f7dm-source/nix/home-module.nix:107:48:
          106|     lib.mkIf cfg.enable {
          107|       home.packages = lib.optional useApp2Unit pkgs.app2unit;
             |                                                ^
          108|
       Did you mean cppunit?
```

I dug a little deeper, and the culprit turned out to be that I recently standardized my config on `25.05` and `app2unit` is only available on `unstable`. I pull in `unstable` so that I can install the occasional thing like `app2unit` (but it is under the name `pkgsUnstable`).

Anyway, for this to work for me I would need to be able to configure the `app2unit` package to be used.

ixxie: unstable is required for Noctalia and documented

nix/package.nix
```diff
@@ -0,0 +1,110 @@
...
+      ../Bin
+      ../Commons
+      ../Helpers
+      ../Modules
+      ../Services
+      ../Shaders
+      ../Widgets
+      ../shell.qml
+      ../LICENSE
+      ../README.md
```
Rexcrazy804: Seems like you've overlooked the point of doing this in the first place so I will elaborate.
Essentially what we are doing is narrowing down the exact set of files (fileset) required to build our derrivation. That is to say, if file X is irrelevant for building our derrivation, it won't be included in the final fileset. Thus, in this case, LICENSE and README.md are not required for building our derrivation.

The primary reason we do this is to avoid needless rebuilds, i.e. changing the README.md file will never (and should never) cause the package derivation to be rebuilt, that would be silly. 

you could use the following
```nix
 (fs.fileFilter (file: builtins.any file.hasExt ["qs"]) ../.)
```
to quickly include every qs file in the repo.
and then declare the inclusion of the scripts folder.

nix/package.nix
Rexcrazy804, line 1: Besides the improper use of filesets this looks good to go.
Great work :)

flake.nix
```diff
@@ -29,266 +29,24 @@
...
-                ];
-              };
-            };
-
-            environment.systemPackages = [ cfg.package ];
-          };
-        };
+      nixosModules.default = import ./nix/nixos-module.nix {
+        packages = self.packages;
+      };
```
Rexcrazy804: remember how for the package we passed in exactly what the needed? i.e. `quickshell.packages.${system}.quickshell`. The same applies here.

However this isn't what I had intended by citing how the package was sent in in the lanzaboote module.
We're looking at something like this

```nix
nixosModules.default  = {pkgs, ...}: {
  imports = [./nix/nixos-module.nix];
  services.noctalia-shell.package = self.packages.${pkgs.system}.default;
};
```

> you must remove the default = XXXX line in the package field of the nixos options.

This way we can remove the module needing an extra attribute set up top, ensuring its independence of the flake.

nix/nixos-module.nix
```diff
@@ -0,0 +1,63 @@
+{
+  packages,
+}:
```
Rexcrazy804: get rid of this, see review in flake.nix

nix/nixos-module.nix
```diff
@@ -0,0 +1,63 @@
...
+let
+  cfg = config.services.noctalia-shell;
+in
+{
+  options.services.noctalia-shell = {
+    enable = lib.mkEnableOption "Noctalia shell systemd service";
+
+    package = lib.mkOption {
+      type = lib.types.package;
+      default = packages.${pkgs.system}.default;
```
Rexcrazy804: we'll needa remove this default line, it will be set on the flake.

nix/home-module.nix
Rexcrazy804: Allow me a closer look at what this is doing, I'll follow up soon.

Rexcrazy804:
> I'm grateful for your review; I already learned some interesting things. I think I addressed most of your comments, but I'm unsure of some of the solutions

Everyone starts somewhere, feel welcome to ask me to elaborate on anything that you are uncertain of.

Rexcrazy804:
Here is my take on the home-module.nix. To summarize:

- used `options.apply` attribute
- introduced `extractAttr` to accept `attrset|string|path` to yeild an attrset.
  gives us the advantage of getting nix errors for invalid json passed in as a string or a file by the user
  as a side effect of forcing it through fromJson
- changed default values from `null` to `{}`

haven't tested this since I don't use home manager, so let me know if there's anything that doesn't work.
```diff
diff --git a/nix/home-module.nix b/nix/home-module.nix
index 05fadd93..90ddae41 100644
--- a/nix/home-module.nix
+++ b/nix/home-module.nix
@@ -7,22 +7,10 @@
 let
   cfg = config.programs.noctalia-shell;
   defaultSettings = builtins.fromJSON (builtins.readFile ../Assets/settings-default.json);
-
-  mergedSettings =
-    if cfg.settings == null then
-      defaultSettings
-    else if builtins.isAttrs cfg.settings then
-      lib.recursiveUpdate defaultSettings cfg.settings
-    else
-      cfg.settings;
-
-  toConfigFile = value:
-    if builtins.isAttrs value then
-      { text = builtins.toJSON value + "\n"; }
-    else if builtins.isString value then
-      { text = value; }
-    else
-      { source = value; };
+  extractAttrs = x: 
+    if builtins.isAttrs x then x
+    else if builtins.isString x then builtins.fromJson x
+    else builtins.fromJson (builtins.readFile x);
 in
 {
   options.programs.noctalia-shell = {
@@ -36,7 +24,8 @@ in
           str
           path
         ]);
-      default = null;
+      default = {};
+      apply = x: lib.recursiveUpdate defaultSettings (extractAttrs x);
       example = lib.literalExpression ''
         {
           bar = {
@@ -70,7 +59,7 @@ in
           str
           path
         ]);
-      default = null;
+      default = {};
+      apply = extractAttrs
       example = lib.literalExpression ''
          {
            mError = "#dddddd";
@@ -101,7 +90,7 @@ in
       restart = ''
         ${pkgs.systemd}/bin/systemctl --user try-restart noctalia-shell.service 2>/dev/null || true
       '';
-      useApp2Unit = mergedSettings.appLauncher.useApp2Unit or false;
+      useApp2Unit = cfg.settings.appLauncher.useApp2Unit or false;
     in
     lib.mkIf cfg.enable {
       home.packages = lib.optional useApp2Unit pkgs.app2unit;
@@ -109,13 +98,12 @@ in
       xdg.configFile = {
         "noctalia/settings.json" = {
           onChange = restart;
-        } // toConfigFile mergedSettings;
-
-        "noctalia/colors.json" = lib.mkIf (cfg.colors != null) (
-          {
-            onChange = restart;
-          } // toConfigFile cfg.colors
-        );
+          text = builtins.toJson cfg.settings;
+        };
+        "noctalia/colors.json" = lib.mkIf (cfg.colors != {}) {
+          onChange = restart;
+          text = builtins.toJson cfg.colors;
+        };
       };
     };
 }
```
use the following to quickly apply this diff on your worktree (ofc copy the above first)
```sh
wl-paste | git apply
```

nix/home-module.nix
cdata: I was able to change Noctalia to track `nixos-unstable`, but actually getting this to work was another matter. This build failure was happening in a separate Home Manager-specific module, and by default it was receiving the channel that is configured for my Home Manager (not the one that is tracked by the Noctalia input in general). So, in order to build successfully, I had to override like so:

```diff
   home-manager.users.${user.username} = { config, ... }: {
     # import the home manager module
     imports = [
-      inputs.noctalia.homeModules.default
+      # NOTE: Noctalia depends on `nixos-unstable` at the time of this configuration
+      # In other places we can track the unstable channel automatically, but here we
+      # need to explicitly override `pkgs`.
+      ({ ... } @ args:
+        let
+          noctalia = import inputs.noctalia.homeModules.default (args // {
+            pkgs = pkgsUnstable;
+          });
+        in
+        noctalia)
     ];
```

Ideally it would be sufficient to change _only_ which channel is tracked by the Noctalia input. It might be possible to do this if `homeModules.default` was a function that closed over the channel provided to the outer flake.

nix/package.nix
ixxie: The files in this repo are mostly `qml` files but there are several other file types. Since I don't really know what to expect, I would rather work by exclusion since I feel it will be more robust.

I tried to find a way to do this with `fileset` but had no luck. Instead, this is what I came up with:
```nix
  src = lib.cleanSourceWith {
    src = ../.;
    filter =
      path: type:
      let
        relPath = lib.removePrefix (toString ../. + "/") path;
      in
      !(builtins.any (prefix: lib.hasPrefix prefix relPath) [
        "Assets/Screenshots"
        "Assets/Wallpaper"
        "Bin/dev"
        "LICENSE"
        "README.md"
        ".git"
        ".github"
      ]);
  };
```

Rexcrazy804: remove the conversion to string and use lib.paths.hasPrefix

ixxie:
> Here is my take on the home-module.nix. To summarize:
> 
>     * used `options.apply` attribute
> 
>     * introduced `extractAttr` to accept `attrset|string|path` to yeild an attrset.
>       gives us the advantage of getting nix errors for invalid json passed in as a string or a file by the user
>       as a side effect of forcing it through fromJson
> 
>     * changed default values from `null` to `{}`
> 
> 
> haven't tested this since I don't use home manager, so let me know if there's anything that doesn't work.

Your patch was corrupt for some reason, but I managed to apply it with an LLM. There was a little mistake (`builtins.toJSON` not `builtins.toJson`) but otherwise it worked great and I like the simplification.

nix/package.nix
ixxie: > remove the conversion to string and use lib.paths.hasPrefix

```nix
  src = lib.cleanSourceWith {
    src = ../.;
    filter =
      path: type:
      !(builtins.any (prefix: lib.path.hasPrefix (../. + prefix) (/. + path)) [
        /Assets/Screenshots
        /Assets/Wallpaper
        /Bin/dev
        /LICENSE
        /README.md
        /.git
        /.github
      ]);
  };
```
something like this?

ixxie:
@Rexcrazy804 I guess everything is good to go now, assuming https://github.com/noctalia-dev/noctalia-shell/pull/469#discussion_r2427273719 is ok

nix/package.nix
```diff
@@ -0,0 +1,107 @@
...
+  makeFontsConf,
+  roboto,
+  inter-nerdfont,
+}:
+let
+  src = lib.cleanSourceWith {
+    src = ../.;
+    filter =
+      path: type:
+      !(builtins.any (prefix: lib.path.hasPrefix (../. + prefix) (/. + path)) [
```
Rexcrazy804, line 34: `/. + path`?  looks odd. Are you sure it works as intended? If yes, then we can keep it as is

Rexcrazy804:
> @Rexcrazy804 I guess everything is good to go now, assuming [#469 (comment)](https://github.com/noctalia-dev/noctalia-shell/pull/469#discussion_r2427273719) is ok

have you tested that it builds correctly? thr (/. + path) looks wrong

nix/package.nix
ixxie: > `/. + path`? looks odd. Are you sure it works as intended? If yes, then we can keep it as is

I thought it was weird too... its Claude suggestion for how to cast `path` (which was apparently a string) to a path. I'm very open to cleaner approaches, I just didn't know how to fix the type error otherwise.

ixxie: I'm not really sure how to validate it works as intended though :hankey:

nix/home-module.nix
ixxie: > Ideally it would be sufficient to change _only_ which channel is tracked by the Noctalia input. It might be possible to do this if `homeModules.default` was a function that closed over the channel provided to the outer flake.

Something like this: https://github.com/noctalia-dev/noctalia-shell/pull/469/commits/6eca74f31827ae4efa3875b375e4b0b9de02cfad

@Rexcrazy804 what do you think of this? Makes sense?

flake.nix
```diff
@@ -43,7 +43,9 @@
 
       defaultPackage = eachSystem (system: self.packages.${system}.default);
 
-      homeModules.default = ./nix/home-module.nix;
+      homeModules.default = { pkgs, ... }@args: import ./nix/home-module.nix (args // {
```
Rexcrazy804: does home manager not support imports list? this looks finicky

nix/package.nix
Rexcrazy804: I'll build and validate tomorrow, its a little late so good ight for today /

flake.nix
ixxie: > does home manager not support imports list? this looks finicky

I'm not sure what you mean... it might but I'm kind of shooting in the dark here with search engines / llms.

nix/package.nix
ixxie: At least it builds with `nix build` and runs on my system (I test by importing the flake locally).

Rexcrazy804: image

seems like its doing its job.
include nix/, flake.nix, flake.lock to the exclusion list.
and remove ".git" (its not really included anyways)

flake.nix
Rexcrazy804: I've figured out that you are effectively doing this to just pass in the app2unit package from the nipxkgs pin used by the noctalia flake.

The better workaround would be creating an `app2unit.package` under `programs.noctalia-shell` options.
this way we can just do 

```nix
{
  homeModules.default = { pkgs, lib, ... }: {
    imports = [./nix/home-module.nix];
    # ensures app2unit package based on noctalia's nixpkgs pin
    programs.noctalia-shell.app2unit.package = lib.mkDefault nixpkgs.legacyPackages.${pkgs.system}.app2unit;
  };
}
```

ixxie:
@Rexcrazy804 I think we are done!

- [x] `home-module.nix` `@args` removed in place of `app2unit.package`
- [x] Fixed `cleanSourceWith` paths to remove `.git` and add `nix` stuff
- [x] Formatted with `alejandra` (and made it the default in my personal editor)

nix/home-module.nix
```diff
@@ -0,0 +1,115 @@
...
+      '';
+      description = ''
+        Noctalia shell color configuration as an attribute set, string
+        or filepath, to be written to ~/.config/noctalia/colors.json.
+      '';
+    };
+
+    app2unit.package = lib.mkOption {
+      type = lib.types.package;
+      default = pkgs.app2unit;
```
Rexcrazy804: needa remove the default from here as we did with the nix-module

s0me1newithhand7s:
> * [x]  Formatted with `alejandra` (and made it the default in my personal editor)

i suggest adding `formatter` in `flake.nix` to use `nix fmt`!

ixxie:
> > * [x]   Formatted with `alejandra` (and made it the default in my personal editor)
> 
> i suggest adding `formatter` in `flake.nix` to use `nix fmt`!

Its already in the flake! It was before the PR...

ixxie:
Merged! Thanks for the review @Rexcrazy804! Very much appreciated.

PR #470, opened 2025-10-11
fix(bar): Correct ActiveWindow layout when empty

lonerOrz:
Before fixed：
图片

### DIFF
--- a/Assets/Translations/de.json
+++ b/Assets/Translations/de.json
@@ -906,9 +906,9 @@
         "search-placeholder": "Widgets suchen..."
       },
       "active-window": {
-        "auto-hide": {
-          "label": "Automatisch ausblenden",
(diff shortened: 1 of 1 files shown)


ItsLemmy:
I'd like to suggest an improvement for how the widget behaves when no window is active.

Currently, it remains visible (but transparent) to prevent other widgets from shifting around, but this can be undesirable depending on the user's layout. To provide more flexibility, I propose we convert the HideAutomatically toggle into a dropdown menu with three choices:

- Always Visible: The widget is always shown.
- Hide When Empty: The widget disappears, allowing other elements to shift. (This would be the new default).
- Transparent When Empty: The widget becomes invisible but still occupies its space to prevent layout shifts.

To maintain consistency, the MediaMini widget should receive the same update.

What do you think of this approach?

lonerOrz:
I think it’s great, let users choose for themselves!

PR #471, opened 2025-10-11
add option for auto-hiding bars

dwuggh:
- Adds an auto‑hide mode for the top/left/right/bottom bar, the bar overlays other windows if not hided;
- Reveal is triggered by a 1-pixel “peek” strip at the bar’s screen edge;
- Show/hide animates with a mix of slide and fade;
- respects the global disbale animation option.

This pr is made by GPT-5, but I reviewed the code manually, though I'm new to QML.

I've tested all 4 directions , with/without animations, on niri. I may not be able to test on hyprland because it does not function well with nvidia GPUs till the last time I try it.


Ly-sec:
This looks quite promising but can you please use our existing i18n implementation? It is totally fine to generate the translation for now.

dwuggh:
> This looks quite promising but can you please use our existing i18n implementation? It is totally fine to generate the translation for now.

done, I manually translated zh-CN and use AI translation for the else. Is directly writing to xx.json is what you mean by generate? I also didn't test it because I failed to find sth like "switch language" option, or is the language determined by `LANG`?

Ly-sec:
Yup you did it correctly :) I'll test it in a bit. You can test different languages by changing the `i18n.qml` debug language if need be, otherwise it's defined by our `LOCALE`.

dwuggh:
thanks!

dwuggh:
I've tested with chinese, for bar it is fine, but note that `Dock` didn't get translated in chinese. Could it be intentional because we sometimes just say "dock" or "dock 栏" even in chinese context. it could also be "程序栏" imo.

For this PR, I've been using it for 2 days, and I think it is ready for merge now, @Ly-sec  is there any other works to do?

ItsLemmy:
Had to revert it.
- Minor issue:  does not play well with Screen Corners. They stay there and without the bar it looks very buggy. I guess we could just force screen corners to off when auto hide is on.
- MAJOR issue:  when switching the bar position (left, right. top, bottom)  Quickshell crashes.

PR #472, opened 2025-10-11
fix: Resolve intermittent OSD brightness display issue

lonerOrz:
* Ensuring the brightness OSD consistently appears for every adjustment.
* Providing accurate brightness feedback for the specific monitor being controlled, especially in multi-monitor configurations.


Ly-sec:
Thank you very much!

PR #479, opened 2025-10-14
Workspace: improved style of active but not focused workspaces

EdenQwQ:
This PR improves style of active but not focused workspace (active workspace of an unfocused monitor) on the bar.
It is more intuitive to use a larger pill for all active workspace, as it shows you the active workspace of an unfocused monitor.
Check the screenshots for comparison (above is new, below is old).

When a workspace (the first one) is occupied and active but not focused:
Screenshot from 2025-10-14 22-50-09

When a workspace (the third one) is empty and active but not focused:
Screenshot from 2025-10-14 22-50-18


ItsLemmy:
Excellent, this is indeed much better. Thanks!

PR #480, opened 2025-10-14
feat(taskbar): add auto hide mode functionality for taskbar widget

Sighthesia:
I added auto hide function like other widgets, so the taskbar widgets could hide when it is empty without leaving background and taking up space.
image

It looks ok, please let me know if changes are needed, thanks a lot!

### DIFF
--- a/Modules/Bar/Widgets/Taskbar.qml
+++ b/Modules/Bar/Widgets/Taskbar.qml
@@ -34,42 +34,45 @@ Rectangle {
     return {}
   }
 
-  readonly property string hideMode: (widgetSettings.hideMode !== undefined) ? widgetSettings.hideMode : (widgetMetadata.hideMode !== undefined ? widgetMetadata.hideMode : "hidden")
+  readonly property string hideMode: (widgetSettings.hideMode !== undefined) ? widgetSettings.hideMode : widgetMetadata.hideMode
 
-  property int __filteredCount: 0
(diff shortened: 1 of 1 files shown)


ItsLemmy:
Thanks a lot  for the contribution. I'm fine with the feature but would like to see a few changes before merging : 
- What's up with all the underscores ?  can you try to follow our code style a bit closer.
- If the filteredCount is only used has a boolean then maybe rename it to something like property bool hasWindow.    And change the code to return as soon as you found one to save some unnecessary computation.

Also this would be better if it was reversed so rather test for !== undefined  and assign the settings and fall back to metaData (default).  Doing so will follow the existing code (in this widgets and the others)  and make easier to read and maintain.
```  
readonly property bool __onlySameOutput: widgetSettings.onlySameOutput === undefined ? widgetMetadata.onlySameOutput : widgetSettings.onlySameOutput  

readonly property bool __onlyActiveWorkspaces: widgetSettings.onlyActiveWorkspaces === undefined ? widgetMetadata.onlyActiveWorkspaces : widgetSettings.onlyActiveWorkspaces
```

Thanks again!

Sighthesia:
oh sorry for my mistake that i forgot to remove the underscores flag, and very thanks for you comprehensive guidances, i have done:
1. removed leading underscores.
2. renamed properties like `filteredCount` to make them move understandable.
3. short‑circuited window detection logic.
4. unify the logical operator of output and workspace checks.

ItsLemmy:
Thanks!

PR #482, opened 2025-10-14
Implement scheduled light/dark mode switching

### LINKED ISSUE #493: [Feature] Auto-switch dark/light mode based on system location
### Feature Description
Enable automatic switching between Dark Mode and Light Mode based on local sunset/sunrise time determined by the user's geolocation and current date.

### Why Is This Needed?
Fixed time-based theme switching (e.g., switching at 7PM) doesn’t work well for all regions, especially during seasons where sunset time changes dynamically.
Users who travel or live in regions with large seasonal daylight variation want a dynamic, natural theme change aligned with the environment.
This improves visual comfort, reduces eye strain, and creates a smarter integration with real-world conditions

### Suggested Solutions
I see that the data obtained in the location information already has time, and this time can be used to set.

### Additional Context
Apps like `macOS` Auto Dark Mode and `Darkman` use similar logic (Astronomical sunrise/sunset).

luleyleo:
This adds automatic switching between light and dark mode using either location based or manual sunrise and sunset times.

Closes #493

### DIFF
--- a/Services/DarkModeService.qml
+++ b/Services/DarkModeService.qml
@@ -60,7 +60,7 @@ Singleton {
   }
 
   function init() {
-    Logger.log("DarkModeService", "Service started")
+    Logger.i("DarkModeService", "Service started")
 
     if (Settings.data.colorSchemes.schedulingMode == "manual") {
       const changes = collectManualChanges()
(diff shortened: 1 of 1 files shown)


luleyleo:
This should be ready now

Ly-sec:
Thank you very much for this PR. Just tested it and it seems to work just fine, well done :heart: !

PR #485, opened 2025-10-15
fix(panel): Refine scaling in BatteryPanel

lonerOrz:
图片


lonerOrz:
I thought there was a conflict, but it turned out to be repaired? 😂

PR #490, opened 2025-10-15
Add a check if blacklisted before adding to specific or generic players

damian-ds7:
Currently the blacklist seems to be ignored, this PR re-adds the check if a player is blacklisted before processing it further


Ly-sec:
Thank you very much! Must've been lost between PRs :).

PR #496, opened 2025-10-16
feat(bar): Introduce core density scaling mechanism

lonerOrz:
before
图片

after
图片

### DIFF
--- a/Modules/Bar/Widgets/MediaMini.qml
+++ b/Modules/Bar/Widgets/MediaMini.qml
@@ -86,6 +86,7 @@ Item {
     visible: false
     text: titleText.text
     font: titleText.font
+    applyUiScale: false
     pointSize: Style.fontSizeS * scaling
   }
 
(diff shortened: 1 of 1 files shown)


Modules/Bar/Widgets/MediaMini.qml
```diff
@@ -86,7 +86,7 @@ Item {
     visible: false
     text: titleText.text
     font: titleText.font
-    applyUiScale: false
```
ItsLemmy, line 89: I believe this should not have been deleted

Modules/Bar/Widgets/MediaMini.qml
```diff
@@ -299,7 +299,7 @@ Item {
               NText {
                 text: hasActivePlayer ? getTitle() : placeholderText
                 font: titleText.font
-                applyUiScale: false
```
ItsLemmy, line 302: This one too should stay... we dont want to apply uiScale to the bar (as much as possible, so yeah excluding margins...)

Modules/Bar/Extras/BarWidgetLoader.qml
```diff
@@ -12,6 +12,19 @@ Item {
   property string section: widgetProps && widgetProps.section || ""
   property int sectionIndex: widgetProps && widgetProps.sectionWidgetIndex || 0
 
+  property string barDensity: "default"
+  readonly property real scaling: (function() {
```
ItsLemmy: (function()    is unecessary.

lonerOrz: If don't add `function()`, the code will give an error. Maybe I missed something, so I used ternary operations to write instead

ItsLemmy:
Thanks

PR #499, opened 2025-10-16
Tab navigation in session menu and launcher

damian-ds7:
I've added Tab/Shift+Tab keyboard navigation to cycle through session menu / launcher items. I personally find this more comfortable than arrow keys or Ctrl+j/k and thought it would be a nice addition for others who prefer tab-based navigation.

### DIFF
--- a/Modules/SessionMenu/SessionMenu.qml
+++ b/Modules/SessionMenu/SessionMenu.qml
@@ -119,18 +119,6 @@ NPanel {
   }
 
   // Navigation functions
-  function selectNext() {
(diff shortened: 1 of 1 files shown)


ItsLemmy:
Cool, but I'm unsure why would Up/Down be clamped and Tab/Shift+Tab be wrapped.  Wouldn't it make more senses to have them all wrap ?

damian-ds7:
I can also change it for the other key binds no problem, I just wasn't sure if you'd be willing to change behavior for all of them.

ItsLemmy:
Thanks!

Issue #502, opened 2025-10-16
[Feature] Stop notification timeout when being hovered


Amadoabad:
### Feature Description
Well, It's as simple as that, when hovering a notification stop the timeout then continue after that. 

### Why Is This Needed?
It helps when the notification has a lot of text 

### Suggested Solutions
Just stop the countdown when being hovered

PR #504, opened 2025-10-16
Notifications freeze when hovered

Amadoabad:
Well I've tried this locally and it works. It freezes notifications when being hovered by the cursor.
It adds this feature #502


Ly-sec:
Thank you very much, I was thinking about a good solution for this. That seems to be the best one :). Merged

Amadoabad:
Thanks to Gemini-cli XD

PR #511, opened 2025-10-17
ActiveWindow: implement dynamic width with max width setting

sakarie9:
Introduce a maximum width setting for the ActiveWindow widget, allowing it to adapt its size dynamically based on content while respecting user-defined constraints.


ItsLemmy:
Thanks!

PR #512, opened 2025-10-17
feat(ActiveWindow): add adaptive width

Sighthesia:
### Changes overview
The widget can now dynamically adjust its width to match the length of the current window title, accompanied by animations for hiding, becoming transparent, and reappearing. Note that vertical bars only support visibility-based animations (no width transitions).
GIF:

### What I did
- Added five new settings options:
  1. `Enable adaptive width`: Allows the widget to expand its width based on the current focused window’s title length, up to the limit defined by Maximum adaptive width.
  2. `Maximum adaptive width`: Sets an upper bound for the widget’s width. Defaults to 290 (twice the default `widgetWidth`).
  3. `Use title width as minimum width` (visible only when Enable adaptive width is on): Ensures the widget’s width is at least as wide as the current window title, preventing unnecessary truncation.
  4. `Minimum adaptive width` (visible only when Enable adaptive width is on): When enabled, the original `Widget Width` setting becomes the minimum width of the widget.
  5. `Force title scrolling`: Enables scrolling even when the widget is wide enough to display the full title. This is intended to work with the `Scroll On Hover feature` because scrolling looks cool, yeah...
- Implemented width transition animations for the widget. These apply not only during width changes but also when the widget hides, fades out, or appears, creating a fluid layout shift that pushes adjacent widgets.
- Unified scrolling logic: The title scrolling now resets to the starting position and pauses for 1 second whenever the focused window changes. This improves readability by ensuring users see the beginning of each new title.

Under the hood, I used `implicitWidth` to obtain the current focused window’s title width and reused the width calculation logic from the existing scrolling module to determine the minimum width needed to display the title without truncation.

### Potential issues
1. The width calculation may have minor inaccuracies, though they are likely imperceptible in practice.
2. I was unable to build qmlfmt on my machine, so I manually formatted the code to match the project’s existing style as closely as possible.
3. Commits are messy because i took a lot of experiment in it...

If any changes are needed, please let me know, truly appreciate your time and hard work!

### DIFF
--- a/Assets/Translations/de.json
+++ b/Assets/Translations/de.json
@@ -926,15 +926,15 @@
           "label": "Minimale adaptive Breite",
           "description": "Minimale Widget-Breite, die verwendet wird, wenn adaptive Breite aktiviert ist."
         },
-        "autoWidthEnabled": {
+        "adaptiveWidthEnabled": {
           "label": "Adaptive Breite aktivieren",
           "description": "Die Widget-Breite basierend auf der Länge des Fenstertitels automatisch anpassen."
         },
-        "autoWidthMax": {
+        "adaptiveWidthMax": {
           "label": "Maximale adaptive Breite",
           "description": "Maximale Breite, auf die das Widget erweitert werden kann, wenn adaptive Breite aktiviert ist."
         },
-        "autoWidthMinByContent": {
+        "adaptiveWidthMinByContent": {
           "label": "Textbreite als Minimum verwenden",
           "description": "Wenn aktiviert, verwendet das Widget die Textbreite des Fenstertitels als Mindestbreite anstelle eines festen Minimums."
         },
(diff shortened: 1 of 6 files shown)


ItsLemmy:
Hey, we merged a concurrent PR that addressed both ActiveWindow and MediaMini. (they are very similar)
Not sure if that fix your use case, or if you want to rework this one to adapt.
If you decice to rework it, I'd like to avoid adding too many new settings to avoid cluttering the UI, 4 new toggles seems a bit overkill to me.

Sighthesia:
Thank you for reviewing! I just noticed that #510 already implements the same feature in a cleaner way. I fully support merging that PR instead.
I’ll close this one. Sorry for the duplicate effort, and thanks again for your time. I’ve learned a lot from both the implementation and the review process.

ItsLemmy:
Thanks a lot for your contribution none the less, your PR was good, it's just unfortunate that the two of you addressed the same issue simultaneously.  Don't hesitate to contribute again!

PR #516, opened 2025-10-17
Make panels open in overlay layer

### LINKED ISSUE #477: [Feature] Make App launcher thingy etc. appear on fullscreen windows
Fuzzel puts itself on the "overlay" layer, making it visible even while in fullscreen applications.

Noctalia doesn't do this. It should do this because it's confusing when the shortcut for launcher/clipboard/something else is pressed and nothing shows up while in a fullscreen window.

damian-ds7:
Currently panels render on the Top layer, but if a panel is opened while a program is fullscreen, it's likely intentional (via keybind or command). This change renders panels over fullscreen windows. If there are issues with this approach I haven't considered, I can add a wrapper over NPanel to selectively open some panels in the overlay layer.

Closes #477


Ly-sec:
I think having a wrapper for it would be a cleaner approach given that not everything has to be an overlay.

damian-ds7:
Cool, I'll make a wrapper then, which panels do you think should use it? I think it would make most sense with Session menu, launcher, control center and notifications history.

damian-ds7:
I created a new component for PanelWindow and reuse it in both Panels, this should make it easy to modify any behavior for both Panels separately if there's need for it, let me know if there is something I should change in my implementation

Ly-sec:
Hey there, we talked about this implementation internally and it might be much better if we just have a simple `isOverlay` in the `NPanel.qml` instead of creating a separate file, this way the users could just globally toggle all NPanels as overlay (inside the User Interface tab) . What do you think?

damian-ds7:
Yeah that makes more sense tbh, I can implement that

damian-ds7:
Should it be enabled by default or should current behavior be the default setting?

damian-ds7:
For now I made panels open in overlay layer by default, let me know if you want me to change that

ItsLemmy:
Yep, it's fine to go Overlay by default, thanks!

idlip:
Imo bar should not be over full screen?

Except for bar everything makes sense to be overlay at top.
But bar is something which is seen at all time, but would be sensible to not look at with full screen.

damian-ds7:
@idlip IIRC there was a rework to NPanel which changed some things from the time I sent this PR so it's probably a bug

ItsLemmy:
Not a defect, from now on panels and bar live on the same wayland layer so we can have a proper unified look (bar+panels attachement, shadows, etc...)     So if you choose to have that layer set to overlay, then the bar will be there too.    Notifications, toasts and OSD are not affected as they are on a their own separate layer.

I have yet to decide if this option is worth keeping, as it may lead to confusion for some  users. What do you think?

damian-ds7:
If they will be on one layer now it's probably not worth keeping, unless there's a way to detect Fullscreen and hide bar when it's active. I don't think having bar In fullscreen mode will be beneficial

PR #520, opened 2025-10-18
feat: Implement language selection feature

lonerOrz:
This PR introduces an in-app language selection feature.

图片

### DIFF
--- a/Assets/Translations/zh-CN.json
+++ b/Assets/Translations/zh-CN.json
@@ -76,7 +76,8 @@
         },
         "select": {
           "label": "应用程序语言",
-          "description": "选择应用程序界面中使用的语言。"
+          "description": "选择应用程序界面中使用的语言。",
+          "auto-detect": "自动检测"
         }
       }
     },


ItsLemmy:
I have to say I'm not super excited about this one.   We discussed it many times internally and we came to the conclusion that it's better to follow the system's locale. 

That being said it's cool to be able to switch language for testing and also to show to new users that we support localization, so I think we should probably implement it with a few adjustments.

The way you implemented it totally bypass the locale detection as you set the default settings to "en". 

My suggestions: 

-  property string language: ""           empty strings by default, mean no override, use the old way (system locale detection)

- add an entry, as the first entry of the NComboBox  "Automatic [detected locale]"  which equals ""  so that people can switch back to the automatic behaviour.  Obviously 'detected locale' should be replaced with the detected locale: en, de, etc...

- remove i18n.debugForceLanguage  and all associated code, as it is no longer necessary.


Thank you!

lonerOrz:
> I have to say I'm not super excited about this one. We discussed it many times internally and we came to the conclusion that it's better to follow the system's locale.

I hope my system language is English, but this shell program is Chinese, so I want to have an explicit setting option to select the language.

I also agree with your suggestion

lonerOrz:
Done！

ItsLemmy:
Thanks, good job!

PR #521, opened 2025-10-18
Add grouped Windows Widget

herobrauni:
No idea if there is anybody else who wants this. Used it for a few days and works quite nice.

Creates a Widget that shows all windows grouped by workspace. Think of a merge of the workspace and taskbar widgets that already exist.

image

You can click on icons to switch to the app/window.

Full disclaimer, I don't know much about qml, so there is a lot of ML help involved.

Feel free to change or close, however you see fit.

### DIFF
--- a/Assets/Translations/de.json
+++ b/Assets/Translations/de.json
@@ -533,6 +533,10 @@
         }
       },
       "templates": {
+        "section": {
+          "label": "Vorlagen",
+          "description": "Farben auf externe Anwendungen anwenden."
+        },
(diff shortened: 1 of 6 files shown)


maddie-m:
I would absolutely love this to be merged, I was thinking about implementing something similar for myself.

ItsLemmy:
Thank you for the contribution.   I'm merging this as it's a new widget and pretty safe, will see how it behave in the upcoming days.

PR #525, opened 2025-10-19
MediaMini: handle empty media content with active player for rounded background

Sighthesia:
## Overview
image

## Changes

Add additional width judgement for active media player with no content, intended for some stubborn media events.

Not sure if it is better to hide the widget in this context, let me know if there are changes needed, thank you very much!


Ly-sec:
I think that's the best solution, hiding it might cause some weird shifting in the bar :). Going to merge it, thank you very much!

PR #526, opened 2025-10-19
Display SSID in WiFi bar pill

luleyleo:
Looking like this:
image

I plan to do the same with the Bluetooth widget next.


ItsLemmy:
Thanks

PR #531, opened 2025-10-20
Smarter n text input

MrDowntempo:
Adds a clear button when text is entered. Also adds an optional inputIconName for NTextIput which is only currently used to add a magnifying glass when it is a seach field. I added that magnifying glass to the NSearchableComboBox, and reworked NFilePicker to just appear to use one text entry field.


MrDowntempo:
While I didn't do the Home and Up tooltips. I did find there was already placeholder text for Search.... in our translations, so I just pointed to I18n.tr("placeholders.search") So that's one down.

PR #533, opened 2025-10-20
Add support for streaming command outputs in CustomButton

shouya:
This PR adds a toggle that enables CustomButton's text command to run continuously and use the streamed output lines as the display text.

### DIFF
--- a/Assets/Translations/en.json
+++ b/Assets/Translations/en.json
@@ -1073,9 +1073,14 @@
           "description": "Command to execute when the button is middle-clicked."
         },
         "dynamic-text": "Dynamic text",
+        "text-stream": {
+          "label": "Stream",
+          "description": "Streamed lines from the command will be displayed as text on the button."
+        },
         "display-command-output": {
           "label": "Display Command Output",
-          "description": "Enter a command to run at a regular interval. The first line of its output will be displayed as text."
+          "description": "Enter a command to run at a regular interval. The first line of its output will be displayed as text.",
+          "stream-description": "Enter a command to run continuously."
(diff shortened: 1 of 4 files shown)


Ly-sec:
Hey there :)! We were just talking about your PR and @ItsLemmy mentioned there is https://quickshell.org/docs/v0.2.0/types/Quickshell.Io/StdioCollector/ that one, wouldn't it make more sense to use that instead of polling?

shouya:
Do you mean using the property change signal of the `text` field? If the text command doesn't use line-buffering, or if the command flushes multiple lines at once, I'm not sure what the `text` property will be. The SplitParser is specified that it spits out well-behaved lines, that's why I chose it.

BTW, I never wrote any QML before, so it would be great if you can point out any coding improvements. Also, I'm struggling to make the widget run `runTextCommand()` on startup, so it would be nice if you can suggest a way do it.

Ly-sec:
> Do you mean using the property change signal of the `text` field? If the text command doesn't use line-buffering, or if the command flushes multiple lines at once, I'm not sure what the `text` property will be. The SplitParser is specified that it spits out well-behaved lines, that's why I chose it.
> 
> BTW, I never wrote any QML before, so it would be great if you can point out any coding improvements. Also, I'm struggling to make the widget run `runTextCommand()` on startup, so it would be nice if you can suggest a way do it.

We talked in discord a few days back and I mentioned that you might want to talk to @ItsLemmy about it, perhaps you guys can figure something out :).

ItsLemmy:
Thanks

PR #542, opened 2025-10-21
MediaMini: improvements on visual effect

Sighthesia:
## Overview
Fix some glitches on title scorlling and make it's visual effect consistent with ActiveWindow

## Changes

1. Fix incontinous title scrolling when media isn't playing: Before that, there was only one title text so the scorlling wasn't seamless.
2. Fix insufficient width for title in low scaling factor which causes unexpected scrolling.
3. Sychronize title width animation with widget width: Utilizing background rectangle width transition to crop the title.
4. Fix width jump to `maxWidth` when widget appears.
5. Fix insufficient width that triggered scrolling when no active player: Casued by inadequate icons margin.
6. Improve title scorlling behavior: Always reset scrolling postion and pause for readability when media changes.
7. Add fade-in/out animation: Opacity and width transition.

## Potential Modifications

- When `Always Visible` is on and one media player event ends, there will be a lag due to media evnet change, so the fade-out animation will behave like shrinking to minimum width(media icon width) then bounce to the width of "No active player".
  Although, it looks funny :)

If any changes are needed, please let me know. Thanks a lot for reviewing.

### DIFF
--- a/Assets/Translations/de.json
+++ b/Assets/Translations/de.json
@@ -32,16 +32,6 @@
           "reset": "Eckenradius des Bildschirms zurücksetzen"
         }
       },
-      "lockscreen": {
-        "section": {
-          "label": "Sperrbildschirm",
-          "description": "Hier kannst du deinen Sperrbildschirm konfigurieren."
-        },
-        "lock-on-suspend": {
-          "label": "Beim Standby sperren",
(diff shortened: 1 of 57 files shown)


ItsLemmy:
Your previous PR conflicts with this new one :P

ItsLemmy:
Thanks!

PR #543, opened 2025-10-21
feat: Add toggle to enable or disable calendar event display

lonerOrz:
This update introduces a new setting that allows users to control whether calendar events are displayed and loaded in the bar clock.

Note
I personally hope to see broader support for different calendar or event applications in the future, as I don’t use the current one myself.


ItsLemmy:
Nice, I wanted to do that but it was too late yesterday, thanks!

PR #549, opened 2025-10-21
 IPC for PowerProfile cycle

Sridou:
Added IPC call for PowerProfile cycle.

### DIFF
--- a/Services/IPCService.qml
+++ b/Services/IPCService.qml
@@ -222,8 +222,7 @@ Item {
     }
   }
   IpcHandler {
-    target: "PowerProfile"
-
+    target: "powerProfile"
     function cycle() {
        PowerProfileService.cycleProfile()
     } 
--- a/Services/PowerProfileService.qml
+++ b/Services/PowerProfileService.qml
@@ -49,6 +49,10 @@ Singleton {
     }
   }
 
+  function init() {
+   Logger.d("PowerProfileService", "Service started")
+  }
+
   function setProfile(p) {
     if (!available)
       return
--- a/shell.qml
+++ b/shell.qml
@@ -91,7 +91,8 @@ ShellRoot {
         HooksService.init()
         BluetoothService.init()
         BatteryService.init()
-        IdleInhibitorService.init()
+	IdleInhibitorService.init()
+	PowerProfileService.init()	
       }
 
       Background {}


Sridou:
Noticed a bug. The first call on quickshell reload/load does not change but the subsequent calls change, will look into it when I get the time.

Sridou:
I could not resolve it, the first call returns with `hasPerformanceProfile` as `false` - here's the debug:
```
DEBUG qml: === PowerProfileService initialized ===
 DEBUG qml: powerProfiles: qs::service::upower::PowerProfilesQml(0x7f1c467db9f0)
 DEBUG qml: powerProfiles null?: false
 DEBUG qml: powerProfiles undefined?: false
 DEBUG qml: hasPerformanceProfile: false
 DEBUG qml: current profile: 1
 DEBUG qml: available: false
 DEBUG qml: === cycleProfile() called ===
 DEBUG qml: available: false
 DEBUG qml: Not available, returning
 DEBUG qml: === cycleProfile() called ===
 DEBUG qml: available: true
 DEBUG qml: current profile from powerProfiles: 2
 DEBUG qml: local profile property: 2
 DEBUG qml: powerProfiles object: qs::service::upower::PowerProfilesQml(0x7f1c467db9f0)
 DEBUG qml: hasPerformanceProfile: true
 DEBUG qml: Switching Performance -> PowerSaver
 DEBUG qml: setProfile called with: 0
 DEBUG qml: Setting powerProfiles.profile to: 0
 DEBUG qml: [20251022-145311]    ToastScreen Queuing notice : Power profile changed "Power saver"
 DEBUG qml: [20251022-145311]    ToastScreen Queuing notice : Power profile changed "Power saver"
 DEBUG qml: Successfully set profile
 DEBUG qml: After setProfile, powerProfiles.profile: 0
 DEBUG qml: === cycleProfile() done ===
```
I checked how the bar/widget implements and it has the same function  `onClicked: PowerProfileService.cycleProfile()`

damian-ds7:
You should add an `init` function to `PowerProfileService`:

```qml
function init() {
  Logger.d("PowerProfile", "Service started")
}
```

Then call it in `shell.qml` along with the other services, that should fix it.

Also, it would be nice if you renamed the IPC call target from `PowerProfile` to `powerProfile` to match the others.

PR #550, opened 2025-10-22
Add more consistency for expression and interpunction & fix some errors

notiant:
It probably makes sense to don't mix between addressing the user as "Du" and "Sie".


Ly-sec:
Thank you once again :). It does indeed make sense, we did generate the tranlsations and I was slowly going through the german one but didn't get far yet. If you find any other inconsistencies just let us know!

PR #551, opened 2025-10-22
Fix wrong day names in the weather forecast

notiant:
The lock screen shows 3 times the current day name.
Copied the fix from CalendarPanel.qml


Ly-sec:
Oh! Thank you very much!

PR #557, opened 2025-10-23
BrightnessService: add minimum brightness to prevent backlight from t…

Sighthesia:
## Changes

Add a minimum brightness value of 1% to prevent the display backlight from being completely turned off. And when increasing brightness from 1%, it will jump to minimum brightness step for consistency.

## Potential required changes

I'm not sure if all machines will completely turn off the backlight at 0% brightness. This is a workaround borrowed from other shell(illogical-impulse). Maybe we could add a setting option for the minimum brightness?

Let me know if changes are needed, thanks for reviewing!

### DIFF
--- a/Assets/Translations/de.json
+++ b/Assets/Translations/de.json
@@ -162,6 +162,10 @@
         "brightness-step": {
           "label": "Helligkeits-Schrittgröße",
           "description": "Schrittgröße für Helligkeitsänderungen anpassen (Mausrad und Tastenkombinationen)."
+        },
+        "enforce-minimum": {
+          "label": "Mindesthelligkeit erzwingen (1%)",
+          "description": "Löst das Problem, dass die Hintergrundbeleuchtung auf einigen Displays bei 0% Helligkeit vollständig ausgeschaltet wird."
         }
       },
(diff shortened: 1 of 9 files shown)


Ly-sec:
Hello there, I really like this idea. Makes sense given it can be hard to see with 0% brightness on some displays.
If you want, you can add a toggle for it in the DisplayTab if you want I can do it too, should be quick and easy.

Sighthesia:
Ok, How about adding an slider for the minimum brightness value, and then representing the lowest brightness value with a vertical bar on the OSD brightness slider? Hmm... This might make more sense when it comes to setting a maximum volume limit. ... Which is better? I'm a bit worried that this might make the settings a bit bloated.

Ly-sec:
Hmm I do feel like we should have a toggle for it to begin with since there a people (including me) using 0% brightness. For the bar on the OSD, that might actually be a really cool visual display of it :)

Sighthesia:
Ok, i will keep things simple. I've added a toggle for this feature. After struggling, i found that brightness range might be complicated to use ;-;

ItsLemmy:
Thanks

PR #561, opened 2025-10-24
chore: add battery manager uninstall script to gitignore

AlphaNecron:


Ly-sec:
Hello, is there any specific reason why the uninstall script should be in `.gitignore`?

AlphaNecron:
When you install battery-manager, it creates an uninstall script, which prevents `git pull`. Ik you can commit it but there's literally no reason to do so as that script is personalized.

damian-ds7:
That's a good point I totally forgot about it, I even added it locally and forgot about it

ItsLemmy:
@damian-ds7 Would be cool if you would join our discord server, we sometimes have questions / request about the battery manager that is discussed over there.  Thanks!

PR #562, opened 2025-10-24
Fix app icon theming in light mode

notiant:
Themed app icons are basically black in light mode, they shouldn't be themed with a dark color.


Ly-sec:
That's a good call! You forgot about the tray and Taskbar though, mind adding that too :)?

notiant:
I'm not using them so I didn't know, but I've updated them too

notiant:
It also seems that TaskbarGrouped doesn't have any widget settings, so the theming can't be enabled

Ly-sec:
> It also seems that TaskbarGrouped doesn't have any widget settings, so the theming can't be enabled

Yeah we still need to add settings to it :)
I'll test it real quick and then merge it

PR #565, opened 2025-10-24
AudioCard: make both sliders have the same width

notiant:
Enforce visual consistency by making both output and input sliders have the same width regardless of the devices names


ItsLemmy:
Thanks!

PR #566, opened 2025-10-24
WeatherCard: Make current weather align with forecast

notiant:
Aligns the current weather with the first day of the forecast to add some spacing to the left


ItsLemmy:
Thanks again, looking much better

PR #567, opened 2025-10-24
Fix/small alignment

MrDowntempo:
# Pull Request

## Motivation
Mainly, this fixes an ugly alignment issue in the settings-audio tab's Excluded Player, by swapping out a NTextInput and NIconButton for an NTextInputButton. I also did the same for inside the Bar's Tray settings window, but there was no initial alignment problem there, just doing it for consistency. I also let NTextInputButton have access to NTextInput's inputIcon feature I added in the past. Although I didn't use it anywhere.

## Type of Change
- [X] Bug fix
- [X] New feature
- [ ] Breaking change
- [ ] Refactoring

## Testing
I tested with different scaling values. 
- [ ] Tested on compositor: Niri only

## Screenshots / Videos
image

## Checklist
- [X] Self-reviewed my code

### DIFF
--- a/Modules/Settings/Bar/WidgetSettings/TraySettings.qml
+++ b/Modules/Settings/Bar/WidgetSettings/TraySettings.qml
@@ -50,17 +50,12 @@ ColumnLayout {
       Layout.fillWidth: true
       spacing: Style.marginS
 
-      NTextInput {
+      NTextInputButton {
         id: newRuleInput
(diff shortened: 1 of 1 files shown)


MrDowntempo:
I kinda think that the NTextInput and NIconButton inside NTextInputButton should be the same height, but after tinkering a bit with that file, it doesn't look as easy to do as I had expected.

PR #589, opened 2025-10-26
Separate screen corners and rectangle hug corners by introducing bar style options

### LINKED ISSUE #487: [Feature] Add Screen Edges to the Screen Edges Even When Having Non Floating Bar
### Feature Description
Can we add an option to add screen edges to the screen while still having the bar edges? so we can have a nicely looking bar with the display edges being curved?
Notice this is only related to bar when it's not floating.

### It should look something like this

Image

but instead of the gray color on the left-hand side it will be solid black and just an edge at the top 

### Why Is This Needed?
For the looks. It will look awesome

### Suggested Solutions
We can just duplicate the edges on the the bar side and make them stick to the edge.

Sighthesia:
# Pull Request
1. Replace floating bar switch with bar style option box, including three bar style:`Floating`, `Rectangle`, `Rectangle (hug to window)`.
2. Pin original corners to screen edges.
3. Copy edge corners as hug corners for `Rectangle (hug to window)`.
4. Set `Rectangle (hug to window)` as default bar style.
5. Update settings and setup wizard and sort the order of option.

## Motivation
Intuitive Settings and Consistent Appearance: Separate screen corners and rectangle hug corners, so they take their own reponsiblity of rounding screen corners and decorating non-floating bar.

## Type of Change
- [ ] Bug fix
- [ ] New feature
- [x] Breaking change: data type changed.
- [ ] Refactoring

## Related Issue
- Closes #487 

## Testing
- [x] Tested on compositor: Niri
- [ ] Tested with multiple monitors: has been taken tnto consideration but i dont have multiple monitors.
- [x] Tesetd with screen corners on/off
- [x] Tested on bar color/opacity changed

## Screenshots / Videos
image

## Checklist
- [x] Self-reviewed my code

## Additional Notes
The previous floating boolean setting has been replaced with a **string-based style option** to support multiple bar styles. A **backward-compatible boolean** fallback is retained for existing configurations, also there's deprecated JSON keys upgrading in `Settings.qml`. So there might be a futher migration of this value. Let me know if changes are needed, thanks!

### DIFF
--- a/Modules/SetupWizard/SetupCustomizeStep.qml
+++ b/Modules/SetupWizard/SetupCustomizeStep.qml
@@ -197,47 +197,89 @@ ColumnLayout {
         Layout.bottomMargin: Style.marginS
       }
 
-      // Dim Desktop section
+      // Bar style selector
       RowLayout {
         Layout.fillWidth: true
         spacing: Style.marginM
-
         Rectangle {
           width: 32
           height: 32
           radius: Style.radiusM
           color: Color.mSurface
           NIcon {
-            icon: "moon"
+            icon: "layout-2"
             pointSize: Style.fontSizeL
             color: Color.mPrimary
             anchors.centerIn: parent
           }
         }
-
         ColumnLayout {
           Layout.fillWidth: true
           spacing: 2
           NText {
-            text: I18n.tr("settings.user-interface.dim-desktop.label")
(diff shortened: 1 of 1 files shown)


ItsLemmy:
Thanks for the contribution but I did a large refactor if the windowing/screen system, and we now have separate corners for bar and screen + a whole lot of new features.

PR #596, opened 2025-10-26
SessionMenu: Increase font size & fix missing translation

notiant:
The font size in the session menu seems a bit too small for me, also the pending actions aren't translated.


ItsLemmy:
Thanks, indeed not sure why it was not "M"edium

PR #604, opened 2025-10-27
feat: add weather widget to bar with forecast popup

msdevpt:
# Pull Request

## Motivation
Just a simple widget to show the weather on the bar with the forecast popup.

## Type of Change
- [ ] Bug fix
- [x ] New feature
- [ ] Breaking change
- [ ] Refactoring

## Testing
- [] Tested with multiple monitors (if applicable)

## Screenshots / Videos
imagem

## Checklist
- [x] Self-reviewed my code

### DIFF
--- a/Assets/ColorScheme/Rosepine/terminal/alacritty/Rosepine-dark
+++ b/Assets/ColorScheme/Rosepine/terminal/alacritty/Rosepine-dark
@@ -3,7 +3,7 @@
 
 [colors.primary]
 foreground = "#e0def4"
-background = "#191724"
+background = "#1f1d2e"
 dim_foreground = "#908caa"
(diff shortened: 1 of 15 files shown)


ItsLemmy:
Thanks for the contribution. I would request a few adjusments.
- WeatherCard should be named WeatherPanel.
- WeatherPanel should reside in shell.qml instead of inside the widget, like all other panels.  This will make it easier to open that panel from IPC calls, etc...   It is safe to do so, as all NPanel are wrapped in a loader so they use almost no memory until you open them.
- The WeatherPanel should respect uiScaling, ie:
```
  preferredWidth: 500 * Style.uiScaleRatio
  preferredHeight: 280 * Style.uiScaleRatio
```

Thank you!

PR #617, opened 2025-10-30
Feature/split clocks

MrDowntempo:
# Pull Request

## Motivation
Splits analog and digital clock logic into separate files for a possible minuscule reduction in memory. But primarily means this logic can be shared with the Lock Screen, instead of repeated code there.

## Type of Change
- [ ] Bug fix
- [x] New feature (barely)
- [ ] Breaking change
- [x] Refactoring

## Testing
Tested both analog and digital clocks in the calendar and lock screen. Mostly looks the same as before, lockScreen has slightly different colors now since the code was a tad different, and so does the digital clock on the calendar since everything shares the same code. 

## Checklist
- [x] Self-reviewed my code

## Additional Notes
You know I ain't good at this. Double check plz

### DIFF
--- a/Modules/Bar/Calendar/CalendarPanel.qml
+++ b/Modules/Bar/Calendar/CalendarPanel.qml
@@ -219,8 +219,8 @@ NPanel {
         anchors.right: parent.right
         anchors.rightMargin: Style.marginM
         anchors.verticalCenter: parent.verticalCenter
+        progressColor: Color.mOnPrimary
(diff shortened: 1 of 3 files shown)


MrDowntempo:
Don't merge this yet. I'm gonna make some tweaks to the color selection so it uses the previous colors in some instances where they were the better choice.

MrDowntempo:
_Should_ be good to merge now.

ItsLemmy:
Thank you sir!

PR #628, opened 2025-10-31
Nix: add systemd service to home-module

### LINKED ISSUE #627: [Feature] Add support for a user-level systemd service on Nix systems
### Feature Description
`programs.noctalia-shell` can enable the package and settings but cannot run as a user-level systemd service. The NixOS module supports this but requires extra imports and only works on NixOS.

### Why Is This Needed?
To let non-NixOS users who use Home-manager run Noctalia shell as a user service.

### Suggested Solutions
Add an option like `programs.noctalia-shell.systemd.enable = true` that sets up a systemd --user unit and its defined target automatically.

Alternatively, rename the module from `programs.noctalia-shell` to `services.noctalia-shell` and have it automatically enable the user service without needing a separate option.

**EDIT**: I realised that the current service is already running on user-level. But i am not entirely sure are you able to import the module on non-NixOS systems.

bokicoder:
# Pull Request

## Motivation

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring

## Related Issue
- Closes #627 

## Testing
- [ ] Tested on compositor: (e.g., Hyprland, Sway, Niri)

## Screenshots / Videos

## Checklist
- [x] Self-reviewed my code

## Additional Notes

### DIFF
--- a/nix/home-module.nix
+++ b/nix/home-module.nix
@@ -25,13 +25,13 @@ in {
 
     settings = lib.mkOption {
       type = with lib.types;
-        nullOr (oneOf [
+        oneOf [
           attrs
           str
           path
-        ]);
+        ];
       default = {};
-      apply = x: lib.recursiveUpdate defaultSettings (extractAttrs x);
+      apply = x: if x == {} then x else lib.recursiveUpdate defaultSettings (extractAttrs x);
       example = lib.literalExpression ''
         {
           bar = {
@@ -111,8 +111,9 @@ in {
           Documentation = "https://docs.noctalia.dev/docs";
           PartOf = [ config.wayland.systemd.target ];
           After = [ config.wayland.systemd.target ];
-          X-Restart-Triggers = [ "${config.xdg.configFile."noctalia/settings.json".source}" ]
-            ++ lib.optional (cfg.colors != { }) "${config.xdg.configFile."noctalia/colors.json".source}";
+          X-Restart-Triggers =
+            lib.optional (cfg.settings != {}) "${config.xdg.configFile."noctalia/settings.json".source}"
+            ++ lib.optional (cfg.colors != {}) "${config.xdg.configFile."noctalia/colors.json".source}";
         };
 
         Service = {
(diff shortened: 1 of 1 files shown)


jervw:
Thank you! Works like a charm. Maybe also add a configurable target like in the NixOS module?

bokicoder:
> Maybe also add a configurable target like in the NixOS module?

You can configure `wayland.systemd.target` option in Home Manager. From my observation, all wayland modules in Home Manager use this exact option to set the target for systemd services.

nix/home-module.nix
```diff
@@ -103,18 +105,46 @@ in {
...
+        Install.WantedBy = [ config.wayland.systemd.target ];
+      };
+
       home.packages = lib.optional useApp2Unit cfg.app2unit.package
         ++ lib.optional (cfg.package != null) cfg.package;
 
       xdg.configFile = {
         "noctalia/settings.json" = {
-          onChange = restart;
+          onChange = lib.mkIf (!cfg.systemd.enable) restart;
```
ixxie: If a user uses the NixOS module systemd service, this would disable restarting it, right? We need to still support the other one.

ixxie (changes requested):
Great contribution @bokicoder - have one little (but important) comment.

In addition, I'd like to see this properly documented before merging; could you open a second PR to [the docs repo](https://github.com/noctalia-dev/noctalia-docs) to document the service? Link to this PR and the issue.

nix/home-module.nix
bokicoder: The logic here is to disable `onChange` only when using the Home Manager module's systemd services, because they have better `X-Restart-Triggers`. Therefore, this change shouldn't break anything, right?

ixxie: Aha, I got things mixed up... is there are reason we couldn't use the same approach for the NixOS systemd service? Then we could just ditch this altogether?

bokicoder: IDK, but honestly `onChange` has been pretty bad in my experience. This PR is a better way.

ixxie: If you are right I'd really like to explore reusing the approach for the NixOS service, but its fair to scope this out of this PR.

ixxie:
@bokicoder I've now tested this thoroughly and I'm happy with it.

There are only one thing I'm left wondering about: I was able to enable _both_ the NixOS and the HM systemd services; shouldn't we throw some error for that particular combination of settings? It seems dangerous.

bokicoder:
@ixxie The NixOS Module and Home Manager are two independent configuration systems. When writing a module, it's impossible to reference the other's option values, so I don't think this can be done. However, I anticipated this situation: I made the systemd services in both modules share the same name, so in practice, only the Home Manager systemd service should run.

ixxie:
> @ixxie The NixOS Module and Home Manager are two independent configuration systems. When writing a module, it's impossible to reference the other's option values, so I don't think this can be done. However, I anticipated this situation: I made the systemd services in both modules share the same name, so in practice, only the Home Manager systemd service should run.

You learn something new every day. I would have thought there was a way to interface between the systems.

Anyway your solution is clever! I'll approve and merge this PR, but I would request you document this behavior in the other PR.

PR #631, opened 2025-10-31
Add airplane mode detection, improve NetworkService & add icons to notice toasts

notiant:
# Pull Request

## Motivation
I have a function key for toggling airplane mode on my laptop's keyboard. Currently, when i press it noctalia shows me a notification for disabling bluetooth, but wifi stays enabled in the shell interface (although the wifi radio is off). This PR changes the behavior of BluetoothService.qml to now check on adapter state changes (not just enabled/disabled) to see if the adapter has been blocked. If wifi has also been blocked, it disables wifi in noctalia and sends an airplane mode toast (there is no adapter.state for wifi in quickshell, i think listening to this would require dbus integration). I did not add any ability to trigger airplane mode from noctalia, although that could be added.

I have also found some problems in NetworkService.qml. Enabling/disabling wifi doesn't automatically change the bar widget icon, so I added that a scan should be performed after enabling wifi and the networks variable should be cleared after disabling wifi. I also added a connectivity timer that regularly checks the network connectivity status to keep the UI updated and show a warning toast if the network has lost internet access. Further, I found a bug that doesn't update the wifi panel UI after clicking on connect on non-English systems since an English output is expected.

Lastly, I thought that it would be better if toast notifications like "Bluetooth enabled" would show the bluetooth icon instead of the generic check mark icon, so I've added the ability to specify an icon for ToastService.showNotice() and added icons to all showNotice() commands in the repo.

## Type of Change
- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring

## Testing
- [x] Tested on compositor: Hyprland

## Checklist
- [x] Self-reviewed my code

### DIFF
--- a/Services/NetworkService.qml
+++ b/Services/NetworkService.qml
@@ -50,16 +50,16 @@ Singleton {
     target: Settings.data.network
     function onWifiEnabledChanged() {
       if (Settings.data.network.wifiEnabled) {
-        if (!BluetoothService.airplaneModeToggled)
+        if (!BluetoothService.airplaneModeToggled) {
           ToastService.showNotice(I18n.tr("wifi.panel.title"), I18n.tr("toast.wifi.enabled"), "wifi")
-        
+        }
         // Perform a scan to update the UI
         delayedScanTimer.interval = 3000
         delayedScanTimer.restart()
       } else {
-        if (!BluetoothService.airplaneModeToggled)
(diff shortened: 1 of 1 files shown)


ItsLemmy:
Thanks for your contribution.
It's a bit hard to fully review as it's pretty large, may be next time try to split it in 2 (ex: toast improvement and network/bt service)

Merging!  thanks again

PR #639, opened 2025-11-01
Initial work for wezterm color theme support

MrDowntempo:
# Pull Request

## Motivation
Adding support for Wezterm as first class terminal.

## Type of Change
- [ ] Bug fix
- [X] New feature
- [ ] Breaking change
- [ ] Refactoring

## Testing
Made sure every theme, light and dark, correctly applied. As well as Matugen. Checked for any nonsense in the logs, but didn't spot anything.

## Checklist
- [x] Self-reviewed my code

## Additional Notes
Work in progress


MrDowntempo:
Still have to work on the shell scripting, the default themes, and testing.

MrDowntempo:
Ok. This is ready for merge (or at least as much as close as I can get it currently). A couple points though. colors-apply.sh assumes you have a wezterm.lua in ~/.config/wezterm/ which is where it's supposed to be, and suggested to be. But wezterm is powerful, and can operate under other circumstances. It also assumes that file is built properly, according to https://wezterm.org/config/files.html But again, it's possible to have a wezterm file that doesn't use a config block, or names it differently for god knows why. I believe it will fail elegantly and point you to that link, but I haven't tested THAT. (I ain't b0rking my config!).

The color themes that are included with wezterm have a lot of extra stuff that's themed. So for those themes (Catpuccin, Nord, etc.) I copied that over. Some like Dracula and Eldritch were dark only, in those cases, the light theme is crafted to match, so also includes that extra stuff. Everything else basically only themes the same values that kitty and the other themes do (22 values).

Some of that extra stuff doesn't appear to change work, but I think that might be a git only feature currently, or I don't understand the precise scenarios they're used. Still, they don't break anything else.

I'm not happy with the Matugen theme, but it's based on the other matugen themes for the other terminals. I think these should all get a once over, because some value are used more than I'd prefer and we don't get as much variety in the themes as we should (tertiary and secondary are used more than once from 0-7 and 8-15, but with minor edits via _dim. I think we can do better, especially since didn't Lemmy create some more internal colors to use? I played with _bright and _dim some but, got a lot of errors, so just went back to mimicking the kitty theme.

But other than all that. I think this is working pretty smoothy and should be good to go. Any additional testing to be sure would be greatly appreciated though. Because this was a MUCH bigger project than I assumed going in. Bash, grep, sed, loads of additional toml files were not what I was expecting at the start!

ItsLemmy:
Thanks!

PR #671, opened 2025-11-06
Fix: Custom templates can use "is_dark_mode" and "mode" placeholders when using predefined color scheme

brunodsf05:
# Pull Request

## Motivation
I wanted to create a matugen user-template that uses the `{{ is_dark_mode }}` var. (You can see an example here https://github.com/InioX/matugen/wiki/Templates#conditionals)

So when testing I was testing my template using `{{ is_dark_mode }}`:
1. Generating the template from a **wallpaper** works.
2. Generating the template with **predefined color schemes** fails because matugen says var `{{ is_dark_mode }}` does not exists.

image

After investigating I found out that the json file passed in the executed command `matugen json <PATH>` only contained the colors but not the remaining vars.

## Type of Change
- [X] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring

## Related Issue

## Testing
- [X] Tested on compositor: (e.g., Hyprland, Sway, Niri)

## Screenshots / Videos

https://github.com/user-attachments/assets/f7eb9b3c-fbc1-4b7e-9e75-6fdf7bd43aaa

## Checklist
- [X] Self-reviewed my code

## Additional Notes


ItsLemmy:
We have reworked the AppTheming system so it's a bit easier to understand and less messy.
Even tho the error you got said "Matugen" it's wrong Noctalia does not use matugen at all for Predefined Color Scheme it uses a very basic serie of 'sed' commands to replace the color names by actual colors code. So it cannot process any "conditional statement".     We would have to replace 'sed' by either 'awk' or something more powerfull, that stays compatible with the matugen syntax.   Closing this for now.    Feel free to contribute another PR  if you find a good way to  do it.  Thanks!

PR #675, opened 2025-11-07
Nix: cleanup package

bokicoder:
# Pull Request

## Motivation
There are many unnecessary dependencies.

## Type of Change
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring

## Related Issue

## Testing

## Screenshots / Videos

## Checklist
- [x] Self-reviewed my code

## Additional Notes

### DIFF
--- a/nix/package.nix
+++ b/nix/package.nix
@@ -1,25 +1,16 @@
 {
   version ? "dirty",
   lib,
-  stdenv,
+  stdenvNoCC,
   # build
-  gcc,
   qt6,
   quickshell,
-  xkeyboard_config,
   # runtime deps
-  bash,
-  bluez,
   brightnessctl,
   cava,
   cliphist,
-  coreutils,
   ddcutil,
-  file,
-  findutils,
-  libnotify,
   matugen,
-  networkmanager,
   wlsunset,
   wl-clipboard,
   gpu-screen-recorder, # optional
@@ -33,6 +24,7 @@
     filter = path: type:
       !(builtins.any (prefix: lib.path.hasPrefix (../. + prefix) (/. + path)) [
         /.github
+        /.gitignore
         /Assets/Screenshots
         /Assets/Wallpaper
         /Bin/dev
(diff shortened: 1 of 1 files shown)


ixxie:
@bokicoder I sent you a message on Discord, Ill like to discuss the broader strokes of this and I thought it easier to do it that way.

I think there is a bit of a tension between supporting a setup which "just works" and providing a "no cruft" experience for more advanced users. I guess the approach you are advocated here is remove all optional dependencies and ask the user to configure this all.

This might be the right approach; I'd like to consider the alternative though. One approach I was thinking about, is what if the systemd services offered an option called `modules`:
- If you don't set `modules` (i.e. its `null`) we enable nothing; the user must manually control what they enable.
- If you set `modules.all = true` then we automatically enable all optional dependencies as services
- If you pass an attribute set you can configure in a more granular way, e.g.

```nix
  modules = {
    networking.enabled = true;
    power = {
      enabled = true;
      service = "tuned";
    };
  }; 
```

bokicoder:
These dependencies weren't useful to begin with. Whether I remove them or not, I still have to configure the relevant NixOS options myself. Noctalia doesn't handle this for users currently, so I believe we're now discussing a new feature.

bokicoder:
First, I don't think it's a good idea to re-wrap existing options. Second, while I could add these NixOS options to the nixos-module with default enabling (like `networking.networkmanager.enable = lib.mkDefault true;`), this cannot be done in the home-module.

ixxie:
> These dependencies weren't useful to begin with. Whether I remove them or not, I still have to configure the relevant NixOS options myself. Noctalia doesn't handle this for users currently, so I believe we're now discussing a new feature.

> First, I don't think it's a good idea to re-wrap existing options. Second, while I could add these NixOS options to the nixos-module with default enabling (like `networking.networkmanager.enable = lib.mkDefault true;`), this cannot be done in the home-module.

Okay, good points. Approved! Thanks again for this, good work :tada:

PR #678, opened 2025-11-07
BatteryManager: add support for Lenovo conservation mode

notiant:
# Pull Request

## Motivation

Many Lenovo laptops don't allow for setting a specific battery threshold (I think that's only possible with ThinkPads). Instead, there is a battery conservation mode (usually sets the threshold at 80%) that can be enabled. This PR adds `/sys/bus/platform/drivers/ideapad_acpi/VPC2004:00/conservation_mode` as a path for the battery manager to check in the installation process and gives the option to enable it in the battery panel. This should work for IdeaPads, but also for laptops from the Yoga and Legion series.

I also changed in the uninstallerProcess that `settings.data.battery.chargingMode = BatteryService.ChargingMode.Disabled`. I think the previous code was intended to save the last state in the settings, but it prevents the battery panel from showing the disabled screen.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring

## Testing
- [x] Tested on compositor: Hyprland

## Checklist
- [x] Self-reviewed my code


ItsLemmy:
We decided to completely remove the battery manager. Closing this for now. We may revisit this kind of feature later if we find a backend that works on all distros.

PR #680, opened 2025-11-08
feat(custom button): Add option to refresh CustomButton text on click

kurrycat2004:
# Pull Request

## Motivation
Similar to waybar, this PR simply adds an enabled-by-default per-click-command option to refresh the displayed text immediately when the button is clicked.
This allows much more responsive buttons, because you don't have to wait for the refresh interval.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring

## Testing

## Checklist
- [x] Self-reviewed my code

### DIFF
--- a/Modules/Bar/Widgets/CustomButton.qml
+++ b/Modules/Bar/Widgets/CustomButton.qml
@@ -213,7 +213,7 @@ Item {
     if (leftClickExec) {
       Quickshell.execDetached(["sh", "-c", leftClickExec])
       Logger.i("CustomButton", `Executing command: ${leftClickExec}`)
-    } else if (!hasExec) {
+    } else if (!hasExec && !leftClickUpdateText) {
       // No script was defined, open settings
       var settingsPanel = PanelService.getPanel("settingsPanel", screen)
       settingsPanel.requestedTab = SettingsPanel.Tab.Bar


lonerOrz:
Thanks so much for your implementation! I think it's a bit confusing as it stands. There are two modes for the custom buttons: one where clicking the button runs a command on its own, and another where the `textCommand` runs and displays something. So, I think we could improve it a bit:

* Move the option to update the button text immediately upon click into a separate setting.
* The specific settings should use checkboxes, and if a button is selected, hide the corresponding setting options for that button.

Just like that
图片

kurrycat2004:
@lonerOrz Well this isn't what the toggle actually does. Its not meant to replace the click-commands, it only *additionally* updates the text. 
Maybe as a reference: A specific usecase would be to implement a "Like" button for playing music. Clicking it would like the song, and then it would update the icon with a solid thumbs-up immediately because it instantly runs the update text command, which sets the icon based on the queried like-state.
So disabling the command settings for the button would be counterproductive.
I could move the buttons down, but as theyre meant as just an additional toggle you can enable for each setting, in my opinion its better when the toggles are placed on the side, but let me know if you think differently

lonerOrz:
> Maybe as a reference: A specific usecase would be to implement a "Like" button for playing music. Clicking it would like the song, and then it would update the icon with a solid thumbs-up immediately because it instantly runs the update text command, which sets the icon based on the queried like-state.

This is an interesting case; if this is to be implemented, the button should be turned off by default.
I also encountered some errors during testing:
```
  INFO qml: [20251111-173028] CompositorServ Display scales updated
  INFO qml: [20251111-173028] ProgramChecker Detected Discord clients: vesktop
  INFO qml: [20251111-173038]        Network Internet connectivity: full
  INFO qml: [20251111-173051] ProgramChecker Detected Discord clients: vesktop
  WARN scene: qs:/home/loner/Downloads/qs/shell/Modules/Panels/Settings/Bar/WidgetSettings/CustomButtonSettings.qml[129:-1]: ReferenceError: TooltipService is not defined
  WARN scene: qs:/home/loner/Downloads/qs/shell/Modules/Panels/Settings/Bar/WidgetSettings/CustomButtonSettings.qml[130:-1]: ReferenceError: TooltipService is not defined
  WARN scene: qs:/home/loner/Downloads/qs/shell/Modules/Panels/Settings/Bar/WidgetSettings/CustomButtonSettings.qml[83:-1]: ReferenceError: TooltipService is not defined
  WARN scene: qs:/home/loner/Downloads/qs/shell/Modules/Panels/Settings/Bar/WidgetSettings/CustomButtonSettings.qml[84:-1]: ReferenceError: TooltipService is not defined
  WARN scene: qs:/home/loner/Downloads/qs/shell/Modules/Panels/Settings/Bar/WidgetSettings/CustomButtonSettings.qml[106:-1]: ReferenceError: TooltipService is not defined
  WARN scene: qs:/home/loner/Downloads/qs/shell/Modules/Panels/Settings/Bar/WidgetSettings/CustomButtonSettings.qml[107:-1]: ReferenceError: TooltipService is not defined
```

kurrycat2004:
@lonerOrz should be fixed now, was just a missing import. i also changed it to be disabled by default

lonerOrz:
> @lonerOrz should be fixed now, was just a missing import. i also changed it to be disabled by default

LGTM！ But it would be better if it could support not setting commands and turning on the switch separately to update textCommand

lonerOrz:
I just found out that when textCommand is enabled in streaming mode, you should not be able to manually update the textCommand command.

kurrycat2004:
@lonerOrz i disabled the toggle and functionality when streaming mode is enabled and made it so the toggle works without a command being set

Modules/Bar/Widgets/CustomButton.qml
lonerOrz:
```diff
- } else if (!hasExec) {
+ } else if (!hasExec && !leftClickUpdateText) {
```

PR #688, opened 2025-11-09
Nix: refactor & improve home-module

bokicoder:
# Pull Request

## Motivation
Clean up the code, improve implementation robustness, and align with standard practices. This PR contains no new features or breaking changes.

## Type of Change
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Refactoring

## Related Issue

## Testing

## Screenshots / Videos

## Checklist
- [x] Self-reviewed my code

## Additional Notes
First, I removed "deep-merged with the default settings" because removing it did not cause any issues. Then, I replaced `builtins.toJSON` with the type and generate functions provided by `pkgs.formats.json`, which is the standard practice in Home Manager. The difference caused by this:
- Before 
图片
- After 
图片

### DIFF
--- a/nix/home-module.nix
+++ b/nix/home-module.nix
@@ -5,13 +5,14 @@
   ...
 }: let
   cfg = config.programs.noctalia-shell;
-  defaultSettings = builtins.fromJSON (builtins.readFile ../Assets/settings-default.json);
-  extractAttrs = x:
-    if builtins.isAttrs x
-    then x
-    else if builtins.isString x
-    then builtins.fromJson x
-    else builtins.fromJson (builtins.readFile x);
+  jsonFormat = pkgs.formats.json {};
+  generateJson = name: value:
+    if lib.isString value then
+      pkgs.writeText "noctalia-${name}.json" value
+    else if builtins.isPath value || lib.isStorePath value then
+      value
+    else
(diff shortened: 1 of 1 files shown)


Ly-sec:
CC: @Rexcrazy804

bokicoder:
> - disallows setting values as null <- breaking for anyone who has set this as null

I removed the null type because the original code didn't handle null values at all, and the option's default wasn't null either. So I figured that null was a mistake and took it out.

> - removes merge with defaultSettings <- breaks expected behavior for overriding defaults only

Why do we need to override defaults? I only specify non-default values in my config too, and as I mentioned, it hasn't caused any issues.

Rexcrazy804:
> So I figured that null was a mistake and took it out.

You are right, that was an oversight on my end when we worked on refactoring the old code into what it is now.

> I only specify non-default values in my config too, and as I mentioned, it hasn't caused any issues.

Could you confirm there are no relevant errors in the qs log? It does makes sense with how qs would work, but I'd prefer to be reassured.

bokicoder:
> Could you confirm there are no relevant errors in the qs log?

I can confirm.

Rexcrazy804:
we'll wait for ixxie to take a look (I don't use home-manager so he's better suited for testing this).
This should be ready to merge :)

PR #697, opened 2025-11-10
Background: add startup wallpaper transition

Sighthesia:
# Pull Request

## Motivation
Add wallpaper transition like changing wallpaper normally on shell startup for elegant experience.
Add option in the wallpaper setting tab for chosing transition types.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring

## Testing
- [x] Tested on compositor: Niri
- [ ] Tested with multiple monitors: it may behave ok, but i'm unable to test it becasue i only have one monitor
- [ ] Tesetd with different resolution: there's changes related to resolution changes, but i'm unable to test it becasue my monitor only supports one type of resolution model

## Checklist
- [x] Self-reviewed my code

## Additional Notes
1. The `function onDisplayScalesChanged()` was conflict with this feature, there's refactor on `recalculateImageSizes()` that calls nonexistent  `calculateSourceSize()`.
2. The `disc` type startup transition is fixed to center for better visual effect.

### DIFF
--- a/Assets/settings-default.json
+++ b/Assets/settings-default.json
@@ -120,7 +120,6 @@
     "randomIntervalSec": 300,
     "transitionDuration": 1500,
     "transitionType": "random",
-    "startupTransitionType": "disc",
     "transitionEdgeSmoothness": 0.05,
     "monitors": [],
     "panelPosition": "follow_bar"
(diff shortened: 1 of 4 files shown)


ItsLemmy:
I'm fine with having the transition on start, but I don't think this warrants a new setting. Just use the default IMO.   The Wallpaper settings are already pretty huge.

Sighthesia:
ok, I have removed the individual startup transition setting. It now uses the same transition effect as the standard wallpaper transition.

ItsLemmy:
Thanks!

PR #706, opened 2025-11-10
KeyboardLayout: update on "keyboard layout changed" events

3akev:
# Pull Request

## Motivation
Currently, KeyboardLayoutService polls the compositor periodically to update the current layout. This PR instead receives layout changes from the compositor's event stream and handles it directly. This is more responsive and less expensive.

Because Quickshell doesn't expose Sway `input` events, the only thing we can do is poll periodically for Sway.

## Type of Change
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Refactoring

## Testing

- [x] Tested on niri
- [x] Tested on Hyprland
- [x] Tested on sway

## Checklist
- [x] Self-reviewed my code

### DIFF
--- a/Services/Compositor/SwayService.qml
+++ b/Services/Compositor/SwayService.qml
@@ -4,6 +4,7 @@ import Quickshell.I3
 import Quickshell.Wayland
 import Quickshell.Io
 import qs.Commons
+import qs.Services.Keyboard
 
 Item {
   id: root
@@ -37,10 +38,12 @@ Item {
 
     try {
       I3.refreshWorkspaces()
+      I3.dispatch('(["input"])')
       Qt.callLater(() => {
                      safeUpdateWorkspaces()
                      safeUpdateWindows()
                      queryDisplayScales()
+                     queryKeyboardLayout()
                    })
       initialized = true
       Logger.i("SwayService", "Service started")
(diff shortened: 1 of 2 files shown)


ItsLemmy:
Excellent, thanks!

mikkurogue:
Just asking it here cause I think this might be causing the issue I'm facing but I now get a notification about the system setting my keyboard input from `fi`to `us` and back whenever I take an action like open any window etc. I have 2 keyboards, one is a split keyboard for all things typing and the other is my generic gaming keyboard that has a finnish layout and I set each of these as input devices themselves and manage their layouts in the hyprland device themsevles so i would only expect to see the notification whenever I type on the keyboard for that locale.

I didnt have this yesterday fyi, just after todays updates - is there then a way i can stop the notification all together otherwise?

ItsLemmy:
@3akev can probably add a setting to make the "Toast" conditional. Not too sure where tho, as we don't have a Keyboard tab in the settings.   Maybe all the way down in notifications tabs ?

3akev:
> @3akev can probably add a setting to make the "Toast" conditional. Not too sure where tho, as we don't have a Keyboard tab in the settings. Maybe all the way down in notifications tabs ?

Done, see https://github.com/noctalia-dev/noctalia-shell/pull/721. As an aside, it may be worth it to add a tab dedicated to toast settings. Right now it uses notification settings, which isn't explicit.

PR #732, opened 2025-11-12
Custom button: add on-wheel command support

kurrycat2004:
# Pull Request

## Motivation
Handling wheel events with custom buttons is useful for stuff like being able to change song volume by scrolling on a custom button

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring

## Testing

- [ ] Tested on niri
- [x] Tested on Hyprland
- [ ] Tested on sway

## Checklist
- [x] Self-reviewed my code

### DIFF
--- a/Assets/Translations/en.json
+++ b/Assets/Translations/en.json
@@ -141,6 +141,11 @@
           "label": "Middle click",
           "update-text": "Update displayed text on middle-click"
         },
+        "wheel": {
+          "description": "Command to execute when the scroll wheel is used.\nUse $delta for the scroll wheel delta in the command",
+          "label": "Scroll wheel",
+          "update-text": "Update displayed text on scroll"
+        },
         "parse-json": {
           "description": "Parse the command output as a JSON object to dynamically set text and icon.",
           "label": "Parse output as JSON"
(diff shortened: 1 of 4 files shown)


lonerOrz:
Here are a few ideas:

1. The default value of `$delta` should be set to `+/- 1` (currently it's `+/- 120`), and it should support arithmetic operations like `+`, `-`, `*`, `/` to modify the `$delta` value. For example, you could use: `testWheel.sh $delta*5`

2. Add a "separate mode" so that you can set commands for scrolling up and scrolling down independently.

3. The separate mode should also support the `$delta` logic, making it compatible with the rest.

kurrycat2004:
1. The 120 value depends on the driver/mouse/whatever and it makes sense to expose that value because it can be different for some people. I could add a "Raw wheel delta" toggle or something that is disabled by default and if it is, provide `signum(delta)` instead. 
For the arithmetic operations, this would be pretty complicated to handle correctly, as I'd need to have a way to escape all of that syntax too, its probably out of scope for this PR, but a nice idea to add.

2. Yeah, I can add this as another "Seperate wheel direction" toggle which shows two command inputs if its enabled

So you'd have two toggles, one for raw input and one to seperate the wheel dir. 
Let me know what you think

lonerOrz:
> 1. The 120 value depends on the driver/mouse/whatever and it makes sense to expose that value because it can be different for some people. I could add a "Raw wheel delta" toggle or something that is disabled by default and if it is, provide `signum(delta)` instead.

You can：
```
      let normalizedDelta = delta > 0 ? 1 : -1;
```

lonerOrz:
You can refer to this implementation https://github.com/lonerOrz/noctalia-shell/commit/ebb231fee321adc384d018d0e687a5507deef59c

lonerOrz:
Hello, are you still planning to submit this PR? I might submit the wheel-related feature implementation. If you intend to continue submitting, I can wait for your version.

kurrycat2004:
I dont really have time rn, so it would take a bit, you can submit yours and close this if you want

lonerOrz:
> I don't really have time rn, so it would take a bit; you can submit yours and close this if  you want

It's already implemented, you can try it！♥️

PR #739, opened 2025-11-13
NetworkService: re-add improved internet connectivity check

notiant:
I've seen this has been removed for too many false positives. I have changed how the connectivity is tested and added a redundancy check. It should now work as intended, but the parameters can be tweaked if there are problems again.
The main reason for this process is actually not the internet connectivity warning, but to keep the bar icon updated if a wifi network gets connected/disconnected without having the wifi panel open.

### DIFF
--- a/Services/Networking/NetworkService.qml
+++ b/Services/Networking/NetworkService.qml
@@ -18,6 +18,8 @@ Singleton {
   property bool ethernetConnected: false
   property string disconnectingFrom: ""
   property string forgettingNetwork: ""
+  property string networkConnectivity: "unknown"
+  property bool internetConnectivity: true
   property bool ignoreScanResults: false
   property bool scanPending: false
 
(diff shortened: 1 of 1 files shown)


ItsLemmy:
Thanks, merging and gonna give it a try.

ItsLemmy:
Any chance this could work with wired connection too ?
I have added the proper icon (ethernet-off) and proper check on the new service variable.   But ```nmcli networking connectivity check``` always returns full when wired so the "pings" are never triggered.

notiant:
I don't have an ethernet port so I can't test this myself, but can you check if your `20-connectivity.conf` file is configured properly and monitor the  journal with `journalctl -u NetworkManager -f`? How do you test the limited state exactly? When I block the internet access to my computer in my router settings (portal state), i get "NetworkManager state is now CONNECTED_SITE" in the journal.

You can get more logging details with `sudo nmcli general logging level DEBUG domains CONCHECK` and `sudo nmcli general logging level DEBUG domains ETHER`.

PR #742, opened 2025-11-13
MangoWC support

xtheeq:
# Pull Request

## Motivation
MangoWC a wayland compositor support.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring

## Testing

- [ ] Tested on niri
- [ ] Tested on Hyprland
- [ ] Tested on sway
- [x] Tested on Mangowc

## NOTE:
Requires `XDG_CURRENT_DESKTOP=mango` env to be set for detection.
In mango config it can be set as `env=XDG_CURRENT_DESKTOP,mango`

## Demo Video
https://github.com/user-attachments/assets/90cc665f-6551-4ec4-a500-1f64bc8b9bd0

### DIFF
--- a/Assets/settings-default.json
+++ b/Assets/settings-default.json
@@ -297,13 +297,7 @@
     "walker": false,
     "code": false,
     "spicetify": false,
-    "enableUserTemplates": false,
-    "discord_vesktop": false,
-    "discord_webcord": false,
-    "discord_armcord": false,
-    "discord_equibop": false,
-    "discord_lightcord": false,
-    "discord_dorion": false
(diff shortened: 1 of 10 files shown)


neonvoidx:
Switching tags doesn't seem to actually update the  tag selector
i.e if im on tag 2, tag 1 is still selected
image

Taskbar isn't correct, it shows all mango window clients in one pill
image

Tags in workspace bar don't actually update unless you restart quickshell

ItsLemmy:
Thank you!

PR #747, opened 2025-11-14
HostService: add user display name and use it in UI

### LINKED ISSUE #717: [Feature Proposal]: Use system real name (GECOS) in greeting instead of capitalized $USER
### Feature Description
I'd like Noctalia to display a more human-friendly name in its greetings, using the system real name (GECOS) or an explicit override (e.g. `NOCTALIA_REALNAME`), instead of just capitalizing the `$USER` login name.

### Why Is This Needed?
For example, with:

```bash
$ echo "$USER"
oluijks

$ getent passwd "$USER"
oluijks:x:1000:984:Olaf Luijks:/home/oluijks:/usr/bin/zsh
```

the lock screen / profile greeting shows:

```qml
text: I18n.tr("lock-screen.welcome-back") + " " +
      (Quickshell.env("USER").charAt(0).toUpperCase()
       + Quickshell.env("USER").slice(1)) + "!"
```

which results in Oluijks, instead of the more natural Olaf Luijks that's already present in the system account's GECOS field.

### Suggested Solutions

Proposal

Introduce a small User singleton that resolves a displayName like this:

1. Prefer an explicit override via an env var: `NOCTALIA_REALNAME`
2. Otherwise, read the GECOS "real name" via getent passwd "`$USER`"
3. Fall back to the current behaviour (capitalized $USER) if that fails

This keeps behaviour sane out of the box, but also lets users override the display name easily if they want something different from their system account.

Example Commons/User.qml

Something along these lines (trimmed for clarity):

```
// Commons/User.qml
pragma Singleton

import QtQuick
import Quickshell
import Quickshell.Io

Singleton {
    id: root

    readonly property string username: Quickshell.env("USER") || ""
    readonly property string envRealname: Quickshell.env("NOCTALIA_REALNAME") || ""

    // filled from getent on startup
    property string realName: ""

    readonly property string displayName: {
        if (envRealname.length > 0)
            return envRealname
        if (realName.length > 0)
            return realName
        if (!username || username.length === 0)
            return ""
        // current behaviour as fallback
        return username.charAt(0).toUpperCase() + username.slice(1)
    }

    Process {
        id: getentProc
        command: ["sh", "-c", "getent passwd \"$USER\" | cut -d: -f5 | cut -d, -f1"]
        running: true

        stdout: StdioCollector {
            onStreamFinished: {
                const name = text.trim()
                if (name.length > 0)
                    root.realName = name
            }
        }
    }
}
```

### Additional Context
Before I open a PR, I wanted to check if this approach is something you'd be interested in:

Are you okay with using getent (which is pretty standard on Linux), or would you prefer to only rely on environment variables?

Is `Commons/User.qml` a good place for this singleton, or would you rather see it live somewhere else?

Are you happy with the `NOCTALIA_REALNAME` override, or would you prefer to keep the API simpler and only use GECOS + fallback?

If you're generally on board with the idea, I can polish this up, replace the existing `Quickshell.env("USER")` usages with `User.displayName`, and submit a PR.

I'm not very familiar with the codebase yet, so I might overlook some relevant details.

Thanks for considering it

oluijks:
- Add username/envRealName/realName and displayName to HostService
- Resolve real name from `getent passwd "$USER"` with `NOCTALIA_REALNAME` override
- Use `HostService.displayName` on the lock screen and in the Control Center profile card

# Pull Request

## Motivation
The current lock screen and Control Center profile card display a capitalised `$USER` value, which is often not the user’s actual name. This PR introduces a single source of truth for the user’s display name in `HostService` and uses it across the UI.

The display name is resolved in the following order:
1. `NOCTALIA_REALNAME` environment variable (explicit override)
2. GECOS field from `getent passwd "$USER"` (real name)
3. Capitalised `$USER`
4. Final fallback: `"User"`

This improves the default experience on typical Linux setups while still allowing an explicit override via environment variable.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring

## Related Issue
- Closes #717

## Testing

Manual testing:

- Default setup (no `NOCTALIA_REALNAME` set):
  - Lock screen shows `Welcome back <real name>!` using `getent`’s GECOS field.
  - Control Center profile card shows the same real name.
- With `NOCTALIA_REALNAME="Some Other Name"` set:
  - Lock screen greeting uses the override.
  - Control Center profile card uses the override.

Checklist:

- [x] Tested on niri
- [ ] Tested on Hyprland
- [ ] Tested on sway

## Screenshots / Videos
If applicable, include screenshots or videos to help

## Checklist
- [x] Self-reviewed my code

## Additional Notes
If `getent` is unavailable or fails, `HostService.displayName` falls back to capitalised `$USER`, and finally to `"User"`, so behaviour remains sane even in minimal environments.

### DIFF
--- a/Modules/LockScreen/LockContext.qml
+++ b/Modules/LockScreen/LockContext.qml
@@ -2,6 +2,7 @@ import QtQuick
 import Quickshell
 import Quickshell.Services.Pam
 import qs.Commons
+import qs.Services.System
 
 Scope {
   id: root
@@ -40,7 +41,7 @@ Scope {
   PamContext {
     id: pam
     config: "login"
-    user: Quickshell.env("USER")
+    user: HostService.displayName
 
     onPamMessage: {
       Logger.i("LockContext", "PAM message:", message, "isError:", messageIsError, "responseRequired:", responseRequired)
--- a/Modules/LockScreen/LockScreen.qml
+++ b/Modules/LockScreen/LockScreen.qml
@@ -15,6 +15,7 @@ import qs.Services.Location
 import qs.Services.Media
 import qs.Services.Compositor
 import qs.Services.UI
+import qs.Services.System
 import qs.Widgets
 import qs.Widgets.AudioSpectrum
 
@@ -336,7 +337,7 @@ Loader {
 
                   // Welcome back + Username on one line
                   NText {
-                    text: I18n.tr("lock-screen.welcome-back") + " " + (Quickshell.env("USER").charAt(0).toUpperCase() + Quickshell.env("USER").slice(1)) + "!"
+                    text: I18n.tr("lock-screen.welcome-back") + " " + HostService.displayName + "!"
                     pointSize: Style.fontSizeXXL
                     font.weight: Font.Medium
                     color: Color.mOnSurface
--- a/Modules/Panels/ControlCenter/Cards/ProfileCard.qml
+++ b/Modules/Panels/ControlCenter/Cards/ProfileCard.qml
@@ -7,6 +7,7 @@ import Quickshell.Widgets
 import qs.Commons
 import qs.Modules.Panels.ControlCenter.Cards
 import qs.Modules.Panels.Settings
+import qs.Services.System
 import qs.Services.UI
 import qs.Widgets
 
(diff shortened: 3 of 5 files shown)


ItsLemmy:
Excellent, thanks!

PR #756, opened 2025-11-15
Add wallpaper selector to settings

notiant:
I thought there probably should be an option to choose the current wallpaper directly from the wallpaper settings instead of just from the control center or the bar widget.


Ly-sec:
Good job! Thank you very much :heart:

PR #762, opened 2025-11-15
Add index + name option to the combo for the Workspace widget label mode

dmolin:
# Pull Request

## Motivation

Adds the new “Index + Name” workspace label mode throughout the UI. Currently we can choose to display the workspaces labels either as their index or as their name, but not both.  This additional option allows us to have both.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring

## Testing

- [x] Tested on niri
- [ ] Tested on Hyprland
- [ ] Tested on sway

## Screenshots / Videos

workspace-number-name-1

workspace-number-name-2

## Checklist

- [x] Self-reviewed my code

## Additional Notes

As for the bar vertical layout: I feel the workspace widget should ignore both the "name" and "index-name" options, since they would never render properly. Right now it does not, so even choosing "Name" will result in a bit of a mess. I didn't change this behaviour but if you agree, I could patch it too, so when showing the bar left or right we would ignore both the label modes "name" and the new "index+name" (and we would always display the workspace pills either as empty or with the index only)

### DIFF
--- a/Modules/Bar/Widgets/Workspace.qml
+++ b/Modules/Bar/Widgets/Workspace.qml
@@ -73,9 +73,7 @@ Item {
 
     // For name mode, calculate width based on actual text content
     if (["name", "index-name"].includes(labelMode) && ws.name && ws.name.length > 0) {
-      const wsName = ws.name.substring(0, characterCount)
-      // build displayText based on labelMode. if index-name, include index
-      const displayText = labelMode === "index-name" ? ws.idx.toString() + " " + wsName : wsName
+      const displayText = root.getWorkspaceLabel(ws)
       const textWidth = displayText.length * (d * 0.4) // Approximate width per character
       const padding = d * 0.6 // Padding on both sides
       return Math.max(d * factor, textWidth + padding)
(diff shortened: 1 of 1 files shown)


ItsLemmy:
How does it look in a vertical bar ?

dmolin:
@ItsLemmy  This is how it currently looks in vertical mode:

Name only:
screenshot-20251116-125950

Index + Name:
screenshot-20251116-125914

If we want to keep "name", my suggestion would be to allow "name" (and "index") when in vertical mode; if "index + name" was chosen, we would fallback to "index"...

ItsLemmy:
Thanks a lot for this PR!   We really appreciate you taking the time to tackle this issue.

As it turns out, another contributor submitted a similar fix. We ended up merging the other one because it also unified the behavior of the TaskbarGrouped and Workspace widgets.

It's always a shame when work overlaps like this, and we're sorry for the conflict. Your approach was great, and we really hope we'll see more contributions from you in the future!

MrDowntempo:
I am the culprit. I didn't see your work when I was doing something similar. My apologies. And as Lemmy said, we hope you keep contributing, and I'll do my best not to step on any more toes!

dmolin:
Sure, no problem at all! 😃

PR #770, opened 2025-11-16
Hide scrollbar in NListView if it's not necessary

notiant:
In the sidebar of the settings panel there is now always a scrollbar visible when hovering over the right edge, even if there are not enough elements in the bar to make scrolling necessary. This PR hides the scrollbar in NListView completely if there are not enough elements a list.
I also reversed a recent change in the vertical spacing of the sidebar tabs, i think it looks to crowded this way.


Ly-sec:
Thank you very much for the PR, I'll merge that one :heart:.

PR #811, opened 2025-11-21
Show workspaces from the focused screen

shouya:
# Pull Request

## Motivation

I have two monitors and only want to show the bar on one of them. However, the Workspace bar widget only shows the workspaces from the same screen as the bar. I expect there is a way to see the workspaces from the other screen as well.

This PR adds a toggle that changes the behavior of showing only the workspaces from the same screen as the bar into showing the workspaces from the focused screen, allowing either screen's workspaces to be shown by switching focus between the screens.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring

## Testing

- [x] Tested on niri
- [ ] Tested on Hyprland
- [ ] Tested on sway

## Screenshots / Videos
image

## Checklist
- [x] Self-reviewed my code


ItsLemmy:
Good idea! 
But you are missing an important part in this PR. Widgets settings and their default value must be declared in Services/UI/BarWidgetRegistry.qml    (the default should be false to not disturb existing users)

Also the english description should be as follow : 
label:   Follow Focused Screen
description:   Display workspaces from the currently focused screen, rather than the screen where the bar is located.

ItsLemmy:
Thank you!

PR #824, opened 2025-11-22
Implement emoji picker

### LINKED ISSUE #803: emoji and character selection
### Feature Description
emoji and character selection

### Why Is This Needed?
for easy access like rofimoji 

### Suggested Solutions
a button that lets you select emoji/characters which are subdivided into groups

### Additional Context
inspired by rofimoji

lonerOrz:
# Pull Request

## Motivation
Implement emoji picker

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring

## Related Issue
- Closes #803 

## Testing

- [x] Tested on niri
- [ ] Tested on Hyprland
- [ ] Tested on sway

## Screenshots / Videos
图片

## Checklist
- [ ] Self-reviewed my code

## Additional Notes

### DIFF
--- a/Assets/Translations/de.json
+++ b/Assets/Translations/de.json
@@ -670,7 +670,11 @@
     "clipboard-history-disabled-description": "Zwischenablageverlauf in den Einstellungen aktivieren oder cliphist installieren",
     "clipboard-loading": "Lade Zwischenablageverlauf...",
     "clipboard-loading-description": "Bitte warten",
-    "clipboard-search-description": "Zwischenablageverlauf durchsuchen"
+    "clipboard-search-description": "Zwischenablageverlauf durchsuchen",
+    "emoji": "Emoji-Auswahl",
+    "emoji-search-description": "Emojis suchen und kopieren",
+    "emoji-loading": "Lade Emojis...",
+    "emoji-loading-description": "Bitte warten"
   },
   "quickSettings": {
     "bluetooth": {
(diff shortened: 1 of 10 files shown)


Ly-sec:
I really like this however, would it not make sense to use a different json? https://github.com/muan/unicode-emoji-json has a few json files with almost all emojis, so we wouldn't have to update it as much you know.

lonerOrz:
> I really like this however, would it not make sense to use a different json? [muan/unicode-emoji-json](https://github.com/muan/unicode-emoji-json?rgh-link-date=2025-11-22T12%3A02%3A14Z) has a few json files with almost all emojis, so we wouldn't have to update it as much you know.

I feel like the JSON primary keys in the full files other people provide are hard to manage, and we don’t really need to support every single emoji—just the common ones are enough. If users want to add more emojis, they can just add their own `emoji.json` in their user config. I’ve also handled duplicates. I did think about allowing imports directly in the settings, but I haven’t figured out the best way to do that yet.

Ly-sec:
> > I really like this however, would it not make sense to use a different json? [muan/unicode-emoji-json](https://github.com/muan/unicode-emoji-json?rgh-link-date=2025-11-22T12%3A02%3A14Z) has a few json files with almost all emojis, so we wouldn't have to update it as much you know.
> 
> I feel like the JSON primary keys in the full files other people provide are hard to manage, and we don’t really need to support every single emoji—just the common ones are enough. If users want to add more emojis, they can just add their own `emoji.json` in their user config. I’ve also handled duplicates. I did think about allowing imports directly in the settings, but I haven’t figured out the best way to do that yet.

You're right, I'll merge :) let's keep it simple.

PR #827, opened 2025-11-22
LockScreen: make 'hibernate' optional

notiant:
Since not every system supports hibernation, i think displaying this action on the lock screen should be optional. I also made the weather forecast show 4 days if no media is playing since it looks very stretched out otherwise.


Ly-sec:
This absolutely makes sense, thank you very much!

PR #834, opened 2025-11-23
Some changes for the notification layout & localize timestamps

notiant:
This PR makes some changes to the notification layout:

- add a bit more horizontal margins (also applies to toasts)
- reduce the size for notifications without actions
- change the timestamp format and offer localization

I also fixed some translation issues with "suspend" and "hibernate" in German.


Ly-sec:
Thank you very much :heart: !

PR #840, opened 2025-11-23
Add bar (inline) position for notifications

zoriya:
# Pull Request

## Motivation

I like my notifications to not pop on top of some other window, for that purpose i always include notifications as part of the bar. When switiching to noctalia, i found this missing so here's my PR for it.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring

## Related Issue
/

## Testing

- [x] Tested on niri
- [ ] Tested on Hyprland
- [ ] Tested on sway

Idk how to make this work with vertical bars, should it just noop?

## Screenshots / Videos

It looks like this: 

https://github.com/user-attachments/assets/9adbf992-3330-4013-bd41-d52b3dcbea3b

## Checklist
- [ ] Self-reviewed my code

## Additional Notes

This is my first time working with qml, let me know if i did anything wrong!

### DIFF
--- a/Modules/Toast/ToastScreen.qml
+++ b/Modules/Toast/ToastScreen.qml
@@ -134,7 +134,11 @@ Item {
       screen: root.screen
 
       // Parse location setting
-      readonly property string location: Settings.data.notifications?.location || "top_right"
+      readonly property string location: {
+        if (Settings.data.notifications?.location == "bar")
+          return "top_right"
+        return Settings.data.notifications?.location || "top_right"
+      }
       readonly property bool isTop: location.startsWith("top")
       readonly property bool isBottom: location.startsWith("bottom")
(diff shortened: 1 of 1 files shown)


ItsLemmy:
How does it look/behave with a vertical bar? can you provide a video.

zoriya:
I didn't implement the vertical bar precisely because i didn't know what to do there, i don't think having vertical text make much sense.

How would you make it usable there?

ItsLemmy:
That's why I asked, because you ticked the box "Tested with different bar positions and density settings".

Indeed I don't see a way to have this working with a vertical bar.
At this point we don't want to add features that are limited to specific bar positions. You may want to keep this in your private fork.

PR #850, opened 2025-11-23
LockScreen: Re-add some reverted changes

notiant:
I've seen that some changes to the lock screen I've made in a recent commit have been replaced by the older version shortly after again.

### DIFF
--- a/Modules/LockScreen/LockScreen.qml
+++ b/Modules/LockScreen/LockScreen.qml
@@ -343,13 +343,13 @@ Loader {
                     text: {
                       var lang = I18n.locale.name.split("_")[0];
                       var formats = {
-                        "en": "dddd, MMMM d",
                         "de": "dddd, d. MMMM",
-                        "fr": "dddd d MMMM",
+                        "en": "dddd, MMMM d",
                         "es": "dddd, d 'de' MMMM",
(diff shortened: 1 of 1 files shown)


Modules/LockScreen/LockScreen.qml
```diff
@@ -525,7 +525,7 @@ Loader {
                 }
                 Text {
                   id: hibernateText
-                  text: I18n.tr("session-menu.hibernate")
+                  text: Settings.data.general.showHibernateOnLockScreen ? I18n.tr("session-menu.hibernate") : ""
```
Ly-sec, line 528: Not sure why we'd need this if we don't even show the button in the first place?

Ly-sec (commented):
Hey there, there are 2 things that need fixing/explaining :).

I noticed a problem: `ERROR:   caused by @Modules/LockScreen/LockScreen.qml[352:25]: Expected token `}'`

Modules/LockScreen/LockScreen.qml
notiant: Fixed the syntax error. I added that line because the hibernate text would still be used for the calculation of the minimum button width even if it's not shown.

PR #862, opened 2025-11-24
Remove minimum height from wifi & bluetooth panel

notiant:
This removes the empty space if there aren't enough devices to fill the the panel.

Screenshot_20251124_17h27m11


ItsLemmy:
Simple and efficient, nice!

PR #880, opened 2025-11-26
Add some missing translations & more consistency for English

notiant:
This adds missing translations for the connection status in the bluetooth panel and the dark mode toast. I also went over the English language file and tried to make it more consistent in terms of capitalization of words.

I also found that in BluetoothDevicesList.qml, there is a function for signal strength. However, that property doesn't seem to exist according to the quickshell documentation. Is this supposed to be working?


ItsLemmy:
Thanks

PR #902, opened 2025-11-28
fix recurrent events in calendar

AdrienPiechocki:
# Pull Request

## Motivation
The calendar widget didn't show recurrent events correctly. For example, birthdays and monthly or weekly events were not repeated every year/month/week. If I had my birthday set to the year 2000, gnome-calendar would repeat it every year but not noctalia-shell's calendar widget.

Now, it does.

HOWEVER, when testing the events settings in gnome-calendar, some recurrence types would not work as intended.
For example, "Monday to Friday events" are seen as "weekly events" by ICalGLib which causes inconsistencies between gnome-calendar and noctalia's widget...
If somebody knows how to fix this, please do :)

## Type of Change

- [X] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring

## Related Issue
Didn't find anyone talking about this.

## Testing

- [ ] Tested on niri
- [X] Tested on Hyprland
- [ ] Tested on sway
- [X] Tested with different bar positions and density settings

## Screenshots / Videos

## Checklist
- [X] Self-reviewed my code

## Additional Notes
None.


ItsLemmy:
@ThatOneCalculator do you have some free time to review and test that?

ThatOneCalculator:
I can test this week!

ThatOneCalculator:
Also @AdrienPiechocki could you remove the multiline comments you made, especially the test ones?

Edit: ty :)

AdrienPiechocki:
crap.. I messed something up. Was making another change (unrelated to this one) and pushed it on this pull request

AdrienPiechocki:
OK. I fixed it ! Sorry about that...

Ly-sec:
Since I don't really have recurrent events, what's the current state of this PR if I may ask? @ThatOneCalculator @AdrienPiechocki

AdrienPiechocki:
Sure, no problem ! For reference, here is the calendar app I use (gnome-calendar) with this month's events : 

GNOME Calendar

There are 3 recurrent events here. Two of them are my friends' birthdays (12 of December) and the other one is on the 28th. None of these are shown in the non-fixed calendar widget because it only gets the initial date of these events (which means I have the birthdays only shown at the initial date of birth : in 2003). As shown here : 

Without fix

With my fix, events are registered and their recurrence state makes them duplicate every year/month/week/...  
As shown here : 

With fix

You can see the difference between the two last screenshots by checking the event markers (pink dots) at each day where events are supposed to be, according to the first screenshot.

Hope this helps ! :)

ThatOneCalculator:
For some reason, it's not showing any of my recurring events

AdrienPiechocki:
That's weird :/ 
What calendar app do you use ? I use gnome-calendar. 
Also, you may want to wait for `~/.cache/noctalia/calendar.json` to reload (it takes some time).  You can check the value of "lastUpdate" at the end of the file to make sure it updated recently.

ThatOneCalculator:
I'm on gnome calendar too, and the recurring events have been in my calendar for months

AdrienPiechocki:
That is even weirder...
You mean recurring events have been working for you before I updated the script ? 
I'll triple check the code I made tomorrow but I don't see why it isn't working for you. Sorry !

I doubt it is related to the issue but just in case I'll share my setup here : 
OS: Arch Linux x86_64
Kernel: Linux 6.17.9-arch1-1
WM: Hyprland 0.52.2 (Wayland)
everything is up to date

ThatOneCalculator:
Pretty much the same setup as you but with Niri and 6.18 :P

AdrienPiechocki:
So, I think I tried everything but can't find why `calendar-events.py` (the one without my fix) doesn't show recurring events on my side but does on yours... 

I tried multiple things and found out it isn't an Hyprland specific issue ; I installed Niri with a fresh Arch install on a virtual machine and I  still have the issue without my fix.

I know `evolution-data-server` (which is required for calendar events to work) is installed and properly configured because when starting noctalia from terminal shows `Calendar Service started` and `Calendar EDS libraries available` when opening the calendar.

@ThatOneCalculator I have a few questions : 
- Can you confirm the calendar widget gets all events directly from `~/.cache/noctalia/calendar.json` ? I know `CalendarService.qml`  generates this file via the related python scripts but I don't know if there is another service related to getting events for the calendar widget. If so, maybe that is why I can't get events outside the content of `~/.cache/noctalia/calendar.json`

- When you said : "recurring events have been in my calendar for months", do you mean they aren't anymore ? even when you revert back from my fix ? If so, the issue might be somewhere else (maybe?)

- Could you tell me the content of `~/.cache/noctalia/calendar.json` with AND without my fix. Or at least tell me if there are any differences. What about recurring events ; are they duplicated properly (with or without my fix) ?

I triple checked my code and just cannot see how it would cause issues. Which means this PR is only useful for very specific cases (mine, at least) where recurring events wouldn't show up in the calendar widget. 
However, please tell me if it did cause an issue. 

Thank you for your time !

ThatOneCalculator:
1. Interestingly, it shows up in ` ~/.cache/noctalia/calendar.json` but under `cachedEvents` as a regular event, not repeating
2. By "recurring events have been in my calendar for months", I meant it was in my Gnome Calendar/Google Calendar for months, not Noctalia
3. There seems to be no differences, I'll have to do some investigation

AdrienPiechocki:
Okay, thank you for your answers. 
Your first answer was just why I made this fix. But if the new script doesn't work for you, that is strange.

Something I have been doing from the beginning is removing the content of  `~/.cache/noctalia/calendar.json`  to force a reload. So maybe you can try this : 
1. Delete (or cut to somewhere else) the content inside `~/.cache/noctalia/calendar.json`.
2. Click on the calendar widget (this should restart `CalendarService`)
3. Wait a few seconds for content in `~/.cache/noctalia/calendar.json` to show up (`cachedEvents` may start as an empty list while it fetches the events).
4. Reopen the calendar widget : every event should be here, even the recurring ones. (`cachedEvents` should look bigger than before since it literally duplicates recurring events)

If that fixes it for you, it means reloading via `CalendarService` doesn't work properly.

ItsLemmy:
confirmed as working by another discord user, will see how it goes.

PR #907, opened 2025-11-29
Prevent empty bar pills from expanding

notiant:
This change is intended to prevent the wifi and bluetooth bar widget to expand when hovered but no networks/devices are connected (for horizontal bar positions).


ItsLemmy:
Thanks, I also noticed that earlier but had not time to check.

PR #950, opened 2025-12-02
Add helper packages for live development in NixOs

sheijk:
# Pull Request

## Motivation
I found using quickshell's live development capabilities in NixOs need some non-obvious setup. This adds two small helper scripts for easier live development under NixOs.

noctalia-live-devel will shutdown the systemd service for noctalia and start quickshell with the noctalia-shell linked into ~/.config/quickshell/noctalia-shell so live updates can take places. Once qs is exited using C-c the systemd service will be started, again.

qs-noctalia is a drop in replacement for noctalia-shell which detects whether the live server is running and will dispatch all commands to qs with the noctalia config or to noctalia-shell. I install this in NixOs and configure Niri to call qs-noctalia instead of noctalia-shell directly so shortcuts keep working with the live dev server.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring

## Testing
Installed noctalia-shell in NixOs, linking my noctalia-shell repo into ~/.config/quickshell/noctalia-shell and editing a few things in noctala like the bar and lock screen and using this for while.

- [x] Tested on niri
- [ ] Tested on Hyprland
- [ ] Tested on sway

## Checklist
- [x] Self-reviewed my code

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -18,6 +18,8 @@
     packages = eachSystem (
       system: {
         default = pkgsFor.${system}.noctalia-shell;
+        noctalia-live-devel = pkgsFor.${system}.callPackage ./nix/noctalia-live-devel.nix {pkgs = pkgsFor.${system};};
+        qs-noctalia = pkgsFor.${system}.callPackage ./nix/qs-noctalia.nix {pkgs = pkgsFor.${system};};
       }
     );
 
--- a/nix/noctalia-live-devel.nix
+++ b/nix/noctalia-live-devel.nix
@@ -1,0 +1,9 @@
+{pkgs, ...}:
+pkgs.writeShellScriptBin "noctalia-live-devel" ''
+  echo "Stopping noctalia-shell service and starting qs"
+  echo "Use qs-noctalia ipc to invoke the dev server or regular service"
+  [ -d ~/.config/quickshell/noctalia-shell ] || echo "error: noctalia-shell repo needs to be in ~/.config/quickshell"
+  systemctl stop --user noctalia-shell
+  trap 'systemctl start --user noctalia-shell; echo "Restarted noctalia-shell service"' EXIT
+  qs -p ~/.config/quickshell/noctalia-shell
+''
--- a/nix/qs-noctalia.nix
+++ b/nix/qs-noctalia.nix
@@ -1,0 +1,8 @@
+{pkgs, ...}:
+pkgs.writeShellScriptBin "qs-noctalia" ''
+  if qs list --all|grep /home/|grep noctalia-shell >/dev/null; then
+  qs -c noctalia-shell "$@"
+  else
+  noctalia-shell "$@"
+  fi
+''
--- a/nix/shell.nix
+++ b/nix/shell.nix
@@ -1,4 +1,5 @@
 {
+  pkgs,
   quickshell,
   alejandra,
   statix,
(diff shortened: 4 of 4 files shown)


Ly-sec:
CC: @bokicoder

bokicoder:
I don’t want NixOS to be even more idiosyncratic. What you perceive as help might seem like total nonsense to others — those who want to contribute are free to do so. We only need to provide the tools, not anticipate every possible scenario for them or dictate how they should proceed.

sheijk:
@bokicoder I'm not sure I get you. You think this is nonsense? Or some hypothetical other person? These are convenience tools for development to make it easier to work on Noctalia on NixOs. They won't be used by users or change how noctalia-shell works on NixOs. Maybe I'm missing the right workflow here. How is noctalia-shell development on NixOs usually done?

bokicoder:
This kind of thing doesn’t exist in other distributions, and it’s not standard practice even in NixOS. I’ve never seen it in any other flake-based repo.

sheijk:
It's just a helper script like you'd normally have somewhere in a tools or script dir. These might be useful for other distributions as well. I could update this so they're stand-alone scripts in Bin/dev if you're interested in it then?

bokicoder:
> It's just a helper script like you'd normally have somewhere in a tools or script dir. These might be useful for other distributions as well. I could update this so they're stand-alone scripts in Bin/dev if you're interested in it then?

I have no problem with this approach — let’s see what @Ly-sec has to say.

Ly-sec:
I'm not really sure how useful this would be given we'd have to check for multiple different shell locations (basically no package installs it to `~/.config/quickshell/noctalia-shell`. For example we install it in `/etc/xdg/quickshell/noctalia-shell/` on arch and who knows where other maintainers install it. Also I kind of assume that people who want to work on noctalia-shell are capable of simply restarting it. I'm unsure if it's necessary, what do you think @ItsLemmy ?

sheijk:
Ah so developers are doing changes and then restarting the server each time? I mostly found noctalia due to quickshell's ability to reload on-the-fly which for me makes it very productive to work on. If this workflow is not used by noctalia-shell developers this whole helper won't be very useful.

The path ~/.config/quickshell/noctalia-dev is mostly because 'qs -c noctalia-shell' will work that way w/o needing to know the full path when invoking quickshell/noctalias ipc. I believe the noctalia-shell binary on is a work-around for this so the full nix store path doesn't need to be passed. Or am I mistaken there?

bokicoder:
> Ah so developers are doing changes and then restarting the server each time?

Not really.

bokicoder:
> The path ~/.config/quickshell/noctalia-dev is mostly because 'qs -c noctalia-shell' will work that way w/o needing to know the full path when invoking quickshell/noctalias ipc. I believe the noctalia-shell binary on is a work-around for this so the full nix store path doesn't need to be passed. Or am I mistaken there?

I wouldn’t want to put a Git repo in such a deep path. When developing, I’d cd into this directory, so I’d use the  `qs -p .`  command.

sheijk:
I just symlink it to the repo I'm working on. How do you use the ipc while running the development build? If I run qs like that my keybindings configured in Niri will stop working (as they use noctalia-shell) and I don't see a way to do something like 'noctalia-shell ipc call launcher toggle', 'qs -c noctalia-shell ...' only works with this symlink in ~/.config/quickshell for me

Ly-sec:
So what's the plan here @bokicoder ? Do you think its reasonable to be added or nah?

bokicoder:
> So what's the plan here @bokicoder ? Do you think its reasonable to be added or nah?

Nah.

Ly-sec:
I'll close this one then, thank you for your effort though :heart: !

PR #890, opened 2025-11-27
add brightness controls and noctalia performance mode switch to battery panel

art0rz:
# Pull Request
- Adds brightness controls to the BatteryPanel
- Adds Noctalia performance mode switch to BatteryPanel

## Motivation
On my small laptop screen, there were too many icons on my bar. I think these options can belong to the BatteryPanel

## Type of Change
Mark the relevant option with an "x".
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring

## Testing
Describe how you tested your changes and mark the relevant items.

- [x] Tested on niri
- [ ] Tested on Hyprland
- [ ] Tested on sway
- [ ] Tested with different bar positions and density settings
- [ ] Tested at different interface scaling values
- [ ] Tested with multiple monitors (if applicable)

## Screenshots / Videos
image

## Checklist
- [x] Code follows project style guidelines
- [x] Self-reviewed my code
- [x] No new warnings or errors
- [ ] Documentation or comments updated (if relevant)

PR #956, opened 2025-12-03
BatteryPanel: add the power profiles option back

notiant:
# Pull Request

## Motivation
The option to choose the current power profile was recently removed from the battery panel for being redundant. Personally, I don't like having an extra widget just for changing power profiles and the battery panel is a natural place for this option since it directly impacts the remaining battery time.
I also added a toggle for the Noctalia performance mode, although not sure about that.
Both controls could be made optional like in PR #890.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring

## Testing

- [ ] Tested on niri
- [x] Tested on Hyprland
- [ ] Tested on sway

## Screenshots / Videos
Screenshot_20251203_08h41m49

## Checklist
- [x] Self-reviewed my code

### DIFF
--- a/Modules/Panels/Battery/BatteryPanel.qml
+++ b/Modules/Panels/Battery/BatteryPanel.qml
@@ -339,7 +339,7 @@ SmartPanel {
           id: controlsLayout
           anchors.fill: parent
           anchors.margins: Style.marginL
-          spacing: Style.marginL
+          spacing: Style.marginM
 
           ColumnLayout {
             visible: root.powerProfileAvailable && root.showPowerProfiles


Ly-sec:
Hey, I would be okay with bringing it back but only if it's disabled by default, mostly because we already have the power profiles in the control center by default :p. Also is there any reason why you're using a slider for the power profiles?

notiant:
OK, then I will add the widget options.
For the slider, I just used the deleted code, I guess it's inspired by the UI of Windows 10 and KDE Plasma which also use a slider for the power profiles. I could change it to an icon button or maybe just make the slider thinner?

Ly-sec:
> OK, then I will add the widget options. For the slider, I just used the deleted code, I guess it's inspired by the UI of Windows 10 and KDE Plasma which also use a slider for the power profiles. I could change it to an icon button or maybe just make the slider thinner?

Hmm I'm not sure, perhaps use the same style that we got on the control center or do you think that wouldn't look nice?

notiant:
@Ly-sec This PR is now ready, I've added the widget settings. I kept the slider layout for the energy profiles but added icons to have it make more sense:

Screenshot_20251205_12h00m39

ItsLemmy:
Thanks!

PR #969, opened 2025-12-04
fix(compositor): add z-order handling for focus in hyprland

ahmet-cetinkaya:
### 🚀 Motivation and Context

In Hyprland float mode, clicking on application icons in the taskbar does not bring windows to the front. The window remains in the background while the click action appears to do nothing visually. This affects workflow as users cannot restore or focus windows by clicking taskbar icons.

This PR fixes the window focusing behavior by:
1. Adding proper z-order handling using modern `alterzorder` command
2. Updating window focus logic to handle float mode stacking

### 📋 Checklist for Reviewer

- [x] Tested on Hyprland
- [x] Commit history is clean and descriptive
- [x] Code quality standards were met (e.g., linter passed)
- [x] Documentation updated (if applicable)

### DIFF
--- a/Services/Compositor/HyprlandService.qml
+++ b/Services/Compositor/HyprlandService.qml
@@ -442,12 +442,8 @@ Item {
       }
 
       const windowId = window.id.toString();
-
-      // Focus the window
       Hyprland.dispatch(`focuswindow address:0x${windowId}`);
-
-      // Bring the focused window to the top (essential for Float Mode)
-      Hyprland.dispatch(`alterzorder top,address:0x${windowId}`);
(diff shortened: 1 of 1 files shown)


Ly-sec:
Just to make sure that I understand that correctly... when a window is floating ("free mode?") and you are focused on a different window and then click on the window in the taskbar, it doesn't get focus? If that's what you mean, I can't replicate that issue at all. Just tested it and it for sure did get focus.

ahmet-cetinkaya:
@Ly-sec Before the fix, when the two windows were stacked in Hyprland and I clicked the app on the taskbar, it did focus the window, but the window didn’t come to the front.

ItsLemmy:
Thanks

PR #985, opened 2025-12-06
DockMenu: use uiScaleRatio for width calculation

notiant:
This is necessary because Text instead of NText is used.


Ly-sec:
Hey there, I think we could also just switch to `NText`, must've missed that.

notiant:
Yes, this works as well, I've just changed it.

Ly-sec:
Thank you very much :heart: !

PR #1007, opened 2025-12-08
Tooltip corrections

notiant:
The recent tooltip changes had some entries that were only applied to English. I also don't think that "Close button" is a better tooltip than just "Close", not every tooltip must describe an object :)


Ly-sec:
Thank you very much and yeah you're right, I went a bit overboard :p!

PR #1012, opened 2025-12-09
fix(notification): respect performance mode when playing notification sounds

ahmet-cetinkaya:
### 🚀 Motivation and Context

Notification sounds were playing before checking the system's performance mode setting. This caused notification sounds to play even when performance mode was enabled.

The changes reorder the notification handling logic to respect the performance mode setting by:
- Moving the performance mode check to occur immediately after history tracking
- Ensuring no notification processing (including sound) occurs when performance mode is active
- Calling playNotificationSound() only after all validation checks pass
- Maintaining proper rate limiting and exclusion logic within the sound function

### 🔗 Related Topics

- #935

### 📋 Checklist for Reviewer

- [x] Tests passed locally.
- [x] Commit history is clean and descriptive.
- [x] Documentation updated (if applicable).
- [x] Code quality standards were met (e.g., linter passed).

### DIFF
--- a/Services/System/NotificationService.qml
+++ b/Services/System/NotificationService.qml
@@ -209,26 +209,16 @@ Singleton {
     playNotificationSound(data.urgency, notification.appName);
   }
 
-  // Function to play notification sound using existing SoundService
   function playNotificationSound(urgency, appName) {
-    // Rate limiting - prevent sound spam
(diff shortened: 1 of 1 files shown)


Ly-sec:
Thank you very much :heart: !

PR #1058, opened 2025-12-15
Fix linebreaks in labels & some punctuation corrections

notiant:
I noticed that linebreaks in NLabels don't work anymore since changing the text format to `StyledText`. This PR replaces instances of `\n` with `<br>` to fix this issue and also addresses some punctuation inconsistencies in the language files.


Ly-sec:
Thank you very much :heart: !

PR #1064, opened 2025-12-15
feat(niri): adjust window sorting logic

GoodbyeNJN:
# Pull Request

## Motivation

This PR adjusts the window ordering logic under the Niri compositor to be more intuitive for users.

The specific sorting rules are as follows:

1. First, sort by monitor position
2. Next, sort by workspace order
3. Then, sort by window position
4. Finally, sort by window ID to ensure consistent ordering

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring

## Related Issue

Nope

## Testing

- [x] Tested on niri
- [ ] Tested on Hyprland
- [ ] Tested on sway

## Screenshots / Videos

Nope

## Checklist
- [x] Self-reviewed my code

## Additional Notes

Nope

### DIFF
--- a/Services/Compositor/NiriService.qml
+++ b/Services/Compositor/NiriService.qml
@@ -22,11 +22,15 @@ Item {
   signal windowListChanged
   signal displayScalesChanged
 
+  property var outputCache: ({})
+  property var workspaceCache: ({})
+
   function initialize() {
     niriEventStream.connected = true;
     niriCommandSocket.connected = true;
 
     startEventStream();
+    updateOutputs();
     updateWorkspaces();
     updateWindows();
     queryDisplayScales();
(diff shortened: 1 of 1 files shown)


ItsLemmy:
The motivation behind this PR is very vague. What real problem is it trying to solve ?

GoodbyeNJN:
> The motivation behind this PR is very vague. What real problem is it trying to solve ?

Assume I have two monitors arranged horizontally with the taskbar's "Only from same monitor" option disabled. When windows are open on both monitors, the taskbar icon ordering becomes confusing.

To address this, I've implemented sorting logic that orders taskbar icons based on monitor position: icons from the left monitor appear before those from the right monitor, and icons from the upper monitor appear before those from the lower monitor. This provides a more intuitive user experience.

Additionally, the Niri compositor's workspace `id` property does not reflect workspace order accurately. Instead, the `idx` property should be used for sorting.

However, for multi-monitor setups, Niri's `idx` property is per-display rather than global, meaning workspace indices can repeat across different displays. To ensure consistent and unique sorting when switching from the `id` property to the `idx` property, display arrangement information must be incorporated.

This PR includes these changes.

PR #1065, opened 2025-12-16
Fix VSCode theming

eric-handley:
# Pull Request

## Motivation

Fixes VSCode theme settings not being applied correctly + theme extension not installed in all cases

Also re-installs extension to force VSCode to load the new theme, feels slightly hacky to me but otherwise the new theme can't be previewed without manually reloading VSCode after every change

Previous (`sed` replacement was done incorrectly):
image

Fixed:
image

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring

## Testing

- [ ] Tested on niri
- [x] Tested on Hyprland
- [ ] Tested on sway

N/A:
~~Tested with different bar positions and density settings~~
~~Tested at different interface scaling values~~
~~Tested with multiple monitors (if applicable)~~

## Checklist
- [x] Self-reviewed my code

### DIFF
--- a/Modules/Panels/Settings/Tabs/ColorScheme/ColorSchemeTab.qml
+++ b/Modules/Panels/Settings/Tabs/ColorScheme/ColorSchemeTab.qml
@@ -812,7 +812,7 @@ ColumnLayout {
               var clientName = client.name === "code" ? "VSCode" : "VSCodium";
               clientInfo.push(clientName);
             }
-            return "Detected: " + clientInfo.join(", ");
+            return "Applied to default profile. Detected: " + clientInfo.join(", ");
           }
         }
         Layout.fillWidth: true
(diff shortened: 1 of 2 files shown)


Modules/Panels/Settings/Tabs/ColorScheme/ColorSchemeTab.qml
```diff
@@ -812,7 +812,7 @@ ColumnLayout {
               var clientName = client.name === "code" ? "VSCode" : "VSCodium";
               clientInfo.push(clientName);
             }
-            return "Detected: " + clientInfo.join(", ");
+            return "Applied to default profile. Detected: " + clientInfo.join(", ");
```
eric-handley, line 815: Added this because if you're enabling for the first time and aren't on the default profile, it looks like the theme isn't working

Ly-sec:
Since @tuibird made it, maybe she can say what she thinks about it :).

tuibird:
Looks good to me! It’s definitely an imperfect solution overall, but outside of publishing a theme to the vscode store officially I couldn’t think of a better approach. If you have any ideas on a more elegant solution don’t be afraid to speak up! 🙂

Ly-sec:
I'll merge it then :) thank you :heart: !

PR #1077, opened 2025-12-18
Wifi and Bluetooth panel enhancement

DanniSec:
# Pull Request

## Motivation
Improved the information provided and clarity of the Wifi and Bluetooth panels, by adding info button that displays additional useful information about selected connection for both. 

Tweaked the Bluetooth connecting system to be more clear instead of connect/disconnect clearly state Pair, Connect, Disconnect, Remove.

There was an issue with duplicated devices in Bluetooth connection, prevented that by filtering for same address, toggle button for bluetooth advertising/discoverable mode.

Added new translations for the new strings.

Additional feedback (toasts) for both panels. 

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring

## Related Issue
N/A

## Testing

- [ ] Tested on niri
- [x] Tested on Hyprland
- [ ] Tested on sway

## Screenshots / Videos
image
image

## Checklist
- [x] Self-reviewed my code

## Additional Notes
Could add settings/config to configure in UI the displayed info etc. or whole customization for the Network/Bluetooth panels if wanted by the community. But for now this small improvement makes the panels more usable. 

There is more stuff to see outside the two screenshots, so try it out.

### DIFF
--- a/Modules/Panels/WiFi/WiFiNetworksList.qml
+++ b/Modules/Panels/WiFi/WiFiNetworksList.qml
@@ -12,6 +12,10 @@ NBox {
 
   property string label: ""
   property var model: []
+  // While a password prompt is open, we freeze the displayed model to avoid
+  // frequent scan updates recreating items and clearing the TextInput.
+  property var cachedModel: []
+  readonly property var displayModel: (passwordSsid && passwordSsid.length > 0) ? cachedModel : model
   property string passwordSsid: ""
   property string expandedSsid: ""
   // Currently expanded info panel for a connected SSID
@@ -28,6 +32,22 @@ NBox {
   signal forgetRequested(string ssid)
   signal forgetConfirmed(string ssid)
   signal forgetCancelled
+
+  onPasswordSsidChanged: {
+    if (passwordSsid && passwordSsid.length > 0) {
+      // Freeze current list ordering/content while entering password
+      try {
+        // Deep copy to decouple from live updates
+        cachedModel = JSON.parse(JSON.stringify(model));
+      } catch (e) {
+        // Fallback to shallow copy
+        cachedModel = model.slice ? model.slice() : model;
+      }
+    } else {
+      // Clear freeze when password box is closed
+      cachedModel = [];
+    }
+  }
 
   Layout.fillWidth: true
   Layout.preferredHeight: column.implicitHeight + Style.marginM * 2
@@ -57,7 +77,7 @@ NBox {
     }
 
     Repeater {
-      model: root.model
+      model: root.displayModel
 
       Rectangle {
         id: networkItem
@@ -524,7 +544,7 @@ NBox {
                   focus: visible
                   passwordCharacter: "●"
                   onVisibleChanged: if (visible) {
-                                      text = "";
+                                      // Keep any text already typed; only focus
                                       forceActiveFocus();
                                     }
                   onAccepted: {


ItsLemmy:
Thanks for the PR!

- why the change of code style everywhere ?  why replace all the short-hand syntax by longer versions that do the exact same thing?
This make the PR harder to be reviewed as most of the changes are purely syntactic.
Try to keep the changes to what is really essential.

- The (i) button on the right of the connect/button in the  Bluetooth screenshot should be on the left. Lets try to keep both panels as similar as possible

- The panels' UI could probably be a little bit more compact, maybe in a grid ?   

BT: 
Signal | Battery
Paired | Trusted
Address

WiFi:
Security | Internet Connected
Link Speed | IPV4
Gateway | DNS

Looking forward to it!

DanniSec:
Hi, 
it's a first for me working with QML so it's a bit messy, so I'll try to clean it up.

The grind layout seems good, haven't thought of that.

DanniSec:
@ItsLemmy hardcoding the layout feels restrictive, I think it would be best if it would be configurable in settings, the layout etc. but I'm not sure where to put it, right now I added it to Settings>Network as it has toggles for both and a lot of empty space. But doesn't really fit the purpose of the settings tab, I guess?

image

where do you think it would be best to put the settings for those panels?

ItsLemmy:
Seems way overkill to me. People don't spend their life ricing their SSID list. 
IMO, giving control on ellipsis to the user means we failed at the job of designing a proper interface :P

I would just come up with good sane default. and If REALLY needed then add toggle directly on the panel to switch between two layout modes. (A bit like the launcher does gridview vs listview)

DanniSec:
I get what you mean, probably just a toggle between list and grid is enough, as using just grid might look cleaner in some cases for me it doesn't really work as I have 3 different DNS setup and in gridview I would only see one of them, thus listview being the cleaner option. 

To make it compact yet understandable I think it could be just the icons with alt text on hover (so it has a hint of what it represents) to give more space for the actual information.

So it would look like this: 
image

with the red rectangle being the layout toggle. (Can't screenshot the on hover text but hovering over the icons shows the text descriptions that were on the previous screenshot). 

And similar stuff for BT panel. 

If you think this would align more with the direction of Noctalia then I'll do it like that.

ItsLemmy:
The more I think of it, the more I find this button (red square) a bit misleading. It's location implies it would changet the SSID list to a grid view. But in fact it's just the extra information layout.   So I guess it should be inside it.

image

I also believe if you click the (i) it should collapse all others extra information so the panel does not get crazy tall.

DanniSec:
So you mean by collapse alll others is that if you press (i) the other available networks will be hidden until you toggle off the info? 

I do believe that will make it more compact and cleaner even. 

And for the button location that does make more sense, I'm not really a frontend guy so thank you for your feedback!

turannul:
I noticed one thing here the applet don't have forget icon... Which i would like to see it go. 
Can i suggest showing networks/bluetooth devices on settings within list (with forget icon).

DanniSec:
@turannul I added that already, as that was one of the main reasons of why I started with this. 
Soon I'll send out a new commit with the changes made that Lemmy requested and hopefully will be good for merge!

turannul:
I'll look forward for this PR to get merged.

DanniSec:
Hey, should be good now, only thing I noticed is a weird bluetooth agent behavior with some devices that require confirmation for pairing, though I can't reliably test it. 

Would be nice if someone who has a pin/passkey enforced BT device could have a look a it. 
Otherwise with the current agent for yes/no pairing confirmations it should pair.

ItsLemmy:
Thanks, I'll give it a test run

PR #1094, opened 2025-12-19
Apply new notification drop shadow to toasts

notiant:
This applies the recent change for the notification drop shadow to toasts as well. I also slightly reduced the notification vertical margins to make it look more like before the change.


ItsLemmy:
Thanks!

ItsLemmy:
Had to revert your PR, as it was causing the content of the toast to overflow out of the box, and I had no time to fix it.

image

image

PR #1102, opened 2025-12-20
Notification: reduce left/right margin

notiant:
The recent drop shadow change increased the margins, so this seems necessary.
I also removed some unused entries for the desktop widgets in the i18n.


Ly-sec:
Thank you very much :heart: !

PR #1115, opened 2025-12-22
Improve brightness control on external displays

eric-handley:
# Pull Request

## Motivation
The issue that I've encountered with the current implementation is that each `ddcutil` command takes ~1s to execute. This means if you adjust the brightness rapidly, commands queue up faster than the monitor can respond and it causes really glitchy behavior where the brightness jumps up and down between a bunch of different values. Also, the current usage doesn't feel very responsive as the monitor brightness takes quite a while to change and settle on the final value.

This PR adds brightness adjustment commands to a queue to prevent overlapping executions and should fix this problem. The brightness panel UI also now updates immediately and I've hidden the OSD for brightness changes when the panel is open.

I only have the one display/computer to test on but I'm pretty sure this shouldn't break anything for laptop/Apple displays.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring

## Testing
- [ ] Tested on niri
- [x] Tested on Hyprland
- [ ] Tested on sway

## Checklist
- [x] Self-reviewed my code

### DIFF
--- a/Modules/OSD/OSD.qml
+++ b/Modules/OSD/OSD.qml
@@ -190,6 +190,11 @@ Variants {
 
     function onBrightnessChanged(newBrightness) {
       root.currentBrightness = newBrightness;
+      // Don't show OSD if brightness panel is open
+      var brightnessPanel = PanelService.getPanel("brightnessPanel", root.modelData);
+      if (brightnessPanel && brightnessPanel.isPanelOpen) {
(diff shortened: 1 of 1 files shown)


Ly-sec:
Hey @eric-handley this was long overdue! Thank you very much :heart: !

PR #1120, opened 2025-12-22
NotificationsHistory: replace hide badge when zero with hide icon when zero

damian-ds7:
## Motivation
I feel like option to hide hide unread badge when zero doesn't really make sense because why would anyone want the badge to persist when no unread notifs remain. This PR changes that into hide icon when no unread instead

## Type of Change
- [ ] Bug fix
- [x] New feature
- [x] Breaking change (I think, people with hideWhenZero set to true will have this enabled by default)
- [ ] Refactoring

## Testing

- [x] Tested on niri
- [ ] Tested on Hyprland
- [ ] Tested on sway

## Screenshots / Videos
preview

## Checklist
- [x] Self-reviewed my code

## Additional Notes
If you feel like the previous hideWhenZero option should remain I can revert my current chages and add a new option to hide notification icon instead

### DIFF
--- a/Services/UI/BarWidgetRegistry.qml
+++ b/Services/UI/BarWidgetRegistry.qml
@@ -179,7 +179,7 @@ Singleton {
                                   "NotificationHistory": {
                                     "allowUserSettings": true,
                                     "showUnreadBadge": true,
-                                    "hideWhenZero": true
+                                    "hideWhenZero": false
                                   },
                                   "SessionMenu": {
                                     "allowUserSettings": true,


Ly-sec:
Hey @damian-ds7 I don't really see a reason to hide the notification history when there is no unread notifications. It's a notification history for a reason and not just an inidicator :).

damian-ds7:
@Ly-sec Fair, in my case it acts as an indicator tbh, I don't need to see the icon all the time and it's more clear to me when there's a new one than if I'd use just the indicator, but I can understand if it's not the same for you

Ly-sec:
@damian-ds7 after some more thinking you're right. I'll merge this PR.
Thank you very much :heart: !

PR #1155, opened 2025-12-25
Workspace: readd opacity property for appicons

GravityShark:
# Pull Request

## Motivation
Readd opacity property for appicons in the grouped workspace that used to be in TaskbarGrouped

Makes it a lot easier to distinguish between unfocused appicons

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring

## Testing
I looked at it

- [x] Tested on niri
- [ ] Tested on Hyprland
- [ ] Tested on sway

## Screenshots / Videos
TaskbarGrouped (v3.6.2):
Screenshot from 2025-12-26 01-26-23
Before:
Screenshot from 2025-12-26 01-24-56
After:
Screenshot from 2025-12-26 01-21-00

## Checklist
- [x] Self-reviewed my code

## Additional Notes
This is similar in vain to my other PR, apologies if fragmenting it into 2 different PRs is not correct.

I don't know why this was removed when TaskbarGrouped is now consolidated in the Workspace widget, so please let me know, also i didn't really thoroughly look at all issues or pull requests so apologies if this is a duplicate.

### DIFF
--- a/Modules/Bar/Widgets/Workspace.qml
+++ b/Modules/Bar/Widgets/Workspace.qml
@@ -54,7 +54,7 @@ Item {
   readonly property bool showApplications: (widgetSettings.showApplications !== undefined) ? widgetSettings.showApplications : widgetMetadata.showApplications
   readonly property bool showLabelsOnlyWhenOccupied: (widgetSettings.showLabelsOnlyWhenOccupied !== undefined) ? widgetSettings.showLabelsOnlyWhenOccupied : widgetMetadata.showLabelsOnlyWhenOccupied
   readonly property bool colorizeIcons: (widgetSettings.colorizeIcons !== undefined) ? widgetSettings.colorizeIcons : widgetMetadata.colorizeIcons
-  readonly property real unfocusedIconsOpacity: (widgetSettings.unfocusedIconsOpacity !== undefined) ? widgetSettings.unfocusedIconsOpacity / 100 : widgetMetadata.unfocusedIconsOpacity / 100
+  readonly property real unfocusedIconsOpacity: (widgetSettings.unfocusedIconsOpacity !== undefined) ? widgetSettings.unfocusedIconsOpacity : widgetMetadata.unfocusedIconsOpacity
   readonly property bool enableScrollWheel: (widgetSettings.enableScrollWheel !== undefined) ? widgetSettings.enableScrollWheel : widgetMetadata.enableScrollWheel
 
   readonly property int itemSize: Math.round(Style.capsuleHeight * 0.8)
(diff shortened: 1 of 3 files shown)


ItsLemmy:
This was removed on purpose because the semi transparent look pretty bad when the bar is also semi transparent.
The little active indicator below the icon usually do a good job at identifying the focused app.  But you seem to use a very monochromatic colorscheme...
I would be ok to bring back the 0.6 opacity for the unfocused app if this was a widget options with a default to false.

GravityShark:
Thank you for your comment. I'm going to see if I can add an widget options. if I couldn't I would close this PR.

EDIT: or maybe i should close this PR for now?

EDIT: actually I will put it into draft

dinhokusanagi:
It would be nice if the border was optional; in my case, I like a clean look. It could follow the design of the other widgets.

GravityShark:
@ItsLemmy  I just woke up and I've added a new option. This is my first time writing QML so I am not sure if I've done things correctly/idiomatically.

image

Mainly I'm most unsure here: 
1. I wanted to use NSpinBox as a percentage, so i made it take 0 to 100, and then divide it by 100 later.
2. I just replaced the 0.6 with the value directly, by default it would be 100% opacity

I also just want to thank you for maintaining this project, I really love it!

GravityShark:
> It would be nice if the border was optional; in my case, I like a clean look. It could follow the design of the other widgets.

This doesn't seem to be the correct place for this, because my PR only is about app icon opacity, and not about borders.

I do like the idea, but you should probably create a new issue instead.

ItsLemmy:
We usually do it the other way around : Opacity values are stored from 0.0 to 1.0  and displayed as a percentage by * 100.
Also I'd rather have a slider than a spinbox. You can look for NValueSlider in the codebase there are many example.
```
    // Panel Background Opacity
    NValueSlider {
      Layout.fillWidth: true
      label: I18n.tr("settings.user-interface.panel-background-opacity.label")
      description: I18n.tr("settings.user-interface.panel-background-opacity.description")
      from: 0.4
      to: 1
      stepSize: 0.01
      value: Settings.data.ui.panelBackgroundOpacity
      isSettings: true
      defaultValue: Settings.getDefaultValue("ui.panelBackgroundOpacity")
      onMoved: value => Settings.data.ui.panelBackgroundOpacity = value
      text: Math.floor(Settings.data.ui.panelBackgroundOpacity * 100) + "%"
    }
```

Thanks for the PR, almost there!

I will add an option to no show the borders once you are done, stay focused on the goal :P

GravityShark:
@ItsLemmy I wanted to try something different and make the NValueSlider be on a RowLayout to make it feel consistent with the rest of the options, because if you just use NValueSlider it appears at the bottom.

Screenshot from 2025-12-27 16-26-12
Screenshot from 2025-12-27 16-26-29

I'm not sure what kind of spacing I should have used for RowLayout. but from the other examples they seem to just use Style.marginL.

Also I'm just wondering, is it for clarity that you would always put from, to and step size even if the default is already 0, 1 and 0.01?

Other than that it should be good! Thank you again for maintaining this project.

PR #1158, opened 2025-12-26
Some bluetooth improvements

notiant:
1. This PR adds waiting time after `pair` and `trust` in `pairWithBluetoothctl()`. This is necessary for connecting the device after pairing to work, at least in testing with my Bluetooth speakers.
2. I've added a "no devices available" screen similar to the wifi panel so the panel isn't completly empty in case there are no bluetooth devices.
3. I removed the right-click-to-forget function on the disconnect button since this is not documented by a tooltip and there is now an extra button for it.


ItsLemmy:
Thanks, will see how it behaves.

PR #1149, opened 2025-12-25
Refine Wi-Fi applet

turannul:
- Parse frequency from network link output in NetworkService.
- Implement logic to determine Wi-Fi band (2.4 GHz, 5 GHz, 6 GHz) based on frequency ranges.
- Update WiFiNetworksList to display the current band information instead of generic internet status text.
- Improve connectivity feedback: signal icons and badges now reflect actual internet availability (changing to error color if limited/captive portal) rather than just showing connected.

# Pull Request

## Motivation
Provide a clear and concise explanation of what this PR does and why it is needed.

This PR improves the Wi-Fi applet to accurately reflect different network connectivity states reported by NetworkManager.

## Type of Change
Mark the relevant option with an "x".
- [ ] Bug fix
- [X] New feature
- [ ] Breaking change
- [X] Refactoring (not much)

## Testing
quickshell -vv -p shell.qml > "debug_$(date +%Y-%m-%d_%H-%M-%S).log" is running on background, i see my changes after reload delay. 

- [X] Tested on niri
- [ ] Tested on Hyprland
- [ ] Tested on sway
- [ ] Tested with different bar positions and density settings
- [X] Tested at different interface scaling values
- [X] Tested with multiple monitors (if applicable)

## Screenshots / Videos
Screenshot from 2025-12-25 08-44-41
Screenshot from 2025-12-25 08-49-23

## Checklist
- [X] Code follows project style guidelines
- [X] No new warnings or errors
- [X] Self-reviewed my code
- [ ] Documentation or comments updated (if relevant)

## Additional Notes
Is there any suggestions?

PR #1167, opened 2025-12-27
Use existing pingCheckProcess while in full & limited states.

turannul:
# Pull Request

## Motivation
I didn't noticed pingCheckProcess is unused, i found out today (my DNS Server crashed) | meantime I noticed doesn't work it didn't showed like i intended to.
* Completes #1149 
(PS: My bad i should've noticed this before)

## Type of Change
- [ ] Bug fix 
- [ ] New feature
- [ ] Breaking change
- [X] Refactoring

## Testing
quickshell -vv -p shell.qml > "debug_$(date +%Y-%m-%d_%H-%M-%S).log" is running on background, i see my changes after reload delay.

- [X] Tested on niri
- [ ] Tested on Hyprland
- [ ] Tested on sway

## Checklist
- [X] Self-reviewed my code

## Additional Notes
connectivityCheckTimer 15s I didn't changed but might be too short for connectivity check.
Also may cause too frequent pinging/spam behavior, might need somewhat smarter connectivity check logic...


turannul:
I need to know what i can do and what i can't do.
 - I personally hate polling like approach, pinging on regular basis is waste of resources, constantly spawning shell.
The idea of fixing this connecting (and stay connected while in full state) to a server which is reliable eg (google.com) and ignore timeouts.
Which i could have implemented as a separate python script and execute within NetworkService.qml.
I'm open to suggestions other ideas.

PR #1174, opened 2025-12-28
Fix: password input not possible after leaving suspend w/ dual GPUs 

### LINKED ISSUE #917: Lock screen doesn't allow text input when waking from suspend in Hyprland
### Description
When in Hyprland (I cannot reproduce this in Niri), if I put my laptop to sleep and wake it up, the text input for typing in my password in the lock screen loses focus and cannot regain focus. All of the other buttons still work, so it is still possible to e.g. log out and log back in again.

### Steps to Reproduce
1. Suspend computer (either by closing laptop lid or manually suspending)
2. Resume computer activity (and be greeted by the noctalia lock screen)
3. Try to type password into prompt

### Expected Behavior
The password prompt should remain in focus and allow me to type my password

### Screenshots
I cannot access my screenshot utility at the lock screen, but it largely looks identical, except the text input that usually has a colored border around it now has a greyed out border and no blinking typing prompt.

### Environment
- Distro: NixOS
- Compositor: Hyprland
- Noctalia-shell Version: 3.4.0 (tested just after fetching main on rev 078195f54b5ede5d41fa0fea850f6fe220bfc36f)

### LINKED ISSUE #1029: Unable to log in after the screen is locked
### Description
After locking the screen (Session Menu → Lock), if you click the "Log Out" button at the bottom, you will not be able to log in again, because the login menu does not allow you to enter a password.

### Steps to Reproduce
1. Go to Control Center
2. Click on Session Menu
3. Click on Lock
4. Click on LogOut
5. After this, if you try to log in, the password entry field will simply be inactive, making it impossible to type the password and log in.

### Expected Behavior
I would like to see either the standard password input field, or, for example, that when hovering over or clicking on the password entry field, it becomes active so that the password can be entered.

### Environment
- Distro:  CachyOS
- Compositor: Hyprland
- Noctalia-shell Version: 3.6.0

eric-handley:
# Pull Request

## Motivation
Fixes password input losing focus on lock screen after suspend/resume in Hyprland, which seems to only affect dual GPU setups. 
- Added MouseArea with hover detection that automatically restores focus when cursor moves. 
- Based on end4's approach but simplified to use only the [mouse movement workaround](https://github.com/end-4/dots-hyprland/blob/80a7804adedf6f7bf6e3f14641d36461a379ba7b/dots/.config/quickshell/ii/modules/ii/lock/LockSurface.qml#L29-L36) (could make more robust by adding `hypridle` as a package dependancy and following [end4's method](https://github.com/end-4/dots-hyprland/blob/80a7804adedf6f7bf6e3f14641d36461a379ba7b/dots/.config/hypr/hypridle.conf#L8) more closely)

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring

## Related Issue
- Closes #917
- Closes #1029

## Testing
Tested on a laptop w/ dual GPU that previously had this issue

- [ ] Tested on niri
- [x] Tested on Hyprland
- [ ] Tested on sway

## Screenshots / Videos
N/A

## Checklist
- [x] Self-reviewed my code

### DIFF
--- a/Modules/LockScreen/LockScreen.qml
+++ b/Modules/LockScreen/LockScreen.qml
@@ -377,6 +377,18 @@ Loader {
           Item {
             anchors.fill: parent
 
+            // Mouse area to trigger focus on cursor movement (workaround for Hyprland focus issues)
+            MouseArea {
+              anchors.fill: parent
+              hoverEnabled: true
+              acceptedButtons: Qt.NoButton
+              onPositionChanged: {
+                if (passwordInput) {
+                  passwordInput.forceActiveFocus();
+                }
+              }
+            }
+
             // Time, Date, and User Profile Container
             Rectangle {
               width: Math.max(500, contentRow.implicitWidth + 32)


Ly-sec:
Hey there @eric-handley , thank you very much for this PR. We've been trying to find a fix for it for a while but since neither me nor Lemmy could replicate it, it was rather tough :p. About the hypridle part, I'd rather not add it as a dependency given with the next release of quickshell, we're getting access to proper idle management ourselves :).
Going to merge this. Thank you very much :heart: !

PR #1177, opened 2025-12-28
feat(hooks): add shutdown and reboot hooks

sQVe:
Add user-configurable hooks that run before system shutdown/reboot, allowing cleanup tasks like syncing filesystems or stopping services.

#### Changes

- Added `executeShutdownHook` and `executeRebootHook` functions with blocking Process infrastructure
- Integrated hooks into CompositorService shutdown/reboot flow via callbacks
- Added Settings UI inputs for shutdown/reboot hook commands in HooksTab
- Added default settings and translations for new hooks

#### Test plan

- [ ] Configure a shutdown hook (e.g., `notify-send "Shutting down"`)
- [ ] Trigger shutdown and verify hook executes before poweroff
- [ ] Configure a reboot hook and verify it executes before reboot
- [ ] Test with no hooks configured - shutdown/reboot should proceed immediately
- [ ] Test "Test" button in Settings UI executes the command


ItsLemmy:
I feel like executeShutdownHook and executeRebootHook could be factorized into one, maybe "executeSessionHook"  so this could be reused if we need to add more later (suspend, etc..)

sQVe:
@ItsLemmy Sure thing. So basically unify the the hooks and share them between Shutdown and Reboot, right?

ItsLemmy:
Yes, please!

sQVe:
@ItsLemmy Should be good to go now.

ItsLemmy:
thanks

PR #1197, opened 2025-12-30
feat(Launcher): ability to have calculator in inline search

DuckySoLucky:
# Pull Request

## Motivation

https://canary.discord.com/channels/1401598189823590460/1401607843123036162/1455183045299535967
> Is there a specific reason the `> calc` functionality exists? I went thru the code and I can see that you already do regex parsing to get rid of all characters, so letting users do math inside of normal search would make sense to me (at least for me, cuz I always forget to run `> calc` :kek~1:, and most of the time it's just to do some simple multiplication or smth like that), so user could type smth like `17 * 6` inside of search and it would just tell the result below, and in case there was an app with that name it would be shown at the top, and math result below. I don't really see a point why math couldn't be supported in normal search without `> calc`, originally I thought u made custom parser or smth like that, but u just `eval()` the input. I could make a PR where u can toggle it, I would greatly appreciate if toggle like that existed (I'm not entirely sure if you agree with this, so that's why I'm asking first). Thanks :pepeLove:

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring

## Related Issue
- Closes https://canary.discord.com/channels/1401598189823590460/1401607843123036162/1455183045299535967

## Testing

- [ ] Tested on niri
- [x] Tested on Hyprland
- [ ] Tested on sway

## Screenshots / Videos

https://github.com/user-attachments/assets/07e2feca-6270-4989-b271-969a81898fa8

## Checklist
- [x] Self-reviewed my code

## Additional Notes

Removed `evaluateExpression()` cuz it's not being used anywhere, if u want I can add it back :man_shrugging: 

Also I'm not entirely sure if setting property is needed, but added it just in case if someone doesn't like it

### DIFF
--- a/Modules/Panels/Launcher/Plugins/CalculatorPlugin.qml
+++ b/Modules/Panels/Launcher/Plugins/CalculatorPlugin.qml
@@ -10,7 +10,8 @@ Item {
   property string iconMode: Settings.data.appLauncher.iconMode
 
   function handleCommand(query) {
-    return query.startsWith(">calc");
+    // Handle >calc command or direct math expressions after >
+    return query.startsWith(">calc") || (query.startsWith(">") && query.length > 1 && isMathExpression(query.substring(1)));
   }
 
   function commands() {
@@ -29,7 +30,15 @@ Item {
   }
 
   function getResults(query) {
-    let expression = query.substring(5).trim();
+    let expression = "";
+
+    if (query.startsWith(">calc")) {
+      expression = query.substring(5).trim();
+    } else if (query.startsWith(">")) {
+      expression = query.substring(1).trim();
+    } else {
+      return [];
+    }
 
     if (!expression) {
       return [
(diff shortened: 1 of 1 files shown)


Ly-sec:
Hey @DuckySoLucky I was just talking to Lemmy about it and we think the best approach would be not even having a setting for it. Basically getting rid of the calculator plugin and have it always inline instead. That's how most launchers do it anyways :). Would you mind doing that instead :heart: ?

DuckySoLucky:
> Hey @DuckySoLucky I was just talking to Lemmy about it and we think the best approach would be not even having a setting for it. Basically getting rid of the calculator plugin and have it always inline instead. That's how most launchers do it anyways :). Would you mind doing that instead ❤️ ?

That was my original idea, however, I wanted to avoid breaking backward compatibility and potentially confusing existing users. It is is now enabled by default, and I have added a deprecation notice to the `> calc` command. I am not entirely sure how you typically handle feature changes, whether features are removed immediately or deprecated first, but in either case, if the deprecation step is unnecessary  you can just remove the plugin and everything should continue to work as expected

https://github.com/user-attachments/assets/4fe5eb03-f15a-4142-8161-3099a4686e50

Ly-sec:
This is a very good way of doing it in my opinion :heart: ! Thank you very much

PR #1206, opened 2025-12-31
Add function to IPCService to query current visibility of bar

bean-factory:
# Pull Request

## Motivation
This PR adds a function to IPCService which allows users to query the current visibility of the bar. This can be used in external scripts to specifically hide or show the bar instead of just a toggle or to do different behaviours depending on the current visibility of the bar.

## Type of Change
- [ ] Bug fix
- [X] New feature
- [ ] Breaking change
- [ ] Refactoring

## Testing

- [X] Tested on niri
- [ ] Tested on Hyprland
- [ ] Tested on sway

## Checklist
- [X] Self-reviewed my code

## Additional Notes
IPC docs need to be updated

### DIFF
--- a/Services/Control/IPCService.qml
+++ b/Services/Control/IPCService.qml
@@ -27,6 +27,9 @@ Item {
     function toggle() {
       BarService.isVisible = !BarService.isVisible;
     }
+    function isVisible(): bool {
+      return BarService.isVisible;
+    }
   }
 
   IpcHandler {


ItsLemmy:
We already exposed all settings and state variables in the IPC "state all" call
You can retrieve the bar visibility using the following snippet :

``` qs -c noctalia-shell ipc call state all | jq '.state.barVisible' ```

PR #1218, opened 2026-01-01
System Monitor: Show load average metrics

shouya:
## Motivation

Load average is usually a better indicator of system responsiveness than CPU utilization percentage. Since it shows how many processes are waiting for CPU time, which directly correlates with scheduling delays and user-perceived latency.

This PR adds load average display to the system monitor widget. On the bar widget it shows the 1-minute load average, and all three values (1, 5, 15 min) in the tooltip.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring

## Testing

- [x] Tested on niri
- [ ] Tested on Hyprland
- [ ] Tested on sway

Widget styling should be consistent with other system monitor icons.

## Screenshots / Videos

image

## Checklist
- [x] Self-reviewed my code

## Additional Notes

I'd like to check if this feature is desired for the main codebase. If not, I'm happy to implement it as a plugin instead. My preference would be keeping it here since it maintains visual consistency with the other system monitor widgets.

### DIFF
--- a/Modules/Panels/SystemStats/SystemStatsPanel.qml
+++ b/Modules/Panels/SystemStats/SystemStatsPanel.qml
@@ -230,6 +230,33 @@ SmartPanel {
                 }
               }
 
+              // Load average
+              RowLayout {
+                Layout.fillWidth: true
+                spacing: Style.marginS
+                visible: SystemStatService.nproc > 0
+
+                NIcon {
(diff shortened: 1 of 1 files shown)


ItsLemmy:
Thanks!

PR #1250, opened 2026-01-03
feat: Add Wallhaven API key management via environment variable

c0mpile:
## Motivation
This allows for the Wallhaven API key to be set as an environment variable by the compositor, avoiding storage of it in the settings.json file. If someone wants to include their settings.json in a dotfiles git repository, it's much easier to do it this way to avoid publishing the API key. 

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring

## Testing
I tested by removing the key from my settings and then defining an environment variable in a secrets file that was sourced by Hyprland. I tested all purity settings and they worked as expected.

- [ ] Tested on niri
- [x] Tested on Hyprland
- [ ] Tested on sway

## Screenshots / Videos
image

## Checklist
- [x] Self-reviewed my code

### DIFF
--- a/Assets/Translations/en.json
+++ b/Assets/Translations/en.json
@@ -2934,8 +2934,10 @@
     "no-wallpaper": "No wallpaper found.",
     "panel": {
       "apikey": {
+        "env-active": "API key is managed by the NOCTALIA_WALLHAVEN_API_KEY environment variable.",
         "help": "An API key is required to access NSFW content.",
         "label": "API Key",
+        "managed-by-env": "Managed by environment variable",
         "placeholder": "Enter your Wallhaven API Key"
       },
       "apply-all-monitors": {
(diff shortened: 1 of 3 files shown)


ItsLemmy:
Cool! but it's a bit repetitive.

I would put the full warning "API key is managed by..."  straight inside the NTextInput placeholder.  You can probably make it a tad shorter too : "Managed via NOCTALIA_WALLHAVEN_API_KEY environment variable."  so it fits.

c0mpile:
> Cool! but it's a bit repetitive.
> 
> I would put the full warning "API key is managed by..." straight inside the NTextInput placeholder. You can probably make it a tad shorter too : "Managed via NOCTALIA_WALLHAVEN_API_KEY environment variable." so it fits.

Noted and fixed.

PR #1296, opened 2026-01-07
fix(plugins): support compositeKey for custom plugins in settings popup

anthonyhab:
Fixes the issue where custom plugins (installed from non-official repositories) could not have their settings opened through the Noctalia settings UI.

## Problem
Custom plugins are loaded with composite keys (e.g., `c4ed72:polkit-auth`) but the settings popup was trying to access them using their manifest ID (`polkit-auth`), causing the error: "Cannot open settings: plugin not loaded"

## Solution
Modified `NPluginSettingsPopup.qml` to use `compositeKey || manifest.id` when accessing plugins, ensuring it works for both custom plugins (with composite keys) and official plugins (with plain IDs).

## Changes
- Modified `Widgets/NPluginSettingsPopup.qml` line 114
- Changed from: `var pluginId = pluginManifest.id;`
- Changed to: `var pluginId = pluginManifest.compositeKey || pluginManifest.id;`

## Impact
- ✅ Fixes settings access for all custom-source plugins
- ✅ Maintains compatibility with official plugins
- ✅ No breaking changes
- ✅ Minimal, targeted fix

## Testing
Tested with polkit-auth plugin installed from custom repository - settings now open correctly.

### DIFF
--- a/Widgets/NPluginSettingsPopup.qml
+++ b/Widgets/NPluginSettingsPopup.qml
@@ -122,15 +122,16 @@ Popup {
       return;
     }
 
+    // Get plugin directory
     var pluginDir = PluginRegistry.getPluginDir(pluginId);
     var settingsPath = pluginDir + "/" + pluginManifest.entryPoints.settings;
     var loadVersion = PluginRegistry.pluginLoadVersions[pluginId] || 0;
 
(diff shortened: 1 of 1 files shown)


ItsLemmy:
Nice catch, thank you!

PR #1299, opened 2026-01-07
Bluetooth & Network enhancements

notiant:
1. This PR adds back the airplane mode detection that was removed by a previous PR. I also changed the BluetoothService back to a Singleton from a QtObject since this change seems unnecessary to me and breaks in style with every other service.
2. I made some changes to the bluetooth and network panel to align their layout. In the network panel, I made the height adaptable to the content to remove empty space and made the new wifi/ethernet tab bar to appear only if there is ethernet available (on my laptop this seems like clutter otherwise).
3. The grid view of the wifi/bluetooth info panel now uses columns of the same width (the bluetooth info panel seemed very unbalanced).


ItsLemmy:
@DanniSec

DanniSec:
Good PR, only thing I would point out is on line 260 and 263 where are the toast.wifi.enabled/disabled even though it's for bluetooth, probably the best solution as these toast are quite generic would be to put them under "common" (common.enabled/disabled) or something. (Though this "issue" would be on me but looking at the PR made me realize this)

And probably move "bluetooth.address-copied" to the "common" as well for clarity. 

This also made me though about rfkill usage, right now ProgramCheckerService doesn’t publish a "rfkillAvailable" property, and BluetoothService calls rfkill unconditionally, we should add rfkillAvailable to ProgramCheckerService.programsToCheck. 

Only run the rfkill process in BluetoothService when available, or fall back gracefully if unavailable.

Even though it might be unnecessary I think this would stick more into the structure of the project and better clarity in case of errors. 


Other than this the PR is solid and great improvement. @ItsLemmy

notiant:
"common.enabled/disabled" exists but is currently not in use, all the toasts use "toast.wifi.enabled/disabled". This should probably be changed.

I also think that references to the blocked state in BluetoothDevicesList could be removed since blocked devices aren't shown anyway, alternatively an option to display blocked devices could be implemented.

DanniSec:
@ItsLemmy You can merge, and I'll fix up the translations and add the rfkill gate in another PR, unless @notiant want's to do it.

PR #1307, opened 2026-01-08
i18n corrections

notiant:
This is mostly for the German translation.

I changed "clock panel" to "calendar panel" since currently both names are in use and the tooltip is "Calendar", so it should be consistently referred to with the same name.


Ly-sec:
Thank you very much @notiant :heart: !

PR #1376, opened 2026-01-13
Add proper cursor icon for system Monitor widget when hovering.

whiskeyPeak:
Adds the correct mouse icon when hovering over the SystemMonitor widget in the Bar. Given that there is an actual press action, I feel like the cursor icon should reflect that. I didn't create an issue because this seemed to just be such a trivial bug.

Tested on Niri.

Video showing change:

https://github.com/user-attachments/assets/2d543275-314c-4910-a484-225916425c59


Ly-sec:
Thank you very much for this PR :heart: !

Issue #1387, opened 2026-01-14
Battery Threshold Control


twilfredo:
### Feature Description

Possibility of bringing back support for battery threshold support for laptops.

### Why Is This Needed?

As per https://docs.noctalia.dev/configuration/laptop-battery/, battery manager that allowed setting charging thresholds was removed due to compatibility issues. AFAIK, currently there is no way in the GUI to set this. However, this would be a really nice feature to have. To make it convenient for laptop users to set charge thresholds on the go.

### Suggested Solutions

As a starting point, without depending on something like TLP, could we add a simple implementation that makes the battery threshold available as an extension to the battery widget or as a plugin based on if sysfs has the relevant nodes? For example: `/sys/class/power_supply/BAT0/charge_control_end_threshold. Perhaps a slider where a user can set it, it is kept track of and applied on started up etc...

Is this feasible?

ItsLemmy:
closing this and looking forward to your plugin.

PR #1395, opened 2026-01-15
feat: Add battery charging threshold control for laptops

SINGH-RAJVEER:
### Motivation
Restores battery charging threshold support for laptops that was previously removed due to compatibility issues. This implementation directly reads/writes sysfs nodes without requiring external dependencies like TLP, allowing users to set charging limits through the GUI to extend battery lifespan.

### Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Breaking change
- [ ] Refactoring

### Implementation Details
Adds battery threshold control via sysfs (`/sys/class/power_supply/BAT*/charge_control_end_threshold`):
- Python script handles threshold detection, read, write, and permission setup
- GUI slider in battery panel (60-100% range) on supported systems
- One-time sudo authentication for udev rule installation
- Persistent storage in `~/.config/noctalia/battery_threshold`
- Auto-restore on shell startup

### Related Issue
- Closes [#(1387)](https://github.com/noctalia-dev/noctalia-shell/issues/1387)

### Testing
- [x] Tested on systems with threshold support
- [x] Tested permission setup flow
- [x] Tested persistence across reboots
- [x] Verified graceful handling on unsupported systems
- [x] Tested on niri
- [ ] Tested on Hyprland
- [ ] Tested on sway

### Checklist
- [x] Self-reviewed my code
- [x] Documentation or comments updated

### Additional Notes
Only appears on systems with sysfs threshold support. No external dependencies. Requires one-time sudo for permission setup. Systems without support are unaffected.


ItsLemmy:
Thanks for the PR, it's very interesting!
Unfortunately at this point in time we don't want to add any more "niche" features in "core".

Can you re-implement this as a plugin? I'm sure people will love it, and we could have it in the official repo.
The only drawback will be that the battery threshold UI won't be part of the battery panel, but I believe it's fine for this kind of stuff which is more or less "set once and forget".

twilfredo:
> Thanks for the PR, it's very interesting! Unfortunately at this point in time we don't want to add any more "niche" features in "core".
> 
> Can you re-implement this as a plugin? I'm sure people will love it, and we could have it in the official repo. The only drawback will be that the battery threshold UI won't be part of the battery panel, but I believe it's fine for this kind of stuff which is more or less "set once and forget".

I played around with a plugin for this: https://github.com/twilfredo/noctalia-battery-threshold/tree/master

It seems to work so far, the downside is having to install udev rules for (writing to the sysfs attributes)... is there a better way?

twilfredo:
> > Thanks for the PR, it's very interesting! Unfortunately at this point in time we don't want to add any more "niche" features in "core".
> > Can you re-implement this as a plugin? I'm sure people will love it, and we could have it in the official repo. The only drawback will be that the battery threshold UI won't be part of the battery panel, but I believe it's fine for this kind of stuff which is more or less "set once and forget".
> 
> I played around with a plugin for this: https://github.com/twilfredo/noctalia-battery-threshold/tree/master
> 
> It seems to work so far, the downside is having to install udev rules for (writing to the sysfs attributes)... is there a better way?

For context:

image

ItsLemmy:
Closing this as it won't be included in core, feel free to join discord if you wanna discuss it more in the plugins channel.
Thank you!

PR #1402, opened 2026-01-15
Add some missing translations

notiant:
1. This PR adds back the calculation for the minimum button row width in LockScreen.qml that was removed in the most recent commit. I'm not quite sure why, since a width of 750px is not enough for some languages.
2. I've added some missing translation for the file picker and the launcher.


ItsLemmy:
The code was removed because it made each button a different width which was reported as bug many times. The computation was not great. I'd rather just make thing wider if that does not fit.   I tested in German before committing and it was still fitting nicely, at least with my font. 

The lockscreen needs a real good cleanup as all margin and spacing are hardcoded everywhere and is now kindda hard to maintain. I wont be merging the text metric back.

Try to keep PR separated as much as possible so we can merge them sooner. Thanks!

notiant:
@ItsLemmy OK, I removed that part again but set the bottomContainer width to 800 if the hibernate button is enabled since there is barely any padding in German otherwise.

notiant:
I removed all changes to the LockScreen.

ItsLemmy:
If the translations are up on the i18n web site :
- don't include the translations in your PR
- just mention in your PR that it needs the latest translations  and we will fetch them after the merge. 

this will avoid any merge conflict. thanks!

notiant:
@ItsLemmy Just so I understand correctly, when opening a PR that adds new translation entries I should add them in parallel to the web platform? What about when a PR would include the change or deletion of existing translations but then gets rejected? Wouldn‘t that be a bit messy to reverse?
Can you still merge the current PR without any conflict?

ItsLemmy:
You are right, this is not ideal. I gotta think more on this.
Right now I merged your, but I had to force push / overwrite the translation on the website that I don't like.

PR #1408, opened 2026-01-15
Fix/less math

MrDowntempo:
# Pull Request

## Motivation
Reduces unneeded math since a lot of style values*2 etc. equal other style values. 

## Type of Change
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [X] Refactoring

## Testing

I'll be honest, not tested thoroughly. I've not seen any issue though. Since widgets were affected, and this was done in bulk, I could've affected something, somewhere theoretically. But the math *should* all work out precisely the same.

- [X] Tested on niri
- [ ] Tested on Hyprland
- [ ] Tested on sway

## Screenshots / Videos

## Checklist
- [X] Self-reviewed my code

## Additional Notes
In some cases a marginM might be * 2 so it accounts for both sides. So it could be argued changing that to marginXL would be less clear, since we're accounting for 2 marginMs. I think the less math the better though, even if these are marginal performance gains.

### DIFF
--- a/Modules/Bar/Extras/BarPillVertical.qml
+++ b/Modules/Bar/Extras/BarPillVertical.qml
@@ -42,8 +42,8 @@ Item {
   readonly property int buttonSize: Style.capsuleHeight
   readonly property int pillHeight: buttonSize
   readonly property int pillOverlap: Math.round(buttonSize * 0.5)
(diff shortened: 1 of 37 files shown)


Ly-sec:
LGTM, going to merge it :)

PR #1409, opened 2026-01-15
Add option to hide notification widget when 0 notifications

notiant:
This PR adds the options to hide the notification widget if there are no notifications. Currently, there is only the option to hide the widget if there are no unread notifications, but I think it's rather annoying that you can't access read notifications in that case anymore.


ItsLemmy:
Translations should be altered/deleted/added from the webapp. thx!

ItsLemmy:
Same thing, translations should not be in PRs anymore. Thanks

PR #1410, opened 2026-01-15
LockScreenPanel: small layout adjustments

notiant:


Ly-sec:
Thank you @notiant :heart: !

PR #1411, opened 2026-01-15
Bluetooth rework

turannul:
# Pull Request

## Motivation
Aiming for improve bluetooth pairing experience. 
The bash script was a great step to begin with, thanks to @DanniSec.
But sending yes whenever script launched wasn't the greatest implementation I wouldn't have paired with my phone most of time.
Since expect binary required to make a proper script is external resource to install (and i know maintainers hates this idea) 
So i moved to python. ## which is already installed for many other things ##

Pin pairing should work with this, and yes/no is slightly improved. 
## Type of Change
- [x] Bug fix
- [X] New feature
- [ ] Breaking change
- [X] Refactoring

## Related Issue
N/A

## Testing
The usual testing setup. This time has a caveat any device with pair button will open PIN UI, which im not the best of that.

- [X] Tested on niri
- [ ] Tested on Hyprland
- [ ] Tested on sway

## Screenshots / Videos
TODO: UI isn't ready yet.

## Checklist
- [X] Self-reviewed my code

## Additional Notes
i18n and UI still needs more love, and fair bit of testing is required.


ItsLemmy:
Just a heads up, we dont plan on having BT depends on python...

turannul:
> Just a heads up, we dont plan on having BT depends on python...

Well qs don't have it, bash script would rely on external package (expect) so that's only idea i had.
In addition this only matter when pairing to devices previously not paired, Connecting to devices that already paired is qs (bluez) native.

DanniSec:
This implementation we have is currently kind off "just works" workaround until qs has better support for BT handling. 

Depending on python imo is worse than the bash script, I agree that it's not the best solution, but I would rather try improving the bash script for the pairing than using python.

Sadly my knowledge of C++ isn't good enough to contribute to qs directly so we can use it instead of these workarounds.

So I would look more in a direction of improving the bash script.

ItsLemmy:
Major refactoring are  always best discussed with project leads on discord before diving head first into coding.
Thanks!

turannul:
> This implementation we have is currently kind off "just works" workaround until qs has better support for BT handling.
> 
> Depending on python imo is worse than the bash script, I agree that it's not the best solution, but I would rather try improving the bash script for the pairing than using python.
> 
> Sadly my knowledge of C++ isn't good enough to contribute to qs directly so we can use it instead of these workarounds.
> 
> So I would look more in a direction of improving the bash script.

Well i tried that as i told before we need [expect(1)](https://linux.die.net/man/1/expect) without it as far as im aware there is no good way to improve bash script.
I tried to make knockoff versions of it simply didn't worked.

Until qs gets ability to do so we need this, one way or another.
Also python is much much more reliable than shell scripts

turannul:
> Major refactoring are always best discussed with project leads on discord before diving head first into coding. Thanks!

I know that, i just wanted to have sort of working concept before purposing the idea, and yet **blindly** denied just like the first time. **This was a draft pr before closed**

Bin/bluetooth-connect.py
```diff
@@ -0,0 +1,216 @@
...
+    paired = False
+
+    log("Waiting for pairing logic...")
+    while time.time() - start_time < pair_wait_seconds:
+        out = read_output(timeout=0.5)
+        if out:
+            print(out, end='')
+            # Numberic Comparison (NC) 1 of 4 - Tested pairing with my iPhone.
+            if "Confirm passkey" in out or "yes/no" in out or "Request confirmation" in out:
+                log("Detected passkey prompt. Sending 'yes'.")
```
turannul, line 121: Is anyone have read code?

turannul:
@ItsLemmy That wasn’t finished

PR #1424, opened 2026-01-17
swapped IPv4 and gateway

DanniSec:
@ItsLemmy swapped the gateway with IPv4 in for wifi and ethernet info panel.


ItsLemmy:
thanks!

PR #1428, opened 2026-01-17
Do not persist wifiEnabled in settings

spacedentist:
# Pull Request

## Motivation
I am using a laptop that is sometimes docked (wifi off), and sometimes not (wifi on). Now when I connect to or disconnect from the docking station, that triggers a change in my noctalia-shell config (persisting the current state of the wifi). This is not desirable, as I'm keeping the noctalia-shell config in a dotfiles repo with other configs, so there is constant noise of this bit of config being changed.

I also think the current situation of persisting `wifiEnabled` in the settings is inconsistent with similar other bits system state. Whether Bluetooth is on or off, which power profile is activated etc. Those are all part of the state of my system, which noctalia-shell can display and also mutate, but doesn't persist in its own settings to disk. I think having noctalia-shell not persist those (including `wifiEnabled`) is the correct choice.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [x] Breaking change
- [ ] Refactoring

## Related Issue

## Testing

- [x] Tested on niri
- [ ] Tested on Hyprland
- [ ] Tested on sway

## Checklist
- [x] Self-reviewed my code

## Additional Notes
This is my first attempt of contributing to noctalia-shell, so I might not be aware of some aspects of the coding style. I have read the guideline for contributing. Sorry about that - happy to make changes to comply. I guess this PR is also a request for comments, as to whether or not the main developers agree with the take that `wifiEnabled` shouldn't be persisted.

### DIFF
--- a/Commons/Settings.qml
+++ b/Commons/Settings.qml
@@ -507,7 +507,6 @@ Singleton {
 
     // network
     property JsonObject network: JsonObject {
-      property bool wifiEnabled: true
       property bool bluetoothRssiPollingEnabled: false  // Opt-in Bluetooth RSSI polling (uses bluetoothctl)
       property int bluetoothRssiPollIntervalMs: 10000 // Polling interval in milliseconds for RSSI queries
       property string wifiDetailsViewMode: "grid"   // "grid" or "list"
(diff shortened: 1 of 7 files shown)


notiant:
This currently sends a "WiFi enabled/disabled" toast on every startup. It might make more sense to wait for the next Quickshell update since in the current git branch there exists a wifiEnabled property in the new Quickshell.Networking module.

ItsLemmy:
Will indeed wait for upstream improvements.

PR #1435, opened 2026-01-17
Fixed inconsistent font for Media OSD

bean-factory:
# Pull Request

## Motivation
Fixed inconsistent font family and size of Media OSD
## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring

## Testing

- [x] Tested on niri
- [ ] Tested on Hyprland
- [ ] Tested on sway

## Checklist
- [x] Self-reviewed my code

## Additional Notes
this was changed in https://github.com/noctalia-dev/noctalia-shell/commit/cd20b706e072042af863446e6ed356154e3e0854, but it makes it visually inconsistent with other OSD types

### DIFF
--- a/Modules/OSD/OSD.qml
+++ b/Modules/OSD/OSD.qml
@@ -785,6 +785,7 @@ Variants {
               text: root.getDisplayPercentage()
               color: Color.mOnSurface
               pointSize: Style.fontSizeXS
+              family: Settings.data.ui.fontFixed
               font.weight: Style.fontWeightRegular
               horizontalAlignment: Text.AlignLeft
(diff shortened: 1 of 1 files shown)


ItsLemmy:
Reverted the icon changes that was merged earlier today. We already have aliases for play and pause icons used across the shell, I don't see why the play button would change to outlined here.

Also we don't want regular text to be shown as monospace, we use monospace only for numbers so they stay aligned.

So we won't be merging that one.

I'm not sure this OSD mediaplayer will stay:  it does not work well  with vertical OSD as there is no space to print the text.  
Maybe Toast would be more appropriate.

PR #1438, opened 2026-01-17
Remove the code that creates `gui-settings.json` for NixOS users

bokicoder:
For one, there's now the "Copy Settings" button. For another, having two settings is confusing for users.


bokicoder:
https://github.com/noctalia-dev/noctalia-docs/pull/64

PR #1455, opened 2026-01-18
feat(audio): implement volume feedback sound

### LINKED ISSUE #1358: Short sound when changing volume
### Feature Description
When changing the volume via volume keys or via the volume panel, a short sound should be played.

### Why Is This Needed?
Sometimes there is a need to adjust the volume before playing the desired media. Like before starting a presentation being connected to external speakers not knowing how many % is loud enough.

### Suggested Solutions
Play standard FreeDesktop sound effect on changing the volume.

### Additional Context
Can look at how KDE Connect or Gnome implement it for an inspiration.

chenkeao:
# Pull Request

## Motivation

implements the feature requested in issue #1358.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring

## Related Issue

- Closes #1358 

## Testing

- [x] Tested on niri
- [ ] Tested on Hyprland
- [ ] Tested on sway

## Screenshots / Videos

https://github.com/user-attachments/assets/77eb9134-3103-4a0a-bcd9-88c79a76192b

## Checklist

- [x] Self-reviewed my code

## Additional Notes

I utilized sounds from the freedesktop theme. Attribution has been added, but please double-check for license compliance.


wxmup:
I set up a macro on the mouse to adjust the volume, but I found that this volume feedback sound does not take effect when adjusting the volume using the mouse. Is there a solution to this?

chenkeao:
> I set up a macro on the mouse to adjust the volume, but I found that this volume feedback sound does not take effect when adjusting the volume using the mouse. Is there a solution to this?

How exactly is your mouse macro configured? If your macro runs a command (like `pactl set-sink-volume`), it won't trigger the sound because the shell can't capture the adjustment, while if your macro simulates a key press (Volume Up/Down keys), it should work. I suggest mapping your mouse buttons to simulate the keyboard's Multimedia Volume Keys. 

For example, I use logiops and my config file is:
```
     │ File: /etc/logid.cfg
─────┼────────────────────────────────────────────────────────────────────────
   1 │ devices: (
   2 │ {
   3 │     name: "M720 Triathlon Multi-Device Mouse";
   4 │     buttons: (
   5 │         {
   6 │             cid: 0x56;
   7 │             action =
   8 │             {
   9 │                 type: "Keypress";
  10 │                 keys: ["KEY_VOLUMEUP"];
  11 │             };
  12 │         },
  13 │         {
  14 │             cid: 0x53;
  15 │             action =
  16 │             {
  17 │                 type: "Keypress";
  18 │                 keys: ["KEY_VOLUMEDOWN"];
  19 │             };
......
```
and this can trigger the feedback sound.

PR #1471, opened 2026-01-20
BatteryWidget: charging status only if UPowerDeviceState.Charging

notiant:
The battery widget shows that it's charging when my laptop is connected to a charger but in idle state. `charging` should only be true in the case of UPowerDeviceState.Charging.

### DIFF
--- a/Modules/Bar/Widgets/Battery.qml
+++ b/Modules/Bar/Widgets/Battery.qml
@@ -134,11 +134,11 @@ Item {
   function chargingStatus(state) {
     switch (state) {
     case UPowerDeviceState.Charging: // 1
-    case UPowerDeviceState.FullyCharged: // 4
-    case UPowerDeviceState.PendingCharge: // 5
       return true;
     case UPowerDeviceState.Discharging: // 2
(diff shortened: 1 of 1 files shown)


turannul:
Well that's wrong, pendingCharge is usually mean about to start as its name might suggest.

notiant:
pendingCharge is also true if the battery is idle because a charging threshold is set. I tested it and I need to set both UPowerDeviceState.FullyCharged and UPowerDeviceState.PendingCharge to false to not show the battery as charging in this case.

turannul:
> pendingCharge is also true if the battery is idle because a charging threshold is set. I tested it and I need to set both UPowerDeviceState.FullyCharged and UPowerDeviceState.PendingCharge to false to not show the battery as charging in this case.

Technically, at threshold its just stopped charging, but still plugged. Why i think that was correct call

turannul:
Your phone for example doesn't hide lightning bolt while at %80. 
** edit it has ability to show differently (which i didn't have opportunity implement due to how it designed to) **

PR #1474, opened 2026-01-20
Battery: Add new icon and make use of it

turannul:
- Adds battery-charging-2 icon
- Shows new icon while plugged in when battery is not charging.

# Pull Request

## Motivation
Because why not :D?

## Type of Change
- [ ] Bug fix
- [X] New feature ???
- [ ] Breaking change
- [X] Refactoring

## Related Issue
N/A

## Testing

- [X] Tested on niri
- [ ] Tested on Hyprland
- [ ] Tested on sway

## Screenshots / Videos
N/A yet, waiting for battery to fill up.

## Checklist
- [X] Self-reviewed my code

## Additional Notes
@notiant What do you think?
Translations could changed to: Plugged in, not charging

### DIFF
--- a/Modules/Bar/Widgets/Battery.qml
+++ b/Modules/Bar/Widgets/Battery.qml
@@ -42,7 +42,7 @@ Item {
   readonly property bool isLowBattery: isReady && (!charging && !isPluggedIn) && percent <= warningThreshold
 
   // Visibility: show if hideIfNotDetected is false, or if battery is ready (after initialization)
-  readonly property bool shouldShow: !hideIfNotDetected || (isReady && (hideIfIdle ? (charging || battery.state === UPowerDeviceState.Discharging) : true))
+  readonly property bool shouldShow: !hideIfNotDetected || (isReady && (hideIfIdle ? (!charging && !isPluggedIn) : true))
   visible: shouldShow
   opacity: shouldShow ? 1.0 : 0.0
 
@@ -262,14 +262,12 @@ Item {
       if (battery.changeRate !== undefined) {
         const rate = Math.abs(battery.changeRate);
         if (charging) {
-          lines.push(I18n.tr("common.charging"));
           lines.push(I18n.tr("battery.charging-rate", {
                                "rate": rate.toFixed(2)
                              }));
         } else if (isPluggedIn) {
           lines.push(I18n.tr("battery.plugged-in"));
         } else {
-          lines.push(I18n.tr("common.discharging"));
           lines.push(I18n.tr("battery.discharging-rate", {
                                "rate": rate.toFixed(2)
                              }));


notiant:
@turannul I like the different icon for the plugged in state, that's certainly better than before. However, you still need to explicitly set `charging` to `false` for `UPowerDeviceState.FullyCharged` and `UPowerDeviceState.PendingCharge`, otherwise the bar widget still shows the charging icon if in plugged in state and hideIfIdle doesn't work.

I've also noticed for some time that occasionally when my laptop is plugged in, there are small charging/discharging rates that get reported (although the state stays at UPowerDeviceState.FullyCharged):

Screenshot_20260120_11h04m18

Maybe you could set a limit to ignore those small rates and keep the text as "Plugged in" (I think 0.5W would be a good limit).

turannul:
> Maybe you could set a limit to ignore those small rates and keep the text as "Plugged in" (I think 0.5W would be a good limit).

Great idea actually.

turannul:
states (4,5) are moved to another function, So `chargingStatus` only true when .Charging.

notiant:
> states (4,5) are moved to another function, So `chargingStatus` only true when .Charging.

`chargingStatus` defaults to true, so you need to state all the cases where it should be false

turannul:
.

> > states (4,5) are moved to another function, So `chargingStatus` only true when .Charging.
> 
> `chargingStatus` defaults to true, so you need to state all the cases where it should be false

That's where rate will step in (in next commit)

turannul:
Sorry shouldn't be taken that long, something else interrupted the process
@notiant

notiant:
I don't think you have to modify `chargingStatus()` for considering the change rate. At least on my laptop, the battery status stays at `UPowerDeviceState.FullyCharged` or `UPowerDeviceState.PendingCharge` when i get small change rates and doesn't switch to `UPowerDeviceState.Charging` or `UPowerDeviceState.Discharging`. Look at the screenshot I posted, the icon is for the plugged in state. 
What really needs to be changed is the condition for the text that is shown in the panel and the tooltip. 5W also seems a bit too high, I really think something like 0.5W is enough.
And `UPowerDeviceState.FullyCharged` and `UPowerDeviceState.PendingCharge` have to be explicitly stated as cases where `charging` is false.

turannul:
> I don't think you have to modify `chargingStatus()` for considering the change rate. At least on my laptop, the battery status stays at `UPowerDeviceState.FullyCharged` or `UPowerDeviceState.PendingCharge` when i get small change rates and doesn't switch to `UPowerDeviceState.Charging` or `UPowerDeviceState.Discharging`. Look at the screenshot I posted, the icon is for the plugged in state. What really needs to be changed is the condition for the text that is shown in the panel and the tooltip. 5W also seems a bit too high, I really think something like 0.5W is enough. And `UPowerDeviceState.FullyCharged` and `UPowerDeviceState.PendingCharge` have to be explicitly stated as cases where `charging` is false.

for testing i go all the way (openssl speed -multi 8) and it got consumes more

turannul:
> What really needs to be changed is the condition for the text that is shown in the panel and the tooltip. 5W also seems a bit too high, I really think something like 0.5W is enough.
> And `UPowerDeviceState.FullyCharged` and `UPowerDeviceState.PendingCharge` have to be explicitly stated as cases where `charging` is false.

chargingStatus is no longer defaults to true so that's won't be necessary 4,5 is handled by another function getPluggedInStatus

turannul:
> > I don't think you have to modify `chargingStatus()` for considering the change rate. At least on my laptop, the battery status stays at `UPowerDeviceState.FullyCharged` or `UPowerDeviceState.PendingCharge` when i get small change rates and doesn't switch to `UPowerDeviceState.Charging` or `UPowerDeviceState.Discharging`. Look at the screenshot I posted, the icon is for the plugged in state. What really needs to be changed is the condition for the text that is shown in the panel and the tooltip. 5W also seems a bit too high, I really think something like 0.5W is enough. And `UPowerDeviceState.FullyCharged` and `UPowerDeviceState.PendingCharge` have to be explicitly stated as cases where `charging` is false.
> 
> for testing i go all the way (openssl speed -multi 8) and it got consumes more (Even with 5w still showed 2-12w discharging shoed on

This really about the hardware of TDP and charger specs at some point.

turannul:
> > > I don't think you have to modify `chargingStatus()` for considering the change rate. At least on my laptop, the battery status stays at `UPowerDeviceState.FullyCharged` or `UPowerDeviceState.PendingCharge` when i get small change rates and doesn't switch to `UPowerDeviceState.Charging` or `UPowerDeviceState.Discharging`. Look at the screenshot I posted, the icon is for the plugged in state. What really needs to be changed is the condition for the text that is shown in the panel and the tooltip. 5W also seems a bit too high, I really think something like 0.5W is enough. And `UPowerDeviceState.FullyCharged` and `UPowerDeviceState.PendingCharge` have to be explicitly stated as cases where `charging` is false.
> > 
> > 
> > for testing i go all the way (openssl speed -multi 8) and it got consumes more (Even with 5w still showed 2-12w discharging shoed on
> 
> This really about the hardware of TDP and charger specs at some point.

If im still missing something please show me

notiant:
Yeah, with openssl speed -multi 8 the discharge rate gets indeed higher :) 
If you set the default to false there is no handling of the unknown status (wasn't this the whole reason to make the `chargingStatus()` function ?).
Also your current implementation only considers small charging rates, not small discharging rates.
And the conditions for the tooltip and panel text aren't defined by the `charging` variable at all.

turannul:
> Yeah, with openssl speed -multi 8 the discharge rate gets indeed higher :) If you set the default to false there is no handling of the unknown status (wasn't this the whole reason to make the `chargingStatus()` function ?). Also your current implementation only considers small charging rates, not small discharging rates. And the conditions for the tooltip and panel text aren't defined by the `charging` variable at all.

Originally it was, while tinkering i found better way.
Now current attempts were making (trying to) it keep it sensible to keep around.
`not small discharging rates.` That's can be easily implemented. text work is would be done, way easier.

turannul:
Im trying to make logic solid as possible. If i manage to do so this won't be topic in issues for a while...

What i have in mind as long as charger was able to deliver required (That's vary might not fully possible though) power, want to show plugged in

PR #1485, opened 2026-01-21
fix(launcher): intercept left/right key actions in grid view only

### LINKED ISSUE #1347: Move around the text in the launcher search box with left and right arrow keys
### Feature Description
Move around the text in the launcher search box with left and right arrow keys. Accordingly, Ctrl+Left and Ctrl+Right should move by words as is typically done in many programs.

### Why Is This Needed?
Although Home and End work to move the cursor around, the only way to left is by backspacing and erasing characters. This is inconvenient if you realize you mistyped for example the start of the second word. Even if I move to the start of the search box with <kbd>Home</kbd>, I cannot use <kbd>Del</kbd> to delete characters to the right so I still end up having to delete everything from the right.

Currently, the left and right arrow key functions are redundant with up and down.

### Suggested Solutions
The mechanism for moving the cursor around is already present as the functions of the Home and End key prove so there should be a way to map <kbd>Left</kbd> and <kbd>Right</kbd> to also move the cursor by one character (moving through an Arabic string or other complex scripts may be more complex but that is a problem that has been well studied already).

### Additional Context
Tested with noctalia-shell 4.0.0-1 on CachyOS niri (Wayland).

chenkeao:
# Pull Request

## Motivation
When in listview (vertical list), the left and right arrow keys are intercepted to move the focus up and down. This not only duplicates the function of the up and down arrow keys, but also prevents the cursor in input fields from moving left or right (which is even more problematic in calculator mode). Therefore, it makes sense to intercept the left and right arrow keys for horizontal movement only when in gridview.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring

## Related Issue
- Closes #1347 

## Testing

- [x] Tested on niri
- [ ] Tested on Hyprland
- [ ] Tested on sway

### DIFF
--- a/Modules/Panels/Launcher/Launcher.qml
+++ b/Modules/Panels/Launcher/Launcher.qml
@@ -871,10 +871,10 @@ SmartPanel {
                   } else if (event.key === Qt.Key_Backtab) {
                     root.onBackTabPressed();
                     event.accepted = true;
-                  } else if (event.key === Qt.Key_Left) {
+                  } else if (event.key === Qt.Key_Left && root.isGridView) {
                     root.onLeftPressed();
                     event.accepted = true;
-                  } else if (event.key === Qt.Key_Right) {
+                  } else if (event.key === Qt.Key_Right && root.isGridView) {
                     root.onRightPressed();
                     event.accepted = true;
                   } else if (event.key === Qt.Key_Up) {


Ly-sec:
Hey @chenkeao thank you very much. That works really well :heart: !

PR #1495, opened 2026-01-22
Add spaces to units in SystemMonitor

notiant:
This is a minor issue, but I think it would be more consistent with other tooltips if there would be spaces between numbers and units like "5 KB/s" instead of "5KB/s".


Ly-sec:
Hey @notiant , I totally agree with that one :heart: ! Merged

PR #1499, opened 2026-01-22
MediaMini: fix empty player not hiding

notiant:
When i pause the Crunchyroll player, after a while the MediaMini widget goes empty but doesn't hide (the progress circle is still visible) as it should when hideWhenEmpty is selected. The reason is that the player goes into some idle state where it has to  be completely restarted, but in that state its still recognized as a player in has the `canControl` property.
This PR changes the availablePlayers filter to `canPlay` instead of `canControl` which fixes this issue and might apply to other players as well, although I don't know if there are any side effects.


Ly-sec:
Awesome, thank you very much @notiant :heart: !

PR #1500, opened 2026-01-22
Battery: make health work if comma is decimal seperator & some cleanup

notiant:
This PR makes sure that the battery health can also be read on a system where a comma decimal separator is used instead of a period. I've removed capacityLevel from BatteryService since it's not used and not related to battery health anyway. Also hiding battery health completely if it's not available seems better than showing it's not available.

I also cleaned up the logic for the pluggedIn state and made sure it's the same for the panel and the tooltip.

### DIFF
--- a/Modules/LockScreen/LockScreen.qml
+++ b/Modules/LockScreen/LockScreen.qml
@@ -287,6 +287,7 @@ Loader {
                   event.accepted = true;
                 }
               }
+              
               Component.onCompleted: forceActiveFocus()
             }
 
--- a/Services/Hardware/BatteryService.qml
+++ b/Services/Hardware/BatteryService.qml
@@ -1,9 +1,9 @@
 pragma Singleton
+import QtQuick
 
 import Quickshell
 import Quickshell.Io
 import Quickshell.Services.UPower
-import QtQuick
 import qs.Commons
 import qs.Services.Networking
 import qs.Services.UI
@@ -108,9 +108,10 @@ Singleton {
                   })
 
     stdout: SplitParser {
-      onRead: function(data) {
+      onRead: function (data) {
         var line = data.trim();
-        if (line === "") return;
+        if (line === "")
+          return;
 
         var capacityMatch = line.match(/^\s*capacity:\s*(\d+(?:\.\d+)?)\s*%/i);
         if (capacityMatch) {


ItsLemmy:
for the comma vs period I think the proper way to do it is to pass an env var to ensure the resultat are locale agnostic.

```
    environment: ({
                    "LC_ALL": "C"
                  })
```

There is an example with nmcli  in /Services/Networking/NetworkService.qml around line 1150.

notiant:
@ItsLemmy Yes, that's the better way to do this. I've changed it in the code.

ItsLemmy:
@turannul any opinion on that ?     I just want to avoid pilling up PR over PR.

notiant:
@turannul  I removed the part with the rate limit because I don't think it does what it intends to do: on laptops where the state changes alternates between Charging and FullyCharged those short charging cycles would draw a lot more than 5W and you probably don‘t have any trickle discharges while the state is Charging.

turannul:
> @turannul any opinion on that ? I just want to avoid pilling up PR over PR.

Checked seems fine also looks cleaner.

turannul:
> @turannul I removed the part with the rate limit because I don't think it does what it intends to do: on laptops where the state changes alternates between Charging and FullyCharged those short charging cycles would draw a lot more than 5W and you probably don‘t have any trickle discharges while the state is Charging.

I knew that would happen and i don't think there a simple works for everyone kind of solution. 

While you're at it. Can you check,lock screen shows: [battery-exclamation](https://tabler.io/icons/icon/battery-exclamation) 
Where it should be showing this: [battery-charging-2](https://tabler.io/icons/icon/battery-charging-2)

turannul:
> > @turannul I removed the part with the rate limit because I don't think it does what it intends to do: on laptops where the state changes alternates between Charging and FullyCharged those short charging cycles would draw a lot more than 5W and you probably don‘t have any trickle discharges while the state is Charging.
> 
> 
> 
> I knew that would happen and i don't think there a simple works for everyone kind of solution. 
> 
> 
> 
> While you're at it. Can you check,lock screen shows: [battery-exclamation](https://tabler.io/icons/icon/battery-exclamation) 
> 
> Where it should be showing this: [battery-charging-2](https://tabler.io/icons/icon/battery-charging-2)

image

notiant:
The lock screen icon is now fixed. I also implemented stopping the countdown timer with the escape key.

ItsLemmy:
Nice, thanks to you both for the follow up!

Issue #1425, opened 2026-01-17
OSD/Brightness: Text alignment


twilfredo:
### Description

When adjusting the brightness, when the OSD slider shows up (see image) the % text shows outside the OSD box for the very first time. It seems to fix it self for subsequent changes while the OSD box is still up.

### Steps to Reproduce
1. Change brightness once using the control center widget or keybinding

### Expected Behavior

The % text to be inside the OSD pop up box.

### Screenshots

<img width="789" height="261" alt="Image" src="https://github.com/user-attachments/assets/1e97a0fd-ddcc-467e-afa5-23dc111dec77" />

### Environment
- Distro:  Fedora
- Compositor: Niri
- Noctalia-shell Version: 4.1
- Monitor resolution: 2880 x 1800

twilfredo:
On the first render, when the OSD is created I see:

```
  INFO qml: [20260117-192847]            OSD RowLayout.width: -44
  INFO qml: [20260117-192847]            OSD RowLayout.parent (contentLoader.item).width: -18
```

So,

```
diff --git a/Modules/OSD/OSD.qml b/Modules/OSD/OSD.qml
index f822ac21..f29fa13d 100644
--- a/Modules/OSD/OSD.qml
+++ b/Modules/OSD/OSD.qml
@@ -30,7 +30,7 @@ Variants {
 
     required property ShellScreen modelData
 
-    active: false
+    active: true
 
     // OSD State
     property int currentOSDType: -1 // OSD.Type enum value, -1 means none
@@ -568,7 +568,6 @@ Variants {
             osdItem.visible = false;
             root.currentOSDType = -1;
             root.lastLockKeyChanged = "";
-            root.active = false;
           }
         }
```

seems to fix this issue so I suppose it's the way things are initialized...? but I assume we don't want to do this?

ItsLemmy:
Thanks twilfredo for the debugging helped a lot.  I have heard about this bug for a long time but was never able to reproduce.   Testing a blind fix, that should do it.      Please report back.

Probably fixed by 32022eaf

twilfredo:
> Thanks twilfredo for the debugging helped a lot. I have heard about this bug for a long time but was never able to reproduce. Testing a blind fix, that should do it. Please report back.
> 
> Probably fixed by [32022ea](https://github.com/noctalia-dev/noctalia-shell/commit/32022eaf5805abf3d0ad9a2cbae0799700dd7938)

[...]

PR #1513, opened 2026-01-23
fix(osd): defer first show to prevent percentage text overflow

twilfredo:
# Pull Request

## Motivation

On first OSD display, the percentage text would render outside the bounding box because the layout hadn't finished positioning children.

Add a small delay 30ms delay before rendering the OSD, which seems to be enough time to allow for things to synchronize.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring

## Related Issue
- Closes https://github.com/noctalia-dev/noctalia-shell/issues/1425

## Testing

- [x] Tested on niri
- [ ] Tested on Hyprland
- [ ] Tested on sway

## Screenshots / Videos

## Checklist
- [ x] Code follows project style guidelines
- [ x] Self-reviewed my code
- [ x] No new warnings or errors

### DIFF
--- a/Modules/OSD/OSD.qml
+++ b/Modules/OSD/OSD.qml
@@ -740,17 +740,24 @@ Variants {
           }
         }
 
+        // Delay showing the OSD to allow the layout to settle after activation.
+        // Without this, the percentage text renders outside the box on first
+        // show.
+        Timer {
+          id: showDelayTimer
+          interval: 30
+          onTriggered: {
+            osdItem.visible = true;
+            osdItem.opacity = 1;
+            osdItem.scale = 1.0;
+            hideTimer.start();
+          }
+        }
+
         function show() {
           hideTimer.stop();
           visibilityTimer.stop();
-          osdItem.visible = true;
-
-          Qt.callLater(() => {
-                         osdItem.opacity = 1;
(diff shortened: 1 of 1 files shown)


twilfredo:
I am not sure if a 30ms delay here has other implications/is acceptable here, so a review on that would be good. Note: I also tested by making the OSD visible with a differed callback with `callLater()` which also worked.. but this seemed less complicated.

ItsLemmy:
Can you try another of my attempt at blind fixing 6a8ecfb3.
If that does not work we might have to go the Timer way, but I'd like to avoid it as much as possible.

twilfredo:
> Can you try another of my attempt at blind fixing [6a8ecfb](https://github.com/noctalia-dev/noctalia-shell/commit/6a8ecfb321a9142921e2e83d0625f0a6e99dbbbf). If that does not work we might have to go the Timer way, but I'd like to avoid it as much as possible.

I agree, it would be ideal to have this working without a delay. I just tested https://github.com/noctalia-dev/noctalia-shell/commit/6a8ecfb321a9142921e2e83d0625f0a6e99dbbbf  but the issue still persists.

image

ItsLemmy:
Oh well, timer it is then! thank you

PR #1526, opened 2026-01-24
fix(system-monitor): correct cpu temp tooltip to display °C instead of %

chenkeao:
# Pull Request

## Motivation
correct a typo

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring


notiant:
Thanks for correcting that, can you fix gpuTemp in line 161 as well?

PR #1531, opened 2026-01-24
feat: add bar insland style

Cleboost:
# Pull Request

## Motivation
Make a bar style like island (not sure what is said is clear)

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring

## Testing

- [ ] Tested on niri
- [x] Tested on Hyprland
- [ ] Tested on sway

## Screenshots / Videos
image

## Checklist
- [x] Self-reviewed my code

## Additional Notes
Note sure about some part of my code, i used IA for `Modules/MainScreen/SmartPanel.qml` because i'm not sure how can i do that.

### DIFF
--- a/Modules/Panels/Settings/Tabs/Bar/AppearanceSubTab.qml
+++ b/Modules/Panels/Settings/Tabs/Bar/AppearanceSubTab.qml
@@ -10,112 +10,125 @@ ColumnLayout {
     spacing: Style.marginL
     Layout.fillWidth: true
 
-  NComboBox {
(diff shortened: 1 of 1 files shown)


ItsLemmy:
Too many stuff have changed and would need fixing. 

I find it cute and all, but from now on we also need to account for niri's blur region, so any changes to the corners computation may need some equivalent computation in noctalia-qs so the blur is applied similarily.

PR #1554, opened 2026-01-25
fix(lockscreen): retry pam when fingerprint not ready yet

ggyorfi:
## Motivation
  Fix fingerprint authentication timing issue on lock screen. Some fingerprint readers need time to wake up after sleep, causing
  PAM's password prompt to arrive before the sensor is ready. The lock screen now retries authentication (up to 8 times, 500ms
  interval) instead of immediately showing the password prompt.

  ## Type of Change
  - [x] Bug fix

  ## Related Issue
  - (none)

  ## Testing
  - [X] Tested on niri
  - [ ] Tested on Hyprland
  - [ ] Tested on sway

  ## Screenshots / Videos
  (not applicable - behaviour fix)

  ## Checklist
  - [x] Self-reviewed my code

  ## Additional Notes
  - Adds `unlocked`/`failed` signals
  - Keeps upstream's `showInfo`/`showFailure` mutual exclusivity handlers
  - Retry parameters (count, interval) are hardcoded with sane defaults rather than exposed in settings to avoid unnecessary
  complexity for users

### DIFF
--- a/Assets/settings-default.json
+++ b/Assets/settings-default.json
@@ -90,7 +90,7 @@
     "telemetryEnabled": false,
     "enableLockScreenCountdown": true,
     "lockScreenCountdownDuration": 10000,
-    "autoStartAuth": false
+    "autoStartAuth": false,
     "allowPasswordWithFprintd": false
   },
   "ui": {
--- a/Modules/LockScreen/LockContext.qml
+++ b/Modules/LockScreen/LockContext.qml
@@ -7,8 +7,6 @@ import qs.Services.System
 
 Scope {
   id: root
-  signal unlocked
-  signal failed
 
   property string currentText: ""
   property bool waitingForPassword: false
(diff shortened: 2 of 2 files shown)


brettalcox:
i tested this on hyprland with a Digital Person 4500 and works as i'd expect. nice work!

ggyorfi:
@brettalcox Thank you for looking into it! I've updated the branch with the latest upstream changes. Would you mind giving it another go when you have a moment?

brettalcox:
@ggyorfi the behavior is slightly different now. it now is back to the behavior of requiring to hit enter first to prompt for fingerprint. prior to your rebase it was automatically prompting for fingerprint without having to hit enter

ggyorfi:
@brettalcox Yes, this is a consequence of the upstream change. From a security perspective it makes sense not to load PAM config files from user-writable locations like ~/.config.

First you can try turning **Settings > Lock Screen > Auto-start authentication** on

Note: /etc/pam.d/noctalialock support is included only in my fix (so this will not work on upstream/current releases until that change is applied).

With my fix, you can add a Noctalia-specific PAM file so the lock screen uses fingerprint first (with password fallback):

```bash
sudo tee /etc/pam.d/noctalialock > /dev/null <<'EOF'
# PAM config for Noctalia lock screen (used only with my fix)
# Fingerprint first, fallback to password

auth sufficient pam_fprintd.so timeout=-1
auth required pam_unix.so
EOF

sudo chown root:root /etc/pam.d/noctalialock
sudo chmod 644 /etc/pam.d/noctalialock
```

(Optional) Verify:
```bash
ls -l /etc/pam.d/noctalialock
```
> [!CAUTION]
> Editing PAM files incorrectly can lock you out of your system.
> It is strongly recommended to stay logged in on a separate TTY
> and have a root shell available while testing.

Ly-sec:
Is there any need to resolve the conflicts? Seems like we kind of got that toggle.

ItsLemmy:

Cheers!

Issue #1248, opened 2026-01-03
Battery percentage on the lockscreen shows up with a visible delay


TheDidek:
### Description
On Noctalia lockscreen everything shows up fast and smooth, but the battery text is showing up with a visible delay. This is a minor annoying issue.

### Steps to Reproduce
1. Lock the screen (using for example. Quickshell IPC)
2. Look near the keyboard layout indicator
3. See the error.

### Expected Behavior
Battery label should be visible in the same time as the rest of the widgets.

### Screenshots

Right on the lockscreen start:
<img width="1920" height="1440" alt="Image" src="https://github.com/user-attachments/assets/c7c68e8c-ed88-4bbc-a575-78c774e8b63d" />

A second or two later:

<img width="1920" height="1440" alt="Image" src="https://github.com/user-attachments/assets/9d97db1a-40af-4706-ad46-54f836cd32fe" />

### Environment
- Distro:  ArchLinux
- Compositor: Niri
- Noctalia-shell Version: 3.8.1

whiskeyPeak:
This is happening because there is a timer in Lockscreen.qml line 85 though I can't see any reason why it might be there. Removing the timer doesn't seem to impact the UI minus the delay being removed.

```
          Item {
            id: batteryIndicator
            property bool initializationComplete: false
            Timer {
              interval: 500
              running: true
              onTriggered: batteryIndicator.initializationComplete = true
            }
```

@Ly-sec Any thoughts?

Ly-sec:
> This is happening because there is a timer in Lockscreen.qml line 85 though I can't see any reason why it might be there. Removing the timer doesn't seem to impact the UI minus the delay being removed.
> 
> ```
>           Item {
>             id: batteryIndicator
>             property bool initializationComplete: false
>             Timer {
>               interval: 500
>               running: true
>               onTriggered: batteryIndicator.initializationComplete = true
>             }
> ```
> 
> @Ly-sec Any thoughts?

Hey, I'm quite certain that I added that timer so we got enough time to grab Bluetooth device batteries before showing the battery. I'll take a look later again.

turannul:
@TheDidek @whiskeyPeak is this still an issue on latest -git version

whiskeyPeak:
@turannul The issue is still there.

turannul:
This time im confident that's no longer an issue.

whiskeyPeak:
@turannul Yup, issue has been fixed on my end :)

PR #1556, opened 2026-01-25
LockScreen: no delay for battery if laptopBattery

notiant:
This should solve the issue talked about in #1248. I have not tested it on a desktop PC with a connected bluetooth device that has a battery, though.

### DIFF
--- a/Modules/Panels/Settings/Bar/WidgetSettings/BatterySettings.qml
+++ b/Modules/Panels/Settings/Bar/WidgetSettings/BatterySettings.qml
@@ -122,7 +122,7 @@ ColumnLayout {
 
     NIconButton {
       icon: "refresh"
-      tooltipText: "Refresh device list"
+      tooltipText: I18n.tr("tooltips.refresh-devices")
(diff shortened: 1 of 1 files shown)


ItsLemmy:
I'm on desktop with only BT earbuds with 100% battery, and for me the current lockscreen implementation shows an empty battery cell at 0%.

I don't think showing BT device on the lockscreen is very interesting.  So maybe on the lockscreen we could just only show the laptop battery if there is one, and nothing else.  That would keep things simple.  

What do you guys think?

@turannul

turannul:
> I'm on desktop with only BT earbuds with 100% battery, and for me the current lockscreen implementation shows an empty battery cell at 0%.
> 
> I don't think showing BT device on the lockscreen is very interesting. So maybe on the lockscreen we could just only show the laptop battery if there is one, and nothing else. That would keep things simple.
> 
> What do you guys think?
> 
> @turannul

Indeed, bt battery on lock screen doesn't make sense for headphones, can't say same for keyboard/mouse.

PR #1566, opened 2026-01-25
refactor(CurrentScreenDector): Use WM native methods when possible

frap129:
# Pull Request
This PR refactors CurrentScreenDetector uses Quickshell's built in methods for supported WMs, and only falling back to the transparent panel method if no other method is available. This should give more consistent behavior across WMs. 

## Motivation
When using Noctalia on Hyprland with a 2 monitor setup. I noticed that my keybinds open shell panels and windows on my first display, and not the focused display, unless: 

1. There are no windows on the current workspace
2. I use the mouse to focus the bar before using a keybind

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Refactoring

## Testing
I'm on an immutable/atomic system, so switching window managers is a pain. Hoping someone else can test on the other supported WMs

- [ ] Tested on niri
- [x] Tested on Hyprland
- [ ] Tested on sway

## Screenshots / Videos

## Checklist
- [x] Self-reviewed my code

## Additional Notes
This is my first time working with QML and Quickshell, so pardon any mistakes or formatting issues I may have made.

### DIFF
--- a/Services/Compositor/CompositorService.qml
+++ b/Services/Compositor/CompositorService.qml
@@ -316,6 +316,14 @@ Singleton {
     return null;
   }
 
+  // Get focused screen from compositor
+  function getFocusedScreen() {
+    if (backend && backend.getFocusedScreen) {
+      return backend.getFocusedScreen();
+    }
+    return null;
+  }
+
   // Get focused window title
   function getFocusedWindowTitle() {
     if (focusedWindowIndex >= 0 && focusedWindowIndex < windows.count) {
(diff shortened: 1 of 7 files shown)


ItsLemmy:
I like the idea, but the logic for the compositors should be in their respective  implementation.
I would go with CompositorService.getFocusedScreen()   and implement it in each backend.

frap129:
> I like the idea, but the logic for the compositors should be in their respective implementation. I would go with CompositorService.getFocusedScreen() and implement it in each backend.

Good point, fixed

ItsLemmy:
Thanks!

PR #1613, opened 2026-01-29
fix(i18n): add CJK-specific date format unit suffixes

elmitash:
# Pull Request

## Motivation
This PR improves the date localization for CJK (Chinese, Japanese, and Korean) locales. In these languages, date strings feel unnatural without specific unit suffixes (e.g., '년' and '일' in Korean).
Currently, the clock widget displays dates as bare numbers (like "2026 1월 29" in Korean) because the default format strings do not include these units. This PR introduces a cjkDateFix helper function that automatically appends the correct unit suffixes depending on the user's locale, ensuring a much more polished and native experience for CJK users.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring

## Related Issue
- Closes #(N/A)

## Testing

- [x] Tested on niri
- [ ] Tested on Hyprland
- [ ] Tested on sway

## Screenshots / Videos
Display of the date on the Bar widget and the tooltip.
### Before :  
Locale code: ko-KR
image
Locale code: ja
image
Locale code: zh-CN
image
Locale code: zh-TW
image

The clock settings widget
image

### After : 
 Locale code: ko-KR
image
Locale code: ja
image
Locale code: zh-CN
image
Locale code: zh-TW
image

The clock settings widget
image

## Checklist
- [x] Self-reviewed my code

## Additional Notes
The cjkDateFix function in I18n.qml is designed to be non-intrusive. It only acts when the locale is set to ko-KR, ja, or zh-CN, zh-TW, and it carefully avoids modifying weekday tokens (ddd, dddd) or duplicating existing units if the user has already manually added them in the settings.

### DIFF
--- a/Modules/Bar/Widgets/Taskbar.qml
+++ b/Modules/Bar/Widgets/Taskbar.qml
@@ -607,7 +607,7 @@ Item {
 
     return Math.round(calculatedWidth);
   }
-  readonly property real contentHeight: visible ? (isVerticalBar ? Math.round(taskbarLayout.implicitHeight + Style.marginS * 2) : barHeight) : 0
+  readonly property real contentHeight: visible ? (isVerticalBar ? Math.round(taskbarLayout.implicitHeight + Style.marginS * 2) : capsuleHeight) : 0
 
   implicitWidth: contentWidth
   implicitHeight: contentHeight
@@ -786,7 +786,7 @@ Item {
               visible: shouldShowTitle
               anchors.centerIn: parent
               width: parent.width
-              height: root.barHeight
+              height: root.capsuleHeight
               color: titleBgColor
               radius: Style.radiusM
 
--- a/Services/Noctalia/UpdateService.qml
+++ b/Services/Noctalia/UpdateService.qml
@@ -11,7 +11,7 @@ Singleton {
   id: root
 
   // Version properties
-  readonly property string baseVersion: "4.2.5"
+  readonly property string baseVersion: "4.2.6"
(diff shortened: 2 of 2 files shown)


elmitash:
This is my first PR in the Linux ecosystem. I am excited to contribute to a project I use and love. Thank you for your hard work on Noctalia, and I hope this fix helps other CJK users as well.

bokicoder:
I don't need this suffix. Is there a way to disable it?

elmitash:
> I don't need this suffix. Is there a way to disable it?

That's a very valid point. I'll update the logic to exclude format strings that already use delimiters like '/' or '-' or '.', so it won't interfere with those custom preferences.
It's late at night here, so I'll finish the work and push the update tomorrow. Thanks for the feedback!

ItsLemmy:
Feels very hacky.   Can't those suffix just be added manually in the clock settings as you have full controls over the presentation ?

elmitash:
> Feels very hacky. Can't those suffix just be added manually in the clock settings as you have full controls over the presentation ?

I completely agree it’s hacky. However, a "proper" fix would require significant effort, while leaving it as-is makes the shell feel unpolished for CJK users.
I believe this is a practical compromise—a small addition affecting only 4 locales to provide a native default experience.
What do you think?

elmitash:
After thinking it over, I agree with your point. This kind of logic doesn't really belong in the codebase. I'll close the PR—thanks for the feedback!

PR #1607, opened 2026-01-28
feat(widget): display disk space in systemMonitor Widget

thmoee:
# Pull Request

## Motivation
Display the available disk space as percentage or in numbers in the bar. Adding a new threshold for the available disk space.

## Type of Change
Mark the relevant option with an "x".
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring

## Related Issue
- Closes #1460 

## Testing
Describe how you tested your changes and mark the relevant items.

- [ ] Tested on niri
- [x] Tested on Hyprland
- [ ] Tested on sway
- [x] Tested with different bar positions and density settings
- [x] Tested at different interface scaling values
- [ ] Tested with multiple monitors (if applicable)

## Screenshots / Videos
If applicable, include screenshots or videos to help illustrate your changes.

https://github.com/user-attachments/assets/c0063c80-986d-462a-9462-edf0aa1785c5

## Checklist
- [x] Code follows project style guidelines
- [x] Self-reviewed my code
- [ ] No new warnings or errors
- [ ] Documentation or comments updated (if relevant)

## Additional Notes
Didn't translated the new setting to all languages. Only did languages where i was comfortable to know which language it was exactly. Used DeepL for my translations.
If there is anything i missed or you have a different idea how to handle it for vertical aligned bars, let me know

PR #1622, opened 2026-01-29
feat: add CPU frequency and free disk space display options

### LINKED ISSUE #1460: Show available disk space in SystemMonitor widget
### Feature Description
The [SystemMonitor widget](https://github.com/noctalia-dev/noctalia-shell/blob/main/Modules/Bar/Widgets/SystemMonitor.qml) currently only makes it possible to show disk usage as a percentage. It would be nice to be able to display the available disk space (in TB/GB/MB) as well.

### Why Is This Needed?
Percentage isn't enough to assess whether one has enough disk space available for some operations.

### Suggested Solutions
One way of implementing it would be to use a template string with documented variables `{percent}`/`{available}` to let users format the disk usage as they fancy.

### Additional Context
This feature was one of the requests from https://github.com/noctalia-dev/noctalia-shell/issues/1334 , which got closed for some reason, despite the feature hasn't been implemented yet (nor marked as a no-go), hence I'm taking the liberty of raising it again in this issue.

xzeldon:
# Pull Request

## Motivation
The current `SystemMonitor` widget only displays resource usage as percentages. While useful, percentages don't always provide enough context for resource management (e.g., knowing exactly how many GB are left for a large download or monitoring actual CPU clock speeds).

This PR introduces two features:
1. **CPU Frequency display**: Optionally show the actual clock speed in GHz instead of a load percentage.
2. **Available Disk Space**: Optionally show the remaining disk space in GB. Unlike a simple calculation, this uses the `Avail` column from `df` to correctly account for filesystem overhead and reserved blocks, as requested in #1460.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring

## Related Issue
- Closes #1460

## Testing
- [x] Tested on niri
- [ ] Tested on Hyprland
- [ ] Tested on sway

## Screenshots / Videos
image
image

## Checklist
- [x] Self-reviewed my code

## Additional Notes
To prevent layout breaking:
*   **Horizontal Bar**: Displays detailed values (e.g., `3.5GHz`, `441.2G`).
*   **Vertical Bar**: Automatically falls back to the **Percentage view** (`%`). This prevents text clipping and overflow issues on narrow vertical bars where space is extremely limited.

Also i18n keys for new settings are not yet implemented. I can add them if required.

### DIFF
--- a/Modules/Bar/Widgets/SystemMonitor.qml
+++ b/Modules/Bar/Widgets/SystemMonitor.qml
@@ -77,7 +77,7 @@ Item {
     let lines = [];
 
     // CPU
-    lines.push(`${I18n.tr("system-monitor.cpu-usage")}: ${Math.round(SystemStatService.cpuUsage)}%`);
+    lines.push(`${I18n.tr("system-monitor.cpu-usage")}: ${Math.round(SystemStatService.cpuUsage)}% (${SystemStatService.cpuFreq})`);
     if (SystemStatService.cpuTemp > 0) {
       lines.push(`${I18n.tr("system-monitor.cpu-temp")}: ${Math.round(SystemStatService.cpuTemp)}°C`);
     }
@@ -254,7 +254,7 @@ Item {
           // Text mode
           NText {
             visible: !compactMode
-            text: (showCpuFreq && !isVertical) ? SystemStatService.cpuFreq.replace(" ", "") : `${Math.round(SystemStatService.cpuUsage)}%`
+            text: `${Math.round(SystemStatService.cpuUsage)}%`
             family: fontFamily
             pointSize: barFontSize
             applyUiScale: false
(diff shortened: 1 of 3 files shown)


ItsLemmy:
I don't believe CPU Frequency should hide CPU usage. In you screenshot both are active and CPU Usage is not shown.

ItsLemmy:
also #1607

xzeldon:
Thanks for the feedback. I've updated the code, and now CPU Frequency is a standalone component with its own toggle. I also added the current frequency to the tooltip.
image

The component now supports Compact Mode and works well in both horizontal and vertical bars. For this mode, I added logic that automatically tracks the maximum CPU frequency by reading `scaling_max_freq`. It updates itself dynamically, so the gauge stays correct even when you switch power profiles.
image

However, if Compact Mode is off, I had to hide the frequency on vertical bars because the text just doesn't fit the width, and hiding it seemed like the only clean way to handle it. For the icon, I'm using `cpu-usage` for now as a temporary solution since I couldn't find a better one in the font yet. Also, a quick reminder that i18n keys are still missing.

PR #1625, opened 2026-01-29
fix(bar): consistent outer bar margin spacing

anthonyhab:
## Summary
Center bar widgets properly by calculating margins dynamically based on `barHeight` and `capsuleHeight` instead of using fixed `Style.margin` values.

Widgets were positioned using fixed margins which didn't account for different bar densities, leading to inconsistent margins.

## Screenshots
*Before/after comparison showing proper widget centering*

Mini example
image
image

Spacious example
image
image

### DIFF
--- a/Modules/Bar/Bar.qml
+++ b/Modules/Bar/Bar.qml
@@ -267,7 +267,7 @@ Item {
       }
 
       // Calculate margin to center widgets vertically within the bar height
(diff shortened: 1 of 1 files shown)


ItsLemmy:
Nice, can your Math.round() those, to avoid subpixel issues.

ItsLemmy:
thanks

PR #1662, opened 2026-02-02
Improve SystemMonitor tooltip

notiant:
Adding a missing translation to the tooltip and correcting some unit formats. The tooltip is a bit to small for some languages (e.g. French) so I increased the general tooltip width slightly.


Ly-sec:
Thank you very much @notiant :heart: !

PR #1694, opened 2026-02-05
fix(display): external display brightness slider fluidity

Cleboost:
# Pull Request

## Motivation
Makes the external display brightness slider smooth instead of janky.

## Type of Change
- [x] Bug fix
- [x] Refactoring

## Testing
- [ ] Tested on niri
- [x] Tested on Hyprland
- [ ] Tested on sway

## Screenshots / Videos
Before:

https://github.com/user-attachments/assets/ad8722ad-302b-4c41-a260-fdcf05d134d0

After:

https://github.com/user-attachments/assets/aade1ac1-0ade-48ca-87a1-19c448e59b84

## Checklist
- [x] Self-reviewed my code


ItsLemmy:
Never had this lag in the UI, but I believe you... some more DDCUtil madness.
Thanks :P

PR #1697, opened 2026-02-05
Add screen argument to IPC wallpaper random

fboulay:
# Pull Request

## Motivation
I wanted to map a keybinding to change randomly the wallpaper of the current screen.

### Implementation
Add argument to specify on which screen to apply a random wallpaper.
if `all` is provided, the old behavior is kept: a random wallpaper is applied to all screen.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [x] Breaking change
- [ ] Refactoring

I don't know how to add an optional argument to the function `IpcHandler.random`. So the feature is not backward compatible. To have the previous behavior, we can pass `all` as screen name.

## Related Issue
N/A

## Testing

- [ ] Tested on niri
- [x] Tested on Hyprland
- [ ] Tested on sway

Test the IPC with the command line. I tested with different directories for each screen as well.

## Screenshots / Videos
N/A

## Checklist
- [x] Self-reviewed my code

How to update the IPC documentation ?

## Additional Notes

Is it better to keep the feature as is  (and break backward compatibility) ? Or should I create a new function in `IpcHandler` to keep the backward compatibility ?


ItsLemmy:
Thanks for the PR, There are a couple things  I need to fix, as stuff have evolved + some dangerous null != undefined.

PR #1736, opened 2026-02-08
Fix Bluetooth discovering process & some enhancements to Bluetooth- & Wifi-Panel

notiant:
1. The Bluetooth discovery process with Quickshell's native API currently doesn't work. This is because there are `startDiscovery()` and `stopDiscovery()` functions called that don't exist in Quickshell. I fixed this problem and also made sure that the discovering process can be stopped if the fallback process with bluetoothctl is used.
2. I made some changes to the connected device layout in the Bluetooth-Panel to look the same as the connected network and also slightly increased the contrast for the connected network background (which was rather poor with the new color schemes).
3. In the Wifi-Panel, the scanning box is now only shown when the scan is actually running. This has made the `hasHadNetworks` variable redundant.

### DIFF
--- a/Modules/Panels/Network/WiFiNetworksList.qml
+++ b/Modules/Panels/Network/WiFiNetworksList.qml
@@ -295,7 +295,7 @@ NBox {
                 visible: modelData.connected && NetworkService.disconnectingFrom !== modelData.ssid
                 text: I18n.tr("common.disconnect")
                 outlined: !hovered
-                fontSize: Style.fontSizeXS
+                fontSize: Style.fontSizeS
                 backgroundColor: Color.mError
                 onClicked: NetworkService.disconnect(modelData.ssid)
               }


ItsLemmy:
@turannul  can you double check

turannul:
> @turannul  can you double check

I'll when back on PC

Services/Networking/BluetoothService.qml
```diff
@@ -103,7 +103,7 @@ Singleton {
   Process {
     id: fallbackScanProcess
     // Pipe scan on and a long sleep to bluetoothctl to keep it running
-    command: ["sh", "-c", "(echo 'scan on'; sleep 3600) | bluetoothctl"]
+    command: ["sh", "-c", "trap 'kill 0' EXIT; (echo 'scan on'; sleep 3600) | bluetoothctl"]
```
turannul, line 106: That's questionable

turannul:
I think we should implement scanning/discovering while panel open - shouldn't be hard to do.

Services/Networking/BluetoothService.qml
notiant: It's necessary for the piped Bluetooth scan process to exit when you click on the stop button. It doesn't work at the moment, you can try it

notiant:
> I think we should implement scanning/discovering while panel open - shouldn't be hard to do.

This is already implemented in the network panel, but I don't think it's necessary for the Bluetooth panel.

turannul:
In the process, can we do something for wrong passwords... (Wi-Fi Auth)

Instead of forgetting it?

turannul:
> > I think we should implement scanning/discovering while panel open - shouldn't be hard to do.
> 
> This is already implemented in the network panel, but I don't think it's necessary for the Bluetooth panel.

I think it is bg bluetooth scans, if no one goin to notice is pointless. nor running command for an hour.

notiant:
> > > I think we should implement scanning/discovering while panel open - shouldn't be hard to do.
> > 
> > 
> > This is already implemented in the network panel, but I don't think it's necessary for the Bluetooth panel.
> 
> I think it is bg bluetooth scans, if no one goin to notice is pointless. nor running command for an hour.

The scan doesn't run for an hour, the `sleep 3600` is just so the scan is not terminated instantly (not my code, though). The scan currently stops automatically after 10s, but bluetoothctl also has its own stop mechanism that sets in after around 30s.

turannul:
> > > > I think we should implement scanning/discovering while panel open - shouldn't be hard to do.
> 
> > > 
> 
> > > 
> 
> > > This is already implemented in the network panel, but I don't think it's necessary for the Bluetooth panel.
> 
> > 
> 
> > I think it is bg bluetooth scans, if no one goin to notice is pointless. nor running command for an hour.
> 
> 
> 
> The scan doesn't run for an hour, the `sleep 3600` is just so the scan is not terminated instantly (not my code, though). The scan currently stops automatically after 10s, but bluetoothctl also has its own stop mechanism that sets in after around 30s.

As of my experience it does :p 

We should've controlling (scan/discovery) more strictly.

Which i'll be planned doing later.

ItsLemmy:
maybe it's a good time to start a "Bluetooth improvement"  thread on discord?!

turannul:
> maybe it's a good time to start a "Bluetooth improvement" thread on discord?!

I think so

PR #1740, opened 2026-02-08
Make overview blur and tint customizable

whiskeyPeak:
# Pull Request

## Motivation
Makes the blur and tint customizable for the overview. I feel like this would be a rather nice feature to have, since blur is something people can have a wide range of opinions on.

I also made the max blur 64 with a range of 0.0 to 1.5 since that gives the blur a wide range of options. Keeping the max blur value to 32 makes the blur a lot darker at higher values, which should instead be modifiable with the tint. Also not sure about all the translations so I only added English atm.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring

## Testing

- [x] Tested on niri
- [ ] Tested on Hyprland
- [ ] Tested on sway

## Screenshots / Videos

https://github.com/user-attachments/assets/1dfe18ac-a6a9-4dc9-9c31-a386b4b69664

## Checklist
- [x] Self-reviewed my code

### DIFF
--- a/Modules/Background/Overview.qml
+++ b/Modules/Background/Overview.qml
@@ -5,6 +5,7 @@ import Quickshell.Wayland
 import qs.Commons
 import qs.Services.Compositor
 import qs.Services.UI
+import qs.Services.Power
 
 Loader {
   active: CompositorService.isNiri && Settings.data.wallpaper.enabled && Settings.data.wallpaper.overviewEnabled
@@ -100,7 +101,7 @@ Loader {
         layer.enabled: true
         layer.smooth: false
         layer.effect: MultiEffect {
-          blurEnabled: true
+          blurEnabled: !PowerProfileService.noctaliaPerformanceMode
           blur: Settings.data.wallpaper.overviewBlur
           blurMax: 48
         }
--- a/Modules/Panels/Settings/Tabs/Wallpaper/LookAndFeelSubTab.qml
+++ b/Modules/Panels/Settings/Tabs/Wallpaper/LookAndFeelSubTab.qml
@@ -74,7 +74,7 @@ ColumnLayout {
     Layout.fillWidth: true
     label: I18n.tr("panels.wallpaper.look-feel-overview-blur-label")
     description: I18n.tr("panels.wallpaper.look-feel-overview-blur-description")
-    visible: Settings.data.wallpaper.enabled && Settings.data.wallpaper.overviewEnabled
+    visible: CompositorService.isNiri
     from: 0.0
     to: 1.0
     stepSize: 0.01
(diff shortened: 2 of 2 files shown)


Modules/Panels/Settings/Tabs/Wallpaper/LookAndFeelSubTab.qml
```diff
@@ -69,4 +69,30 @@ ColumnLayout {
     text: Math.round(Settings.data.wallpaper.transitionEdgeSmoothness * 100) + "%"
     defaultValue: Settings.getDefaultValue("wallpaper.transitionEdgeSmoothness")
   }
+
+  NValueSlider {
+    Layout.fillWidth: true
+    label: I18n.tr("panels.wallpaper.look-feel-overview-blur-label")
+    description: I18n.tr("panels.wallpaper.look-feel-overview-blur-description")
+    from: 0.0
+    to: 1.5
```
ItsLemmy: "blur" amount is [0..1] per the documentation. https://doc.qt.io/qt-6/qml-qtquick-effects-multieffect.html#blur-prop

blurMax: 64 is gonna be intense on the GPU but I guess it's fine, as it's cached by layer.enabled: true.

Modules/Background/Overview.qml
```diff
@@ -101,15 +101,15 @@ Loader {
         layer.smooth: false
         layer.effect: MultiEffect {
           blurEnabled: true
```
ItsLemmy: good opportunity to replace blurEnabled: true to 
blurEnabled: !PowerProfileService.noctaliaPerformanceMode

notiant:
I think those settings should be only visible on Niri?

Modules/Panels/Settings/Tabs/Wallpaper/LookAndFeelSubTab.qml
whiskeyPeak: I played a bit with different values and found that with blurMax: 48 and blurMultiplier: 0.5, you get a decent blur range that actually still looks quite good within 0.0 to 1.0.

Modules/Background/Overview.qml
whiskeyPeak: ~~I don't know how i feel about only enabling blur for the performance mode. I generally only ever run my laptop in balanced or power saver mode and feel like the UI should look consistent regardless of mode. Even on Power Saver mode on a 6 years old intel iGPU, I see no performance problems at all. Everything still runs like butter~~

whiskeyPeak:
> I think those settings should be only visible on Niri?

It should probably be dependent on whether the user has checked the `enable wallpaper overview` feature.

Screenshot from 2026-02-09 01-03-08

notiant:
> > I think those settings should be only visible on Niri?
> 
> It should probably be dependent on whether the user has checked the `enable wallpaper overview` feature.

Yes, I don't have that setting on Hyprland. Also, Noctalia Performance is a mode that disables animations and shadows, it's not your one of your power profiles.

whiskeyPeak:
Ahhh, I see. That makes sense. Will check for niri and performance mode then.

whiskeyPeak:
@ItsLemmy Any chance this can be merged yet?

PR #1744, opened 2026-02-09
animations: simplify settings and fix collapsible animations

vanillasoap:
simplified the animation settings panel — swapped out the 10 individual controls for a preset dropdown (Default, Fade, Pop, Slide) with per-component overrides in an Advanced section underneath. removed the easing curve config entirely and hardcoded the defaults.

also removed the hover color animations from menus/dropdowns, and fixed NCollapsible expand/collapse being janky.

screenshot_20260209_045128


ItsLemmy:
The AI slop is real

vanillasoap:
> The AI slop is real

😅 tidied up a bit. thanks for building noctalia.

ItsLemmy:
Thank you for this contribution. After reviewing the code, we have decided not to merge this PR at this time.

While we appreciate the intent behind the UI updates, the implementation is quite invasive and introduces a significant number of bindings, which raises concerns regarding resource usage and performance. Additionally, we feel the visual impact of the new animations does not justify the added complexity and long-term maintenance overhead they would require.

PR #1755, opened 2026-02-09
Adds background blur option to the session menu.

brunoorsolon:
# Pull Request

## Motivation
I felt like the session menu needed some visual separation from the screen. Since it could be considered a matter of preference, I have implemented it as a toggle option in the UI. Blur accent color can be chosen from any of the default colors.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring

## Related Issue
N/A

## Testing

- [ ] Tested on niri
- [x] Tested on Hyprland
- [ ] Tested on sway

## Screenshots / Videos
image
image
image

## Checklist
- [x] Self-reviewed my code

## Additional Notes
- I have tested different monitor resolutions, but I could not test multiple monitors at the same time.
- I don't have Niri or Sway desktops for testing but I assume the changes would work in any desktop. I have tried to use as much as possible the same coding standard and style of the current Session Menu, so compatibility should be high.
- I have added translations for all languages, but most of them are AI generated. I only know Portuguese, English and a little bit of Spanish.  I'm not sure what's the guideline for the project in this aspect, if it's better to add AI auto translations or to not add any translations at all and deal with them later when someone who is fluent in the language is available to do them. Let me know if you wish me to remove the AI translated parts.


brunoorsolon:
I'm putting this to draft because the code is a bit stale. I tested it again locally and the effect is tied o the general desktop dimming now (I dunno if it was the case before or just some changes to the core of Noctalia shell)

I will revisit this when I have some free time.


Also, is this a wanted feature at all? If it's something that won't be merged because of the project guidelines, let me know and I can just make this into something for myself, will save me some trouble dealing with rebases and conflict resolving trying to catch up with main branch,

ItsLemmy:
I don't think we want this right now as it's very hacky + niri is on the way to support blur so it's not really interesting to do the blur in QML.

PR #1761, opened 2026-02-10
use the correct axis when calculating content height

tuibird:
# Pull Request
Use the correct axis when calculating content height in vertical bar layout. Have removed what i can only assume is legacy code. Also a minor change to the surrounding comment so its in line with its horizontal equivalent.

## Motivation
I noticed the animation was jerky and incorrect when hovering expanding widgets such as brightness or volume.

## Type of Change
- [x] Bug fix

## Testing

- [x] Tested on niri
- [ ] Tested on Hyprland
- [ ] Tested on sway

## Checklist
- [x] Self-reviewed my code


ItsLemmy:
Thanks!

PR #1779, opened 2026-02-11
style: add centralized easing curve constants

vanillasoap:
This adds the standard Material Design 3 easing curves as named constants in `Style.qml`.

Material 3 defines different Bézier curves for common animation cases: `emphasized` (general), `emphasizedDecelerate` (entering), and `emphasizedAccelerate` (exiting). Centralizing them avoids duplicating magic numbers and keeps motion consistent across components.

It's only six lines of `readonly` properties. No existing behavior changes, and there's no performance impact since they are just constants sitting there until something decides to use them.

Reference: [https://m3.material.io/styles/motion/easing-and-duration](https://m3.material.io/styles/motion/easing-and-duration)


ItsLemmy:
But noctalia is not a Material Design 3 project.

vanillasoap:
Fair point, Noctalia isn't a Material project. These are just well-researched bezier curves for common motion cases though. The math isn't inherently "Material" any more than using a sine wave makes a project "CSS Animations." They just happen to be documented under Google's motion spec.

Right now there are no centralized easing constants, so animations either hardcode curves or fall back to Qt defaults. Happy to rename them if the M3 reference is the concern.

PR #1793, opened 2026-02-11
Added button for plugin debug mode toggle

spiros132:
# Pull Request

## Motivation
To make it easier for plugin developers to enable the "debug mode", I've added a button in the plugins tab, where you can now toggle off and on the debug mode for an individual plugin. This makes it easier to start the debug mode, and only start it for a particular plugin. This should make the shell itself use less resources for people that want to enable the debug mode for one particular plugin instead of all the plugins. 

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring

## Related Issue
No related issue

## Testing
I've tested it while writing the video-wallpaper plugin and it works correctly. As in the screenshots, there's a button now to the left of the settings icon in the installed plugins tab. 

- [x] Tested on niri
- [ ] Tested on Hyprland
- [ ] Tested on sway

## Screenshots / Videos
When debug mode is disabled, there's an icon to the side of the settings button, that has a tooltip that says "Enabled debug mode for this plugin" 
image

When debug mode is enabled, that icon becomes a bit different with a tooltip saying "Disable debug mode for this plugin.
image

The tooltip:
image

## Checklist
- [x] Self-reviewed my code

## Additional Notes
I'm not sure how the translation works here, I've added the English one but I don't know if I'm supposed to do it for all the other ones.

### DIFF
--- a/Assets/Translations/de.json
+++ b/Assets/Translations/de.json
@@ -1030,6 +1030,8 @@
       "fonts-monospace-search-placeholder": "Monospace-Schriftarten suchen...",
       "fonts-reset-scaling": "Skalierung zurücksetzen",
       "fonts-title": "Schriftarten",
+      "keybinds-conflict-description": "Die Tastenkombination ist bereits {action} zugewiesen.",
+      "keybinds-conflict-title": "Tastenbelegungskonflikt",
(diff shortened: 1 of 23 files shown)


ItsLemmy:
Excellent, thanks!

Issue #1641, opened 2026-01-31
Add wallpaper  lockscreen specific


TotoBlues:
### Feature Description
What feature would you like to see?

Hello,

First of all I would like to thanks all of you for the work, it really makes Archlinux simpler to use to have this ecosystem !

**My Question is:**
Do you think it's possible to add a specific wallpaper for lockscreens instead of using screen's wallpaper ?

that's all thanks again for the rest

### Suggested Solution
adding a feature flag with a path to specific wallpaper to the `function requestCachedWallpaper` ? (I don't hard look the code just asking)

D3rJust1n:
If no one else does it before me, I would actually quite like to give it a try at the weekend.

Of course, if no one has anything against it.

chenkeao:
The idea of setting a separate wallpaper for the lock screen makes sense. However, it requires some changes to `WallpaperService` and `WallpaperPanel` to support this properly. I’d avoid implementing a separate selector just for the lock screen, as that feels redundant and could lead to an inconsistent user experience.

I'm happy to open a PR for this, but I wanted to check with the maintainers first. Any thoughts? @ItsLemmy @Ly-sec

Ly-sec:
Thanks so much for reaching out and sharing this with us!

Right now, our team is entirely focused on building v5. Since the v4 lifecycle has officially come to an end, we are wrapping up old issues and won’t be shipping any further updates or fixes to that codebase.

We are closing this issue for now, but we truly appreciate you taking the time to report it. We aren’t quite ready to open up the v5 tracker for public bug reports or feature requests just yet, but we’ll definitely keep your feedback in mind as we develop the new version!

Cheers!

PR #1800, opened 2026-02-12
fix: empty lockscreen and media card background when wallpaper is disabled.

bigsaltyfishes:
# Pull Request

## Motivation
A fix for displaying wallpaper on lock screen and media card even if wallpaper is disabled. Otherwise will display nothing on lock screen(black background) and media card.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring

## Related Issue

## Testing

- [ ] Tested on niri
- [x] Tested on Hyprland
- [ ] Tested on sway

## Screenshots / Videos

## Checklist
- [x] Self-reviewed my code

### DIFF
--- a/Modules/Cards/MediaCard.qml
+++ b/Modules/Cards/MediaCard.qml
@@ -106,7 +106,7 @@ NBox {
       readonly property int dim: Math.round(256 * Style.uiScaleRatio)
       anchors.fill: parent
       visible: source.toString() !== ""
-      source: MediaService.trackArtUrl || (Settings.data.wallpaper.enabled && !Settings.data.wallpaper.useSolidColor ? root.cachedWallpaper : "")
+      source: MediaService.trackArtUrl || (!Settings.data.wallpaper.useSolidColor ? root.cachedWallpaper : "")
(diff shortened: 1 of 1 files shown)


ItsLemmy:
I don't see why we should show wallpaper when wallpaper is voluntarily disabled.

bigsaltyfishes:
> I don't see why we should show wallpaper when wallpaper is voluntarily disabled.

For example, I'm using linux-wallpaperengine on my system(also tested swaybg), so I need to disable wallpaper in noctalia to make it work, but if I disabled it, lock screen will show nothing(pure black background), and there is no way to change it. That's why we need to show wallpaper when wallpaper is voluntarily disabled, to provide a way show lock screen wallpaper.

ItsLemmy:
I see how this would help your use case, but this is not an acceptable work around .
BTW we  support video wallpapers via plugins.

bitbloxhub:
For your use-case, probably see https://github.com/noctalia-dev/noctalia-shell/issues/1641

PR #1812, opened 2026-02-13
SystemMonitor: Add option to pad the text

### LINKED ISSUE #1652: Feature Request: SystemMonitor: Prevent shifting without compact mode
### Feature Description
I would like to be able to set the following config for the SystemMonitor widget:

Whenever the monospaced font is enabled for the widget then there are options to make the SystemMonitor widget not shift.
This can be achieved by using the same width for text when they have a small value as well as a big value.
For example: CPU usage could be displayed as `"5%  "`, `"14% "` and `"100%"` padding the right with a space. (The space should be set to the right, because the icon of the value is to the left.)
Or the CPU usage could be displayed as `"005%"`, `"014%"` and `"100%"` padding the left with a zero (`0`).
Which padding is used could be decided by the user.

### Why Is This Needed?
I have difficulties with the compact mode, because I have it way easier to see a number and know what that means compared to a vertical beam and whenever the widget is shifting then I look up there automatically, because I register it as a huge change.

### Suggested Solutions
I would suggest to add a dropdown menu that only shows up when the monospaced font is enabled, that has the following entries:
- not padded
- padded with `0` to the left
- padded with ` ` to the right
- maybe even ` ` to the left to keep the `%`, `°`, or `KB` at the same spot

Alternatively the padding could be a toggle, when turned on the user could decide which character to use and which side to add padding; although just using one drop-down with good default choices could be less intimidating to configure and not clutter the settings too much.

### Additional Context
I would like to implement this feature.

liquidnya:
# Pull Request

This option is only shown if compact mode is off and a monospace font is enabled. If the bar is vertical the setting will be ignored. Everything should not shift and the width is calculated from the maximum value (which can be system dependent).

## Motivation

This PR adds the option to have the text values of the SystemMonitor padded with leading spaces.
This is done to prevent the layout to shift.
Some people have it easier to parse the number instead of the bars from the compact mode.
Having the numbers without the padding can lead to visual changes that can be distracting.
The setting is not possible if:

- compact mode is enabled (the setting is hidden) -> no text is shown
- the bar is a vertical bar (the setting is disabled) -> using leading padding to the left does not make sense when the icons and values are from top to bottom 
- the font is not monospace (the setting is disabled) -> the bar would start shifting if the letters do not have a fixed size

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring

## Related Issue
- Closes #1652

## Testing

- [x] Tested on niri
- [ ] Tested on Hyprland
- [ ] Tested on sway
- [x] Tested with different bar positions ~~and density settings~~

## Screenshots / Videos
Screenshot_2026-02-13T01-08-28

## Checklist
- [ ] Self-reviewed my code

## Additional Notes

I have not tested the behavior if nproc and/or loadAvg1 is 0.
I have only provided English and German translations, as those are the two languages I can speak.

This implementation uses the (probable) maximum value for how much a value needs to be padded.
Here are some explanations:

- `100%` is the maximum value of a scale that goes between `0%` and `100%`. Note that a previous version of noctalia-shell used `99%` as the maximum value which dropped the `%` for `100%` values which I think is distracting as well.
- `999°`C is hopefully the maximum value for the temperature of the CPU and GPU. I think it is realistic that it could go above 100°C still, but maybe `99°`C would be a reasonable maximum. If it gets higher than that maybe the bar shifting is a good thing. Otherwise the space will always be there. Also note that my CPU got really hot lately; I don't remember if it got to >= 100°C, but the noctalia-shell made me realise and so I removed all the dust.
- `9.9GHz` for the CPU frequency. I looked at this random website I found: https://valid.x86.fr/records.php
- The number of total CPUs is used for the maximum value of the avarage load. E.g. if someone has 20 CPUs, then the average load could reach up to 20.
- For the network speed I picked `9999G` as the maximum. Usually the text is between 4 (e.g. `20KB`) and 5 (e.g. ) in length.
- For the RAM and disk sizes the code is calculating the maximum value. In the case of RAM it is the total amount of RAM and in the case of a disk it is the total size of the disk.

Maybe it would be better to move the icons to the right side of the numbers when using that setting.
I noticed that when the widget is displayed vertically that the icons are also after the text (when reading from top to bottom).

### DIFF
--- a/Assets/Translations/de.json
+++ b/Assets/Translations/de.json
@@ -224,7 +224,11 @@
       "swap-usage-description": "Swap-Speichernutzung anzeigen.",
       "swap-usage-label": "Swap-Auslastung",
       "use-monospace-font-description": "Verwende eine Monospace-Schriftart für eine konsistente Zeichenbreite.",
-      "use-monospace-font-label": "Schriftart mit fester Breite"
+      "use-monospace-font-label": "Schriftart mit fester Breite",
+      "use-padding-description": "Füllt die Textwerte mit Leerzeichen auf. Verhindert Layoutverschiebungen.",
+      "use-padding-description-disabled-monospace-font": "Eine Monospace-Schriftart ist notwendig, um dieses Option wählen zu können.",
+      "use-padding-description-disabled-vertical": "Vertikale Taskleiste unterstützt keine Titelanzeige.",
+      "use-padding-label": "Text mit führenden Leerzeichen auffüllen"
     },
     "taskbar": {
       "colorize-icons-description": "Themenfarben auf Taskbar-Symbole anwenden.",
--- a/Assets/Translations/en.json
+++ b/Assets/Translations/en.json
@@ -224,7 +224,11 @@
       "swap-usage-description": "Show swap memory usage.",
       "swap-usage-label": "Swap usage",
       "use-monospace-font-description": "Use monospace font for consistent character width.",
-      "use-monospace-font-label": "Monospace font"
+      "use-monospace-font-label": "Monospace font",
+      "use-padding-description": "Pads the text values with leading spaces to prevent layout shifting.",
+      "use-padding-description-disabled-monospace-font": "A monospace font is needed for this feature.",
+      "use-padding-description-disabled-vertical": "Vertical taskbar does not support padding text.",
+      "use-padding-label": "Pad text"
     },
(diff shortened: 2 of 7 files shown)


ItsLemmy:
Very nice, thanks!

PR #1827, opened 2026-02-14
Add A new widget for theme color choice

tmarti2:
# Pull Request

## Motivation

Someone on discord recommended the plug-in [slowbongo](https://github.com/tuibird/slowbongo/tree/main/slowbongo) from @tuibird recently, and I liked the way she did the color choices. The current Noctalia way of doing this is via a combobox without any preview of the resulting color, which is not really user friendly.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Refactoring

## Testing

- [x] Tested on niri
- [ ] Tested on Hyprland
- [ ] Tested on sway

## Screenshot

image

## Checklist
- [x] Self-reviewed my code

## Additional Notes

I'm new to Qt/Qml and this is my first PR.

I'm not sure of what we want for default values for this component, and which property should be used to make sure it fits correctly in any situation ? My main concerns are `circleSize`, `diameter`, and the `RowLayout` properties like spacing, margins etc. I've tested with min/max scale and it seems to work, at least for the Bluetooth widget, but I've not tested in other situations.

Also I need to find a way to avoid scaling the icon on the selected choice if it is possible.

### DIFF
--- a/Widgets/NColorChoice.qml
+++ b/Widgets/NColorChoice.qml
@@ -19,6 +19,7 @@ RowLayout {
               "value": defaultValue === "" ? "(empty)" : String(defaultValue)
             });
   }
+  readonly property int diameter: circleSize * Style.uiScaleRatio
 
   signal selected(string key)
 
@@ -31,12 +32,7 @@ RowLayout {
 
   RowLayout {
     id: colourRow
-
-    property real hoverScale: 1.15
-    property int diameter: circleSize * Style.uiScaleRatio
-
-    Layout.margins: Style.uiScaleRatio * hoverScale
-    Layout.minimumWidth: Math.round(root.minimumWidth * Style.uiScaleRatio)
+    Layout.minimumWidth: root.diameter * Color.colorKeyModel.length
 
     Repeater {
       model: Color.colorKeyModel
@@ -48,12 +44,12 @@ RowLayout {
         property bool isHovered: circleMouseArea.containsMouse
 
         Layout.alignment: Qt.AlignHCenter
-        implicitWidth: colourRow.diameter
-        implicitHeight: colourRow.diameter
-        radius: colourRow.diameter * 0.5
+        implicitWidth: root.diameter
+        implicitHeight: root.diameter
+        radius: root.diameter * 0.5
         color: Color.resolveColorKey(modelData.key)
-        border.color: isSelected ? Color.mOnSurface : "transparent"
-        border.width: isSelected ? Style.borderS + 1 : 0
+        border.color: (isSelected || isHovered) ? Color.mOnSurface : Color.mOutline
+        border.width: Style.borderM
         scale: isHovered ? colourRow.hoverScale : 1
 
         MouseArea {
@@ -76,13 +72,6 @@ RowLayout {
           pointSize: Math.max(Style.fontSizeXS, colorCircle.width * 0.4)
           color: Color.mOnPrimary
           visible: colorCircle.isSelected
-        }
-
-        Behavior on scale {
-          NumberAnimation {
-            duration: Style.animationNormal
-            easing.type: Easing.OutCubic
-          }
         }
 
         Behavior on border.color {


tmarti2:
Also, I only did one replacement for now, but if you like this change and one NColorChoice (we can rename this btw) is ready, I can replace most occurrences of comboboxes used to make this choice.

Ly-sec:
Hey @tmarti2 , we really enjoy this :). Just _maybe_ the circles could be a tiny bit smaller? Not really sure to be honest.
We'd very much appreciate if we could factorize it as you mentioned (NColorChoice sounds good to me). 
So if you want, you're welcome to replace the other color pickers too :).

tmarti2:
To which point should I factorize ? I noticed that many Widgets set some value even if they are the default one, like `minimumWidth`.
Also I could make default label and description the one for color selection, so declaring a setting would look like
```qml
  NColorChoice {
    currentKey: valueTextColor
    onSelected: key => {
                  valueTextColor = key;
                  saveSettings();
                }
  }
```
Or
```qml
  NColorChoice {
    label: I18n.tr("common.select-icon-color")
    currentKey: valueIconColor
    onSelected: key => {
                  valueIconColor = key;
                  saveSettings();
                }
  }
```
etc

tmarti2:
Ok, I've reduced circle sizes to 90% of its previous size (in fact this was what was done by the mentionned plug-in).
I've also factorized as much as possible, including labels and descriptions, tell me if you think it's a good idea or not (the only drawback I can think of is that to understand what's this widget is for, you need to read the changed value instead of the label key).

I have a question regarding Bar->AppearanceTab->Capsule color : the color for `None` is different than all other widgets, and so the preview is wrong, but I'm not sure where this difference comes from

Ly-sec:
> To which point should I factorize ? I noticed that many Widgets set some value even if they are the default one, like `minimumWidth`. Also I could make default label and description the one for color selection, so declaring a setting would look like
> 
> ```qml
>   NColorChoice {
>     currentKey: valueTextColor
>     onSelected: key => {
>                   valueTextColor = key;
>                   saveSettings();
>                 }
>   }
> ```
> 
> Or
> 
> ```qml
>   NColorChoice {
>     label: I18n.tr("common.select-icon-color")
>     currentKey: valueIconColor
>     onSelected: key => {
>                   valueIconColor = key;
>                   saveSettings();
>                 }
>   }
> ```
> 
> etc

I think having the label in it makes sense given it would probably never change anyways.
So far I think we're good when it comes to factorizing it, just a matter of getting all the widgets etc to use the new color swatches, which it seems like you already did (not sure if I missed any).

tmarti2:
> I think having the label in it makes sense given it would probably never change anyways. So far I think we're good when it comes to factorizing it, just a matter of getting all the widgets etc to use the new color swatches, which it seems like you already did (not sure if I missed any).

Well, I changed the default label but I can revert this to keep label/description everywhere if you think its better.

The only remaining mention of `Color.colorKeyModel` is in `NColorChoice`, so I think we're probably good (?)

Ly-sec:
> > I think having the label in it makes sense given it would probably never change anyways. So far I think we're good when it comes to factorizing it, just a matter of getting all the widgets etc to use the new color swatches, which it seems like you already did (not sure if I missed any).
> 
> Well, I changed the default label but I can revert this to keep label/description everywhere if you think its better.
> 
> The only remaining mention of `Color.colorKeyModel` is in `NColorChoice`, so I think we're probably good (?)

Nah it's okay, I think we keep it as is :).
Sounds like we are good then, anything you want to do before merging it?

tmarti2:
> Sounds like we are good then, anything you want to do before merging it?

Well, appart from the fact that I'm not confident that I did the right thing when styling the widget (it works in settings but does it work in other contexts ?), and the wrong preview for `Bar->AppearanceTab->Capsule color` that I mentioned above, I think we're good.

For the latter, the `Style.qml:142`, where we change the color in case of `none`. I'm not sure how I could fix that, maybe a public property to say which color to use for the `none` preview (which would stay `Color.resolveColorKey("none")` by default but could be switched to anything ?

Ly-sec:
> it works in settings but does it work in other contexts ?

What other contexts do you mean? Maybe I'm missing something here.

> which would stay Color.resolveColorKey("none") by default but could be switched to anything ?

That does sound like a reasonable approach, what do you think @ItsLemmy ?

tmarti2:
> What other contexts do you mean? Maybe I'm missing something here.

No idea tbh ahah, since the widget is available for anyone who want to use it, maybe there are some use cases I didn't think of.

One more question, is it possible to have a tooltip on such rectangles ? Someone suggested to have a tool tip to show "Default", "Primary", etc. What do you think (if it is possible) ?

tmarti2:
Not that easy lol (without a border the color is the same than the background)
image

Ly-sec:
> One more question, is it possible to have a tooltip on such rectangles ? Someone suggested to have a tool tip to show "Default", "Primary", etc. What do you think (if it is possible) ?

You could just add a tooltip via `onEntered` in the Mousearea perhaps?

tmarti2:
I added the tooltip !

I cannot find a generic way to make border color work in cases where the circle color is too similar to the panel color

Ly-sec:
> I cannot find a generic way to make border color work in cases where the circle color is too similar to the panel color

I think that's fine for now, we can always take a look afterwards :).

ItsLemmy:
Style.qml:142 does not need to change it's fine as it is.

Circles are too big (38pixels height) when not selected. In QML borders are on top of things and not around, so when you set the border to "transparent", it kinda make the circle bigger.  I simply recommend 

```
        border.color: isSelected ? Color.mOnSurface : Color.mOutline
        border.width: Style.borderM
```

image


The rest is good to go

ItsLemmy:
Also not a big fan of the hover scale effect, as it zoom stuff up it will clip to the right unless you reserve some space for it. Which then means everything is not aligned on idle.

I find the widget looks better like this.

``` 
import QtQuick
import QtQuick.Layouts
import qs.Commons
import qs.Services.UI

RowLayout {
  id: root

  property real minimumWidth: 200
  property string label: I18n.tr("common.select-color")
  property string description: I18n.tr("common.select-color-description")
  property string tooltip: ""
  property string currentKey: ""
  property var defaultValue: "none"
  property int circleSize: Style.baseWidgetSize * 0.9
  readonly property bool isValueChanged: currentKey !== defaultValue
  readonly property string indicatorTooltip: {
    I18n.tr("panels.indicator.default-value", {
              "value": defaultValue === "" ? "(empty)" : String(defaultValue)
            });
  }
  readonly property int diameter: circleSize * Style.uiScaleRatio

  signal selected(string key)

  NLabel {
    label: root.label
    description: root.description
    showIndicator: root.isValueChanged
    indicatorTooltip: root.indicatorTooltip
  }

  RowLayout {
    id: colourRow
    Layout.minimumWidth: root.diameter * Color.colorKeyModel.length

    Repeater {
      model: Color.colorKeyModel

      Rectangle {
        id: colorCircle

        property bool isSelected: root.currentKey === modelData.key
        property bool isHovered: circleMouseArea.containsMouse

        Layout.alignment: Qt.AlignHCenter
        implicitWidth: root.diameter
        implicitHeight: root.diameter
        radius: root.diameter * 0.5
        color: Color.resolveColorKey(modelData.key)
        border.color: (isSelected || isHovered) ? Color.mOnSurface : Color.mOutline
        border.width: Style.borderM
        scale: isHovered ? colourRow.hoverScale : 1

        MouseArea {
          id: circleMouseArea

          anchors.fill: parent
          hoverEnabled: true
          cursorShape: Qt.PointingHandCursor
          onEntered: TooltipService.show(parent, modelData.name)
          onExited: TooltipService.hide()
          onClicked: {
            root.currentKey = modelData.key;
            root.selected(modelData.key);
          }
        }

        NIcon {
          anchors.centerIn: parent
          icon: "check"
          pointSize: Math.max(Style.fontSizeXS, colorCircle.width * 0.4)
          color: Color.mOnPrimary
          visible: colorCircle.isSelected
        }

        Behavior on border.color {
          ColorAnimation {
            duration: Style.animationFast
          }
        }
      }
    }
  }
}
```

image

tmarti2:
What about the special `AppearanceTab` case ? The previewed color does not match the effective one

tmarti2:
I applied your suggestions, it's far less complicated this way, thanks

Edit: Just saw @Ly-sec message, sorry. So we can keep it this way. I guess the MR is ready then !

ItsLemmy:
Nice, thanks a lot, merging!  I'll touch up a few details (the "none" case which may vary depending of context)

tmarti2:
> Nice, thanks a lot, merging! I'll touch up a few details (the "none" case which may vary depending of context)

I had a non-generic/hackish way to fix this particular case for "none", but it's not generic in a sense that anyone can create a color choice but ignore one of them and replace it by another one when displayed. At the moment there is only "none" in capsule case but who knows what will happen in the future 😬 

Anyway, thanks for the merge !

PR #1847, opened 2026-02-15
Lockscreen password animation and icons

tmarti2:
# Pull Request

## Motivation

Before Noctalia I tried iNiR which used (end-4) custom shapes to replace the usual `circle-filled` and I thought it was really cute, so here is a attempt to reproduce it without importing hundreds of JS lines.

This MR:
1. Add animations to password icons (which can be disabled like all animations)
2. Change the style a bit to make it look better (I think)
3. Add a setting to use cute icons in addition to the circle one. I selected some that I thought looked nice but we can remove/add/replace if you think it is pertinent. The icon order is not random so better configurations can be find.

To be honest the most important change to make it look better is 1. and 2., 3 is debatable.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring

## Testing

- [x] Tested on niri
- [ ] Tested on Hyprland
- [ ] Tested on sway

## Screenshots
I cannot use screenshots in the lockscreen :grimacing: 

## Checklist
- [x] Self-reviewed my code

### DIFF
--- a/Assets/Translations/en.json
+++ b/Assets/Translations/en.json
@@ -1280,6 +1280,8 @@
       "lock-screen-tint-strength-description": "Applies a tint overlay to the lock screen wallpaper.",
       "lock-screen-tint-strength-label": "Lock screen tint strength",
       "monitors-desc": "Show lock screen on specific monitors. Defaults to all if none are chosen.",
+      "password-chars-description": "Cute icons used to hide your password.",
+      "password-chars-label": "Hidden password icons",
       "show-hibernate-description": "Show the option 'hibernate' in the power controls.",
       "show-hibernate-label": "Show hibernate",
       "show-session-buttons-description": "Allow access to power settings from the lock screen.",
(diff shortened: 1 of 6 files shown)


tmarti2:
Also I have a question regarding translation, I did the english, version I can do the french version, but what about the other languages ?

Ly-sec:
Hey @tmarti2 , while I appreciate your PR I don't think there is any need to have a toggle for this and also you should be providing a video for it - which shouldn't be 2 hard :).


In terms of translations: you only need to translate english, the rest will be done by us / the people that got access to translations.

tmarti2:
Here is a short video with and without the icons.
Animations can be disabled via the lockscreen setting to disable them or the global one.

https://github.com/user-attachments/assets/93226582-0226-4d0a-862b-f21cdf2df764

Ly-sec:
> Here is a short video with and without the icons. Animations can be disabled via the lockscreen setting to disable them or the global one.
> lock.mp4

Looks good to me, going to merge this :heart: thank you very much!

PR #1851, opened 2026-02-16
home-module: coerce restart triggers to strings for path-backed config

contagnas:
This fixes a Home Manager type error when `programs.noctalia-shell.systemd.enable = true` is used with settings, colors, plugins, user-templates, or pluginSettings provided as file paths.

X-Restart-Triggers in HM [expects entries of type package or string](https://github.com/nix-community/home-manager/blob/175532b6275b34598a0ceb1aef4b9b4006dd4073/modules/systemd.nix#L151), but the module was passing raw path values from `config.xdg.configFile.*.source`.

### DIFF
--- a/nix/home-module.nix
+++ b/nix/home-module.nix
@@ -183,12 +183,8 @@ in
 
   config = lib.mkIf cfg.enable {
     systemd.user.services.noctalia-shell = lib.mkIf cfg.systemd.enable {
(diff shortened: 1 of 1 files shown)


bokicoder:
I need to verify the issue you described, but one thing is clear: your Nix code is wrong.

PR #1854, opened 2026-02-16
notification: fix action invoke when animations are disabled

yuzujr:
# Pull Request

## Motivation
Fix a regression in `v4.4.1` where notification actions fail to trigger when interface animations are disabled.

Root cause:
- Commit `330df2b6e` changed action handling in `Modules/Notification/Notification.qml` from immediate invoke to a deferred timer (`deferredActionTimer`).
- At the same time, card removal uses `removalTimer` with `interval: Style.animationSlow`.
- When animations are disabled, `Style.animationSlow` becomes `0`, so the notification delegate is removed immediately.
- Because `deferredActionTimer` lives inside that delegate, it gets destroyed before firing, and action callbacks never run.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring

## Related Issue
- N/A

## Testing
Manual verification focus:
- Repro condition: only happens with `Settings.data.general.animationDisabled = true`.
- With animations enabled, existing behavior remains unchanged.

- [x] Tested on niri
- [ ] Tested on Hyprland
- [ ] Tested on sway

## Screenshots / Videos
N/A

## Checklist
- [x] Self-reviewed my code
- [x] No new warnings or errors from startup run
- [x] Documentation updated

## Additional Notes
Implementation details:
- Added `card.runDeferredAction(actionId, isHistoryRemoval)` to centralize post-click behavior.
- If `Style.animationSlow <= 0`, action/history removal is executed immediately, then card dismissal runs.
- If animations are active, keep the previous UX intent (dismiss animation first, then invoke action shortly after), but compute timer delay defensively to stay before removal timing.

Files changed:
- `Modules/Notification/Notification.qml`

### DIFF
--- a/Modules/Notification/Notification.qml
+++ b/Modules/Notification/Notification.qml
@@ -402,9 +402,7 @@ Variants {
                               return a.identifier === "default";
                             });
                             if (hasDefault) {
-                              card.animateOut();
-                              deferredActionTimer.actionId = "default";
-                              deferredActionTimer.start();
+                              card.runDeferredAction("default", false);
                             }
                           }
               onCanceled: {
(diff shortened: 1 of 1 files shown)


Ly-sec:
Hey @yuzujr , thank you very much for the PR! It would've went unnoticed for much longer I think, most people probably don't turn off animations.
Merging it :heart:

PR #1855, opened 2026-02-16
feat: safe markdown in notification history

tibssy:
## Summary
- add opt-in markdown rendering for notification history entries
- sanitize markdown content and allowed links before display
- add explicit expand/collapse control so links can be clicked safely

## Motivation
Notifications often contain lightweight formatting (bold, italics, links). This change lets history entries render that formatting in a controlled way, while sanitizing untrusted content so it can’t inject HTML or unsafe links. The result is clearer, more readable history without sacrificing safety or accidental activation of unsafe URLs.

### DIFF
--- a/Assets/Translations/en.json
+++ b/Assets/Translations/en.json
@@ -599,6 +599,7 @@
   "notifications": {
     "panel": {
       "click-to-expand": "Click to expand",
+      "click-to-collapse": "Click to collapse",
       "description": "Your notifications will show up here as they arrive.",
       "no-notifications": "No notifications"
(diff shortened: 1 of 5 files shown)


Ly-sec:
Hey @tibssy maybe I'm misunderstaind something here but wasn't part of the goal for this PR to properly render markdown in the notification history?
I just tested it and it did in fact not render markdown (yes, I activated the toggle).

image

tibssy:
hi @Ly-sec I just reopen the pr on the correct way.

PR #1876, opened 2026-02-18
Adjust margins in BluetoothPanel & fix flickering issue in BluetoothSubTab

notiant:
1. Adjusting the margins in the Bluetooth panel to make them the same as in thw Wifi panel.
2. Solving a flickering issue in the Bluetooth settings tab that was reported on Discord (https://discord.com/channels/1401598189823590460/1473323041940439251)

### DIFF
--- a/Modules/Panels/Network/WiFiNetworksList.qml
+++ b/Modules/Panels/Network/WiFiNetworksList.qml
@@ -154,8 +154,8 @@ NBox {
                   visible: modelData.connected && NetworkService.disconnectingFrom !== modelData.ssid
                   color: NetworkService.internetConnectivity ? Color.mPrimary : Color.mError
                   radius: height * 0.5
-                  width: connectedText.implicitWidth + Style.marginS * 2
-                  height: connectedText.implicitHeight + Style.marginXS
+                  width: Math.round(connectedText.implicitWidth + Style.marginS * 2)
+                  height: Math.round(connectedText.implicitHeight + Style.marginXS)
 
                   NText {
                     id: connectedText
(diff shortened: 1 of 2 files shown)


tmarti2:
Thanks for the PR !

It still flickers:
- With "separate window", when I close the widget panel (by clicking on the main setting panel for instance), but it flickers only once
- Without "separate window" when the setting panel opens, and it stop fast too since this closes the widget panel

ItsLemmy:
dont assume marginXL is twice marginM, it's bad practice.
why getting rid of the rounding?

turannul:
I told already?

**Flickering is not A BUG**.

However i understand might be annoying, but state changes when panel opens then something else is changing it again.

Something in the system causes that behavior...

notiant:
@ItsLemmy I'm mostly trying to be consistent to how it's handled in other panels and Math.round() is not widely used since it's seems unnecessary in most cases because implicitHeight is often already an integer (this is not influenced by fractional or UI scaling). I've added it back in some places where implicitWidth actually gives a real number.
For the margins, `Style.marginM * 2` is currently used nowhere in the code due to this [PR](https://github.com/noctalia-dev/noctalia-shell/pull/1408/changes), so if you think this is bad style all these changes should be reverted.


@tmarti2 The issues you mentioned should now also be fixed.

ItsLemmy:
Oh I see, well I have never seen this PR before, IMO should not have not been merged like that, as it makes the code much harder to maintain and understand in particular for new contributors. 

I'll merge your and go and fix all that. By creating proper new properties. e.g. Style.margin2M

Issue #1882, opened 2026-02-18
[Bug]  Pop-up notification actions don't always work


AlexNabokikh:
### Description

Just like the title says. When you try to click on the action in the notification pop-up, nothing happens. However, if you try to open the notifications history menu and click on the action there, it works as expected.

### Steps to Reproduce
1. Fire up some actionable notifications. 
2. Try to click on the action button.
3. It will not work for some cases. I tested Telegram (open action) and Brave (activate, settings), Gnome Pomodoro (take a break, lengthen).

### Expected Behavior

Actions in pop-up notifications should work just like in the notification history.
Additionally, I would appreciate it if Noctalia could switch focus to the application that sent the notification when you click on the pop-up. This behavior is common across all DEs regardless of platform (Windows, Mac, Linux).

### Video

https://github.com/user-attachments/assets/8ad4638a-ac47-4468-a8de-735ba78e15d5

### Environment

OS: NixOS 26.05 (Yarara) x86_64
Kernel: Linux 6.19.0
Compositor: Niri 25.11 (Wayland)

Ly-sec:
So I guess we can close this one, correct @AlexNabokikh :p?

AlexNabokikh:
@Ly-sec yes. Sorry, I thought it would be closed automatically.

PR #1883, opened 2026-02-18
fix(#1882): resolve notification popup actions failing to invoke

AlexNabokikh:
## Motivation

Notification action buttons (Reply, Open, etc.) fail silently when clicked in popup toasts, while the exact same actions work correctly from the history panel. 
The root cause is a delegate lifecycle race: the popup used `deferredActionTimer` (50ms) to invoke actions after starting the dismiss animation, but `removalTimer` could destroy the entire delegate (and the timer with it) via `activeList.remove()` before the action ever fired.

Apart from that, this PR also adds focus-on-click behavior: clicking a notification body now activates the sender application's window, matching standard desktop environment behavior.

## Type of Change

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring

## Related Issue
- https://github.com/noctalia-dev/noctalia-shell/issues/1882

## Testing
- [x] Tested on niri
- [ ] Tested on Hyprland
- [ ] Tested on sway

## Videos

https://github.com/user-attachments/assets/7165ec77-9ed5-4b6d-bd8f-0fbe6e69c7c8

## Checklist

- [x] Self-reviewed my code

## Additional Notes

**How to verify:**
1. Run `Scripts/dev/notifications-test.sh` for smoke test
2. Send notifications from real apps (Telegram, Brave, GNOME Pomodoro)
3. Click action buttons in popups - they should now work reliably
4. Click notification body --sender app window should come to front
5. Restart the shell, open history - action buttons should still appear

### DIFF
--- a/Modules/Notification/Notification.qml
+++ b/Modules/Notification/Notification.qml
@@ -502,6 +502,7 @@ Variants {
 
             function runAction(actionId, isDismissed) {
               if (!isDismissed) {
+                NotificationService.focusSenderWindow(model.appName);
                 NotificationService.invokeActionAndSuppressClose(notificationId, actionId);
               } else if (Settings.data.notifications.clearDismissed) {
                 NotificationService.removeFromHistory(notificationId);


Ly-sec:
Hey @AlexNabokikh while this works, it makes it so that the animation isn't played anymore after invoking an action - the notification just disappears instead of slides out.

AlexNabokikh:
@Ly-sec Ah, I disable animations, so I completely overlooked it. Let me fix it.

AlexNabokikh:
@Ly-sec actually, it works correctly on my side after I enabled animations. How can I reproduce your case?

https://github.com/user-attachments/assets/cf6b3ff1-bb6a-4224-9e4a-78d823251b36

AlexNabokikh:
@Ly-sec can you please check again? I think I figured it

Ly-sec:
> @Ly-sec can you please check again? I think I figured it

At least on my end it's still not fixed.
Reproduction Steps:
1: create notification with actions
2: click on action
3: notification instantly disappears.

What should happen
3: animate after clicking on an action

AlexNabokikh:
@Ly-sec thanks! That helped a lot. Can you please test it again?

Ly-sec:
@AlexNabokikh perfect, that works very well now! Going to merge it now, thank you very much :heart:

PR #1884, opened 2026-02-18
Improve default/reset slider

tmarti2:
# Pull Request

## Motivation

At start I just wanted to add reset button to few options which missed it, and I ended creating a new setting widget to do that more easily. I deleted labels for the reset button to generate automatically, to avoid having to create new labels each time we want a button, so I also removed the corresponding keys (Is it a breaking change ?).

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Refactoring

## Testing

- [x] Tested on niri
- [ ] Tested on Hyprland
- [ ] Tested on sway

## Checklist
- [x] Self-reviewed my code


ItsLemmy:
Thanks for the PR. I agree with the motivation behind this change; it is something I’ve wanted to address as well. However, there are a few items to resolve before merging:

- **Component Reuse:** We don't need a new `Nxxx` component. Instead, please add a `showReset` property to the existing `NValueSlider` (defaulting to `true`). This approach is more efficient regarding memory and maintenance, and it reduces the number of file changes.

- **Internationalization (i18n):** Constructing sentences programmatically creates syntax issues for other languages.
  ```javascript
  I18n.tr("common.reset") + " " + root.label.toLowerCase()
  ```
  This assumes English grammar rules (Subject-Verb-Object), whereas other languages might structure this phrase differently (e.g., "Scaling reset" vs. "Reset scaling").

- **Simplification:** Since `NValueSlider` always displays a label, the context is already clear. We can simply use `common.reset` for the tooltip. This also allows us to clean up the extra keys.

- **Translation Workflow:** Please avoid editing the non-English JSON files directly, as our translation website is the single source of truth and manual edits cause merge conflicts. Instead, please list the keys that need to be deleted in the PR description, and Lysec or I will handle the database cleanup after the merge.

tmarti2:
> Component Reuse: We don't need a new Nxxx component. Instead, please add a showReset property to the existing NValueSlider (defaulting to true). This approach is more efficient regarding memory and maintenance, and it reduces the number of file changes.

I've done that before but I was afraid it could cause issues in some cases, like sliders in bbar widgets etc, but if you think it's ok I'll do that instead. 

I'll simplify the tooltip as you said and revert translation changes !

Thanks for the review

tmarti2:
Ok should be good

PR #1894, opened 2026-02-19
Small adjustments to GTK theme

notiant:
Just some small adjustments to the GTK theme files since the colors for --accent-color and --view-bg-color differed from the other definitions. I also restored the behavior that the sidebar takes the main window color if the window is not focused.


Ly-sec:
Thank you very much @notiant ! Going to merge it. Before I do though, is there anything else you wanted to change or is it good to go?

notiant:
> Thank you very much @notiant ! Going to merge it. Before I do though, is there anything else you wanted to change or is it good to go?

No, thats all :)

Issue #1771, opened 2026-02-10
[Bug] Clipboard history provider slow startup


AlexNabokikh:
### Description

After updating Noctalia shell to 4.4.1-git (ebd7387) the clipboard history provider doesn't start until you trigger >clip in the launcher and wait for ~5 sec. (see the video attached).

https://github.com/user-attachments/assets/f53d4225-240c-4228-8d9c-0e6aa645a468

### Steps to Reproduce
1. Restart Noctalia.
2. Trigger clipboard history via `noctalia-shell ipc call launcher clipboard` or other means.
3. See the sign `Clipboard history disabled/ Enable clipboard history in settings ог install cliphist` that will disapear in about 5 seconds. Then clipboard history works normally. 
3a. Note that history doesn't work until you trigger clipboard history. Everything that you copied to the clipboard before won't get to the clipboard history. 

### Expected Behavior

The provider should start immediately with the noctalia-shell as it worked before.

### Environment
OS: NixOS 26.05 (Yarara) x86_64
Kernel: Linux 6.18.9
Compositor: Niri 25.11 (Wayland)

ItsLemmy:
can you update to the latest commit, there is one commit regarding clipboard that you dont have.

AlexNabokikh:
@ItsLemmy bumping to the latest commit has resolved the issue. Thank you!

AlexNabokikh:
It seems like there was a regression somewhere in recent releases. The clipboard history doesn't work until you trigger it. See the video below:

https://github.com/user-attachments/assets/4b3123f0-5d49-4073-9047-b8502db153a4

PR #1909, opened 2026-02-20
fix(1771): start clipboard watchers at shell boot instead of first la…

AlexNabokikh:
## Motivation

After a fresh shell start, the clipboard history doesn't capture any entries until the user opens the clipboard UI for the first time. This happens because `ClipboardService` is a lazily-instantiated `pragma Singleton` whose only references live inside `ClipboardProvider`, which is only loaded when the launcher panel opens. As a result, the `wl-paste --watch cliphist store` watchers never start until that first open, and all clipboard copies before that point are silently lost.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring

## Related Issue
- Closes https://github.com/noctalia-dev/noctalia-shell/issues/1771

## Testing

- [x] Tested on niri
- [ ] Tested on Hyprland
- [ ] Tested on sway

## Checklist
- [x] Self-reviewed my code

## Additional Notes

### DIFF
--- a/shell.qml
+++ b/shell.qml
@@ -30,6 +30,7 @@ import qs.Modules.Panels.Settings
 import qs.Modules.Toast
 import qs.Services.Control
 import qs.Services.Hardware
+import qs.Services.Keyboard
 import qs.Services.Location
 import qs.Services.Networking
 import qs.Services.Noctalia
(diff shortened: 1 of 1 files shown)


Ly-sec:
Looks good to me, going to merge it. Thank you very much :heart:

PR #1924, opened 2026-02-22
Fix Grid SessionMenu right/down press when nothing is selected

tmarti2:
# Pull Request

## Motivation

In case of no selection, we consider we're on the 0,0 index of the grid. but this causes the first Right/Down press to actually skip the first cell. This MR fixes this by staying on `0` when a Right/Down key is pressed on a negative index value.

I also fixed the behavior when trying to display an empty menu: We close the panel if the menu is empty. I'm not sure this is the best solution tho, makes the screen flicker for me.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring

## Testing

- [x] Tested on niri
- [ ] Tested on Hyprland
- [ ] Tested on sway

## Checklist
- [x] Self-reviewed my code

## Additional Notes

### DIFF
--- a/Modules/Panels/SessionMenu/SessionMenu.qml
+++ b/Modules/Panels/SessionMenu/SessionMenu.qml
@@ -327,13 +327,15 @@ SmartPanel {
       newCol = newCol - 1 < 0 ? grid.itemsInRow(newRow) - 1 : newCol - 1;
       break;
     case "right":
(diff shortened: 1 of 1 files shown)


Ly-sec:
Hey @tmarti2 thank you very much for the PR. I'm unsure about the close when no session menu entries are available but I guess that would make sense indeed. I'll merge it as is for now, maybe we'll find a better approach for that later.

PR #1929, opened 2026-02-22
Fix border clipping for tooltips and dock

notiant:
I finally found a way to fix the clipping of borders that can appear when fractional scaling is enabled.

For the tooltips, I added `anchors.margins: border.width / 2` to ensure that the border is always positioned fully within the tooltip. I also reverted some changes from a previous attempt to fix this problem.

For the dock, I've just added a safety margin of 2px. Further, I re-implemented the larger (horizontal) dockContainerWrapper for tooltips that caused some glitches the last time it was added. I've found that they are related to `layer.enabled` and that moving this to dockContainerWrapper prevents the glitches.

Tested on Hyprland, not on Niri.


Ly-sec:
Hey @notiant , just tested it on niri and it does seem to work there too. Going to merge it, thank you very much! Glad someone finally figured it out :heart:

PR #1933, opened 2026-02-22
DockMenu: layout adjustments

notiant:
This PR makes some layout corrections for the DockMenu, mainly ensuring even margins and fixing a hover text color problem. I also increased the padding for the dock peek area a bit since it's otherwise hard to hit when only a few apps are in the dock (only if there is no frame indicator).

### DIFF
--- a/Modules/Dock/Dock.qml
+++ b/Modules/Dock/Dock.qml
@@ -102,7 +102,7 @@ Loader {
       readonly property int barHeight: Style.getBarHeightForScreen(modelData?.name)
       readonly property int peekEdgeLength: {
         const edgeSize = isVertical ? Math.round(modelData?.height || maxHeight) : Math.round(modelData?.width || maxWidth);
-        const minLength = Math.max(1, Math.round(edgeSize * 0.1));
+        const minLength = Math.max(1, Math.round(edgeSize * ((isStaticMode && Settings.data.dock.showFrameIndicator && Settings.data.bar.barType === "framed" && hasBar) ? 0.1 : 0.25)));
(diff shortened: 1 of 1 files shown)


tibssy:
Hey @notiant! Making the peek area easier to hit when only a few apps are open is a great idea! However, I wanted to point out a small issue: modifying frameIndicatorLength won't actually achieve what you're aiming for.

```
const padded = layoutLength + (showFrameIndicator ? Style.marginXL : Style.margin2XL * 6);
```

If you want to increase the size of the trigger area when only a few apps are open and showFrameIndicator is false, you should modify the minLength inside peekEdgeLength. Instead of adding fixed padding. You can simply increase the minimum percentage of the screen edge for example, changing edgeSize * 0.1 to edgeSize * 0.2 for a 20% minimum hit-box. Here is an example solution:

```
      readonly property int peekEdgeLength: {
        const edgeSize = isVertical ? Math.round(modelData?.height || maxHeight) : Math.round(modelData?.width || maxWidth);
        const minLength = Math.max(1, Math.round(edgeSize * (Settings.data.dock.showFrameIndicator ? 0.1: 0.2)));
        return Math.max(minLength, frameIndicatorLength);
      }
```

notiant:
I have changed the peek area length using your suggestion, although I extended the if condition to more than `Settings.data.dock.showFrameIndicator`.

Ly-sec:
Thanks for the PR @notiant , are we good to merge it then :)?

tibssy:
Thanks @notiant, @Ly-sec. It looks good to me.

PR #1937, opened 2026-02-23
fix(Launcher): Allow item images to take advantage of preferredGridCellRatio

krendil:
# Pull Request

This gives the launcher item preview image/icons the same ratio as the overall grid cell, letting the provider display non-square images.

## Motivation

The motivation for this comes from the calibre-provider plugin, which allows users to search their calibre library from the launcher. The ebook cover images are used as preview images in the launcher. Most ebook cover images have an aspect ratio of 1.5 (non-Kindle) or 1.6 (Kindle). Currently, the cover images are cropped to square, but I'd like to be able to show the entire cover in the launcher.

Before:
Screenshot from 2026-02-23 13-20-19

After:
Screenshot from 2026-02-23 13-20-01

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring

I've marked this as a non-breaking bug-fix, although both are arguable. I'm not sure if the current behaviour of not scaling the image size is intentional or not, but it seems like the image ratio should follow the grid cell ratio.

The only other reference to `preferredGridCellRatio` I could find in noctalia-shell and noctalia-plugins is in the official Emoji provider, which sets a ratio of 1.15 to allow more space for the label. It looks like the change has caused the label to shift down slightly, as the image is now taking up 65% of that extra space. I don't know if this is considered 'breaking' or not.

Before:

Screenshot from 2026-02-23 13-32-25

After:

Screenshot from 2026-02-23 13-32-09

It might be the case that the entire reason for adding `preferredGridCellRatio` was to allow a bit of extra space for the label of the Emoji picker and this PR ruins that. If that is the case, I think it should be possible to come up with a better solution for the problem of not having enough vertical space at small grid view sizes. E.g, one could calculate an absolute minimum grid cell height based on current text and padding sizes, regardless of the number of grid columns. I'll post the details of how one might go about this in a comment, as it's not really related to the main PR.

## Related Issue
- None

## Testing

- [x] Tested on niri
- [ ] Tested on Hyprland
- [ ] Tested on sway
- [x] Tested with different launcher density settings

## Screenshots / Videos
See above

## Checklist
- [x] Self-reviewed my code

## Additional Notes


krendil:
If we wanted to automatically calculate a minimum height for the grid cell to make sure we have enough space to show the label, this is how we could do it:

If we let $H$ = the height of the grid cell, $M$ = the amount of space taken up by margins and spacing, and $L$ = the amount of space taken up by the label and $E$ = the amount of extra space, then we have the equation:
$H = 0.65H + M + L + E$

- $M$ is either $4 + 4 + 0 = 8$ or $6  + 6 + 1 = 13$ depending on whether compact mode is enabled or not.
- $L$ is $10pt$, $10$ based on a `baseSize` value of `Style.fontSizeS`. (Note we will need to convert from points to pixels). (Note that sometimes `hideLabel` is true, but we can't guarantee that it will be true for every item)
- We want to ensure that $E >= 0$

Solving first for E, we get $E = 0.65H - M - L$.

Then setting both sides $>= 0$ and solving for $H$, we get $H >= (M + L)/0.35$, or

```javascript
const marginSize = (root.isCompactDensity ? (Style.marginXS * 2) : (Style.marginS * 2 + Style.marginXXS)) * Style.uiScaleRatio;
const labelSize = Style.fontSizeS * (4/3) * Screen.devicePixelRatio * Style.uiScaleRatio; // Not sure if this is the best way to convert to px
const minCellHeight = (marginSize + labelSize) / 0.35;
```

and use that value as a minimum when setting `resultsGrid.cellHeight` in LauncherCore.qml.

If that approach is taken, then this PR would need to be updated to use `gridEntry.width * 0.65 * preferredGridCellRatio` instead, otherwise it will eat the extra space again.

ItsLemmy:
thanks!

PR #1941, opened 2026-02-23
Fix notification hovering not working on content/close button

tmarti2:
# Pull Request

## Motivation
The notification timer to dismiss (and bar animation) was not stopped when hovering content or close button

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Refactoring

## Related Issue

## Testing

- [x] Tested on niri
- [ ] Tested on Hyprland
- [ ] Tested on sway

## Screenshots / Videos

## Checklist
- [x] Self-reviewed my code

## Additional Notes
I used `HoverHandler` to avoid having to define both entered/exited. maybe this could be generalized ?
I still don't understand why we have a counter instead of a boolean, I tried to but a global handler but for some reason I loose focus when there are more than one notification, and I don't understand why yet.

### DIFF
--- a/Modules/Notification/Notification.qml
+++ b/Modules/Notification/Notification.qml
@@ -198,7 +198,7 @@ Variants {
 
             property string notificationId: model.id
             property var notificationData: model
-            property int hoverCount: 0
+            property bool isHovered: false
             property bool isRemoving: false
 
             readonly property int animationDelay: index * 100
@@ -249,34 +249,13 @@ Variants {
               return deltaY;
             }
 
-            // Hover handling
-            onHoverCountChanged: {
-              if (hoverCount > 0) {
-                resumeTimer.stop();
-                NotificationService.pauseTimeout(notificationId);
-              } else {
-                resumeTimer.start();
-              }
-            }
-
-            Timer {
-              id: resumeTimer
-              interval: 50
-              repeat: false
-              onTriggered: {
-                if (hoverCount === 0) {
-                  NotificationService.resumeTimeout(notificationId);
-                }
-              }
-            }
-
             // Animation setup
             function triggerEntryAnimation() {
               animInDelayTimer.stop();
               removalTimer.stop();
               resumeTimer.stop();
               isRemoving = false;
-              hoverCount = 0;
+              isHovered = false;
               isSwiping = false;
               swipeOffset = 0;
               swipeOffsetY = 0;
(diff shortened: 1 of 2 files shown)


tmarti2:
> I still don't understand why we have a counter instead of a boolean

(I understand now)

Ly-sec:
Hey @tmarti2 so I assume we're good to go then :) ? Also thank you very much for the PR!

tmarti2:
Yes, I'll do a separate PR if I ever understand better how this work, since I'm currently working on something different which also requires hover information

tmarti2:
Nvm can you wait a bit ? I'm trying something

tmarti2:
@Ly-sec Ok I need to explain the last two commit so I can get your opinion.

I wanted to use an `HoverHandler` to manage the hovering status of notifications, which seemed a better way, because managing hovering with entered/exited in many sub components  is really error-prone (This issue started as a fix caused by this). But It was harder than I thought : Notification items are way bigger that the actual visible size, due mainly to `shadowPadding`. Because of this, spacing inside `notificationStack` is negative, which causes an overlapping of items. The handler could not be attached to the whole item, so I had to create a sub item which includes all content and `MouseArea` to the right size (`cardBackground` one).

I've tested and did not find any bug, so I think it works, but it required moving a lot of code.

Edit: You can see how the `HoverHandler` work in `Toast.qml`, which basically does the same thing than `Notification.qml` but without the overlapping problem (since there is only one toast at a time).
Edit 2: I rebased this into two separate commits to make reviewing easier.

Note: I actually need a more centralized way to handle this for another thing I'm working on.

Ly-sec:
@tmarti2 I wonder if it wouldn't make more sense to have a "toggle" to stop timers when we hover over a button instead of having to replicate the same logic over and over again - factorizing it. Something like `hoverTimer: bool` instead of having it added for however many situations.

tmarti2:
> @tmarti2 I wonder if it wouldn't make more sense to have a "toggle" to stop timers when we hover over a button instead of having to replicate the same logic over and over again - factorizing it. Something like `hoverTimer: bool` instead of having it added for however many situations.

I'm not sure I understand, what do you mean by "a button" ? What I did is a kind of factorization : We have only one handler for the whole content.

Ly-sec:
> > @tmarti2 I wonder if it wouldn't make more sense to have a "toggle" to stop timers when we hover over a button instead of having to replicate the same logic over and over again - factorizing it. Something like `hoverTimer: bool` instead of having it added for however many situations.
> 
> I'm not sure I understand, what do you mean by "a button" ? What I did is a kind of factorization : We have only one handler for the whole content.



> > @tmarti2 I wonder if it wouldn't make more sense to have a "toggle" to stop timers when we hover over a button instead of having to replicate the same logic over and over again - factorizing it. Something like `hoverTimer: bool` instead of having it added for however many situations.
> 
> I'm not sure I understand, what do you mean by "a button" ? What I did is a kind of factorization : We have only one handler for the whole content.

Right, my bad I was a bit preoccupied yesterday. 


About:

> Note: I actually need a more centralized way to handle this for another thing I'm working on.


How would you want to centralize it?

tmarti2:
> How would you want to centralize it?

Well this is :grimacing:. I was not able to find a solution that could work globally, so for now I had to implement what I want do manually in each concerned files, so having an handler makes it easier for me since the hovering is handled in a single point in a given file.

For the record I wanted to be able to detect if I was hovering a SmartPanel or not. The heuristic I came with is:
- If the panel is directly hovered, then it is
- Else if another "component" is hovered while this panel is opened, then consider it hovered.

The second point is mandatory because smart panels can contain elements which open smaller ones, like comboboxes, file pickers, contextual menus, notifications etc. At the moment I handle each one of them individually, which is not quite practical and won't work in plugins unless they do what I did. This is WIP anyway.

tmarti2:
@Ly-sec I've rebased to resolve conflicts. Since this moved a bit of code there may be new conflicts soon, do you think you can merge it ?

Ly-sec:
Hey @tmarti2 just read through the code once more and yeah I can merge it :) thank you very much :heart:

PR #1950, opened 2026-02-24
Fix DockMenu clipping & change tooltip positioning

notiant:
There can also sometimes be a clipping of the dock menu's border, I've applied the same fix as for the tooltip.

For the dock's tooltip, I've found that it's not a good idea to try to increase the width of `dockContainerWrapper` to get it to position itself to the left of the dock. This will cause an area of empty space next to the dock where underlying windows can't be interacted with. Instead, I think it's better to remove the code that tries to keep the tooltip entirely on screen in Tooltip.qml; this is usually already handled by the compositor (I have no problems at least on Hyprland).

### DIFF
--- a/Modules/Dock/Dock.qml
+++ b/Modules/Dock/Dock.qml
@@ -792,7 +792,7 @@ Loader {
             readonly property int extraRight: (!isVertical && !exclusive && barOnRight) ? barHeight : 0
 
             // Add +2 buffer for fractional scaling issues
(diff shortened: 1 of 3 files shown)


ItsLemmy:
I hate the tooltip code that tries to keep them on screen unfortunately all wlroots based compositor need it, or the tooltips will show across screens.

notiant:
> I hate the tooltip code that tries to keep them on screen unfortunately all wlroots based compositor need it, or the tooltips will show across screens.

Oh, I didn't consider that. I'll close this PR and might think about a different solution.

PR #1952, opened 2026-02-24
Increase notification close button margin

tmarti2:
# Pull Request

## Motivation

With max `Container radius` setting value, the notification's close button appeared out of the notification background. I increased margins to fix this, it still feels good with normal settings I think?

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring

## Testing

- [x] Tested on niri
- [ ] Tested on Hyprland
- [ ] Tested on sway

### Before

Default: 
image

Max radius
image

### After

Default: 
image

Max radius
image

## Checklist
- [x] Self-reviewed my code


Ly-sec:
Thank you very much @tmarti2 , not only fixes it that but it also does look quite a bit better in my opinion :heart: !

PR #1962, opened 2026-02-24
"Per core" cpu stats in SystemMonitor

tmarti2:
# Pull Request

## Motivation

With Waybar I had a widget that showed Core usage independently and I find it way more usefull than global CPU usage. This MR:
- Adds what is needed to compute Cores usage via `SystemStatService.calculateCpuUsage` and exports them in a array `SystemStatService.coresUsage`
- Extracts `minigaugeComponent` from `SystemMonitor`. This new component is just two rectangles and requires an `orientation` (to decide is we have an horizontal or vertical gauge) and a ratio.
- Adds an option to show Cpu usage by cores : 
  - The tooltip will show both cores usage and total usage
  - The widget will show gauges for each cpu (and not the global one) only in compact mode
  - Gauge orientation depends on if the bar is vertical or not

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Refactoring

## Testing

- [x] Tested on niri
- [ ] Tested on Hyprland
- [ ] Tested on sway

## Screenshots / Videos
image
image

## Checklist
- [x] Self-reviewed my code

### DIFF
--- a/Modules/Bar/Widgets/SystemMonitor.qml
+++ b/Modules/Bar/Widgets/SystemMonitor.qml
@@ -229,7 +229,7 @@ Item {
           id: cpuUsageContent
           anchors.centerIn: parent
 
-          property bool verticalDisplay: isVertical && !compactMode
+          property bool verticalDisplay: isVertical && (!compactMode || showCpuCores)
           flow: verticalDisplay ? GridLayout.TopToBottom : GridLayout.LeftToRight
           rows: verticalDisplay ? -1 : 1
           columns: verticalDisplay ? 1 : -1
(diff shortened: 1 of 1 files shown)


tmarti2:
@ItsLemmy I think it is ready. I've tested this by tweaking the code and the display works for 1-32 cores. You might want to check in details `SystemStatService.calculateCpuUsage`, I may lack JS knowledges to write this more efficiently. I do a double iteration which can be avoided mainly as a way to avoid negative numbers (I fill the array with 0 if needed, then keep the previous value if I find a negative number, as done for the normal CPU calculation).

PR #1981, opened 2026-02-26
feat(bluetooth): auto-connect paired & trusted devices with toggle control

cbxcvl:
## Summary

- Adds automatic reconnection of paired and trusted Bluetooth devices when Bluetooth is enabled or when the shell starts
- New setting `bluetoothAutoConnect` (default: `true`) persisted via `Settings.data.network`
- Auto-trust newly paired devices via `Instantiator` watching `onPairedChanged`
- 2s delay after BT enable to allow adapter initialization

## Toggle control — 3 entry points

- **Settings > Connections > Bluetooth** — persistent `NToggle`
- **Bluetooth Panel header** — quick `NIconButton` (icon + color reflect state)
- **IPC** — `toggleAutoConnect`, `enableAutoConnect`, `disableAutoConnect`

```bash
qs ipc call bluetooth toggleAutoConnect
qs ipc call bluetooth enableAutoConnect
qs ipc call bluetooth disableAutoConnect
```

## Behavior

| State | Effect |
|-------|--------|
| ON | On BT enable or shell start, connects all `paired + !connected + !blocked` devices |
| OFF | No automatic connections; existing connections unaffected |
| New pairing (ON) | Device is auto-trusted immediately on `pairedChanged` |

Respects airplane mode — `attemptAutoConnect` returns early if enabled.

## Files changed

| File | Change |
|------|--------|
| `Commons/Settings.qml` | `bluetoothAutoConnect: true` property |
| `Services/Networking/BluetoothService.qml` | Timer, `attemptAutoConnect()`, Instantiator, reactive `onEnabledChanged` |
| `Modules/Panels/Settings/Tabs/Connections/BluetoothSubTab.qml` | NToggle |
| `Modules/Panels/Bluetooth/BluetoothPanel.qml` | NIconButton in header |
| `Services/Control/IPCService.qml` | 3 IPC functions |
| `Assets/Translations/en.json` | i18n keys (en) |

### DIFF
--- a/Services/Networking/BluetoothService.qml
+++ b/Services/Networking/BluetoothService.qml
@@ -67,15 +67,15 @@ Singleton {
   // Internal: temporarily pause discovery during pair/connect to reduce HCI churn
   property bool _discoveryWasRunning: false
   property bool _ctlInit: false
-  property bool _autoConnectInProgress: false
 
-  // Mirror the setting so we can react to changes via onAutoConnectEnabledChanged
-  property bool autoConnectEnabled: Settings?.data?.network?.bluetoothAutoConnect ?? true
-  onAutoConnectEnabledChanged: {
-    if (autoConnectEnabled && adapter && adapter.enabled) {
-      autoConnectTimer.restart();
-    } else {
-      autoConnectTimer.stop();
+  Connections {
+    target: Settings.data.network
+    function onBluetoothAutoConnectChanged() {
+      if ((Settings?.data?.network?.bluetoothAutoConnect ?? true) && adapter && adapter.enabled) {
+        autoConnectTimer.restart();
+      } else {
+        autoConnectTimer.stop();
+      }
     }
   }
 
@@ -105,7 +105,7 @@ Singleton {
       Quickshell.execDetached(["rfkill", "block", "bluetooth"]);
     }
     // Auto-connect on startup if BT is already enabled
-    if (root.autoConnectEnabled && adapter && adapter.enabled) {
+    if ((Settings?.data?.network?.bluetoothAutoConnect ?? true) && adapter && adapter.enabled) {
       autoConnectTimer.restart();
     }
   }
@@ -129,7 +129,7 @@ Singleton {
       checkAirplaneMode.running = true;
     }
     function onEnabledChanged() {
-      if (adapter && adapter.enabled && root.autoConnectEnabled) {
+      if (adapter && adapter.enabled && (Settings?.data?.network?.bluetoothAutoConnect ?? true)) {
         autoConnectTimer.restart();
       }
     }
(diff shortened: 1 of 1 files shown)


Services/Networking/BluetoothService.qml
```diff
@@ -75,6 +86,29 @@ Singleton {
...
+  Instantiator {
+    id: deviceWatcher
+    active: root.autoConnectEnabled && root.adapter !== null
+    model: root.adapter ? root.adapter.devices : null
+    delegate: Connections {
+      required property var modelData
+      target: modelData
+      function onPairedChanged() {
+        if (modelData.paired && !modelData.trusted) {
+          Logger.i("Bluetooth", "Auto-trusting newly paired device:", modelData.name || modelData.deviceName);
```
turannul: In process of pairing, trusting was part of it so: Paired devices are already trusted.

cbxcvl: Correct — the pairing script already calls `bluetoothctl trust` after a successful pair, so the condition `!modelData.trusted` was never true. Removed the block.

Services/Networking/BluetoothService.qml
```diff
@@ -597,6 +624,33 @@ Singleton {
...
+    if (!adapter || !adapter.enabled) return;
+    if (!autoConnectEnabled) return;
+
+    _autoConnectInProgress = true;
+
+    var devList = adapter.devices.values.filter(function(dev) {
+      return dev && dev.paired && !dev.connected && !dev.blocked;
+    });
+
+    var count = devList.length;
```
ItsLemmy: count seems useless here, you could use devList.length directly

Services/Networking/BluetoothService.qml
```diff
@@ -597,6 +624,33 @@ Singleton {
...
+    for (var i = 0; i < devList.length; i++) {
+      Logger.i("Bluetooth", "Auto-connecting to:", devList[i].name || devList[i].deviceName);
+      connectDeviceWithTrust(devList[i]);
+    }
+
+    if (count > 0) {
```
ItsLemmy: devList.length

Services/Networking/BluetoothService.qml
```diff
@@ -67,6 +67,17 @@ Singleton {
   // Internal: temporarily pause discovery during pair/connect to reduce HCI churn
   property bool _discoveryWasRunning: false
   property bool _ctlInit: false
+  property bool _autoConnectInProgress: false
+
+  // Mirror the setting so we can react to changes via onAutoConnectEnabledChanged
+  property bool autoConnectEnabled: Settings?.data?.network?.bluetoothAutoConnect ?? true
```
ItsLemmy: Would be more efficient with "Connections" 
ex : 

      Connections {
        target: Settings.data.network
        function onBluetoothAutoConnectChanged() {
           ....
        }
      }

Services/Networking/BluetoothService.qml
```diff
@@ -597,6 +624,33 @@ Singleton {
     forgetDevice(device);
   }
 
+  function attemptAutoConnect() {
+    if (_autoConnectInProgress) return;
```
ItsLemmy: I don't believe _autoConnectInProgress does anything.   
connectDeviceWithTrust()  calls device.connect() which is async. 
So this is not guarding re-entry.

Services/Networking/BluetoothService.qml
cbxcvl: Done

Services/Networking/BluetoothService.qml
cbxcvl: Done

Services/Networking/BluetoothService.qml
cbxcvl: Done

Services/Networking/BluetoothService.qml
cbxcvl: Done

PR #1988, opened 2026-02-27
IPC dismiss also clear from history

### LINKED ISSUE #1213: Action to dismiss AND delete notification (oldest and newest)
### Feature Description
There are operations to dismiss and to delete old notifications only. Although we could call both, the oldest notification to dismiss and to delete are different notifications if there's another dismissed notification.

### Why Is This Needed?
If we want to delete and dismiss a notification we have to always dismiss the notification we want and then either manually clean it from the notification menu or clear all notifications.

### Suggested Solutions
There's 3 ways I can think about it:
1. Have a single operation to delete and dismiss
2. Make delete always dismiss (but his necessarily means we need to have a "newest" option as the oldest might always be another notification
3. Have options to dismiss/delete the newest as we could call both and the newest is always the same one, but still requires 2 operations which is not ideal.

### Additional Context
Add any relevant screenshots, links, or resources.

tmarti2:
# Pull Request

## Motivation

I added recently a setting to clear notification from history when manually dismissed (close/right click/swipe the notification), but I didn't do the IPC calls, this MR do that.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring

## Related Issue
- Closes #1213

## Testing

- [x] Tested on niri
- [ ] Tested on Hyprland
- [ ] Tested on sway

## Checklist
- [x] Self-reviewed my code

## Additional Notes

### DIFF
--- a/Services/Control/IPCService.qml
+++ b/Services/Control/IPCService.qml
@@ -229,13 +229,13 @@ Singleton {
 
       var actions = JSON.parse(notif.actionsJson || "[]");
       if (actions.length === 0) {
-        NotificationService.dismissActiveNotification(notif.id);
+        NotificationService.userDismissNotification(notif.id);
         return false;
       }
 
       var actionId = actions.find(a => a.identifier === "default")?.identifier ?? actions[0].identifier;
       var result = NotificationService.invokeAction(notif.id, actionId);
-      NotificationService.dismissActiveNotification(notif.id);
+      NotificationService.userDismissNotification(notif.id);
       return result;
     }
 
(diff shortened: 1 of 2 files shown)


tmarti2:
Wait nvm it does not work yet 😬

tmarti2:
Ok, I think this works, at least visually. Tbh I'm not sure I understand well the internal states of notifications, it's really sprawling (use translation for this one lol), several lists, several way to remove/clean/dismiss. This patch works but it may not be the good way to do it

ItsLemmy:

Cheers!

Issue #1716, opened 2026-02-07
Feature Suggestion – Workspace Preview


dinhokusanagi:
Hi Noctalia team! 👋

I’d like to suggest a new feature for the shell: Workspace Preview.

How it would work:

Each workspace would have a visual area in the workspace bar.

If there are open windows in the workspace, only the icon of the focused window would be displayed in that area.

When the focus changes to another window in the same workspace, the icon would update dynamically to reflect the active window.

Benefits:

Clean and intuitive interface, without cluttering the workspace bar with multiple window icons.

Quickly shows which window is active in each workspace.

Easy to implement, as it only relies on the active window per workspace, without tracking all open windows.

I’ve attached a screen recording illustrating the concept and user experience.

I believe this feature could provide an additional way to visualize workspaces in Noctalia.

Thank you for your time and attention!

https://github.com/user-attachments/assets/f1ade38c-5b3e-43e1-a0fc-4787c9fd2941

Ly-sec:
Thanks so much for reaching out and sharing this with us!

Right now, our team is entirely focused on building v5. Since the v4 lifecycle has officially come to an end, we are wrapping up old issues and won’t be shipping any further updates or fixes to that codebase.

We are closing this issue for now, but we truly appreciate you taking the time to report it. We aren’t quite ready to open up the v5 tracker for public bug reports or feature requests just yet, but we’ll definitely keep your feedback in mind as we develop the new version!

Cheers!

PR #1991, opened 2026-02-27
Show workspace applications on hover

### LINKED ISSUE #1350: Workspace widget - add an option to show applications on hover
### Feature Description

Add a hover-based application display feature to the workspace widget.
By default, the widget shows different workspaces in a simple capsule-style view. When hovering over the widget, the applications are displayed.

### Why Is This Needed?

Currently, the workspace widget has a “show applications” option, which is very helpful for quickly locating applications across different workspaces. However, when there are many open windows, the widget becomes a long row of colorful icons, which can be visually distracting during focused work.

The minimal capsule-style display is much less attention-grabbing and helps maintain immersion while working. At the same time, revealing applications on hover would still allow fast and intuitive workspace switching when needed.

### Suggested Solutions

Introduce a hover-to-reveal mode for the workspace widget.

The widget remains in a minimal capsule view by default, and only displays application icons when the user hovers over the widget. This could be implemented as an optional setting alongside the existing “show applications” option.

tmarti2:
# Pull Request

## Motivation
I saw issue #1350 and thought I could give it a try, since I'm into hovering stuff.
This PR add a new setting to show application only when hovered (requires show applications enabled).
I struggled a bit as always and settled on a solution with an evaluation timer:
- If we're hovering the widget directly, show the app
- If the context menu is visible, show the app
- We either  exit the context menu or leave the widget, we start an evaluation timer, which on trigger will check if one of the two above is true or not

This solution avoids race condition on hovered status.

It feels a bit rough as is because I did not put any animation for the transition (shall we ?).

Also I found some properties which were not used about hovering, so I just removed them, tell me if they're somehow required. At the moment I feel like the application display isn't responsive at all, unlike pill display which as smooth animation, hovering effect, etc, application display as nothing except border color. I'm adding this at the bottom of my never ending todolist.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring

## Related Issue
- Closes #1350

## Testing

- [x] Tested on niri
- [ ] Tested on Hyprland
- [ ] Tested on sway

## Screenshots / Videos

## Checklist
- [ ] Self-reviewed my code

## Additional Notes

### DIFF
--- a/Modules/Bar/Widgets/Workspace.qml
+++ b/Modules/Bar/Widgets/Workspace.qml
@@ -131,7 +131,7 @@ Item {
   property bool isHovered: false
 
   HoverHandler {
-    id: hoverHandler
+    id: workspaceHoverHandler
     enabled: showApplications && showApplicationsHover
     onHoveredChanged: {
       if(hovered) {
@@ -145,10 +145,10 @@ Item {
 
   Timer {
     id: hoverEval
-    interval: 50
+    interval: 150
     repeat: false
     onTriggered: {
-      isHovered = hoverHandler.hovered || contextMenu.visible;
+      isHovered = workspaceHoverHandler.hovered || contextMenu.visible;
     }
   }
 
@@ -284,7 +284,11 @@ Item {
   onScreenChanged: Qt.callLater(refreshWorkspaces)
   onScreenNameChanged: Qt.callLater(refreshWorkspaces)
   onHideUnoccupiedChanged: Qt.callLater(refreshWorkspaces)
-  onAppVisibleChanged: Qt.callLater(refreshWorkspaces)
+  onAppVisibleChanged: {
+    if (appVisible) {
+       Qt.callLater(refreshWorkspaces)
+    }
+  }
 
   Connections {
     target: CompositorService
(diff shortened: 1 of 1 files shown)


tmarti2:
I'm drafting this, I need a smooth transition between both versions of the widget, I'll work on that soon.

tmarti2:
Ok I did something simple because I did not succeed in making something smarter.

I added three things (in 3 separate commits for review):
- A scale animation when the display of the widget changes
- A scale animation when hovering an icon (I had to change scale depending on layout to avoid icon exiting the group border)
- Change border color (if enabled) when hovering a group

What do you think ? tbh I'm not good with animations so I went for something straightforward but we can make improvements if you think it's needed.


https://github.com/user-attachments/assets/86947048-2fee-4e6a-bee5-e7d181784961

dinhokusanagi:
I looked at this PR and also considered the layout from #1716. I think this PR brings good interactivity with hover and simple animations, which is great for those who prefer a more discreet interface.

My suggestion would be to try combining the best of both: keep the hover logic and animations from this PR, but adopt the more fluid and visual layout from #1716. This way, the interface would be both interactive and give a better sense of the open applications.

Overall, I like the animations and the attention to visual feedback — even simple, they already improve the experience quite a bit.

ItsLemmy:
- Don't use the same id: hoverHandler (3 times) everywhere, it will work with scope but makes the code harder to maintain.  Maybe: rootHoverHandler, wsHoverHandler, appHoverHandler.

- Calling refreshWorkspaces() on every hover/leave seems overkill, try:
```
  onAppVisibleChanged: {
    if (appVisible) {
      refreshWorkspaces();
    }
 ```

- 50ms is very short, maybe 100-150ms would be more appropriate.

- Don't want any scale animation on app hover, we had that in the past and got removed to ensure all icons look crisp at all time. Can you copy the design of the taskbar that has an alternate color dot indicator.

Thanks

tmarti2:
@ItsLemmy Thanks for the review !

The last commit applies all the changes you asked for, a few comments:
- I'm not a fan of the hover indicator of taskbar (but copied it nonetheless) because you do not see it (it's below the cursor most of the time).
- Tbh I don't like the animation triggered when hovering the widget, I tried my best but I'm not good with animations (I don't know how to make things looks good), so if you have a better idea do not hesitate to say it (or do it if you prefer). I just added this because it looks even uglier without it :grimacing:

ItsLemmy:
another conflict to fix

tmarti2:
done

PR #2000, opened 2026-02-28
Use decimal units for disk space & network speeds

notiant:
Unless RAM, disk space is usually given in decimal units while it's currently displayed in binary units in Noctalia (it should show the same value as e.g. Nautilus or Gnome Disks display).

I also added the border clipping fix to the dock menu.

### DIFF
--- a/Modules/Bar/Widgets/SystemMonitor.qml
+++ b/Modules/Bar/Widgets/SystemMonitor.qml
@@ -110,11 +110,11 @@ Item {
     }
 
     // Memory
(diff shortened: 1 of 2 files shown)


notiant:
I've changed the network speed in the system monitor panel to be calculated in decimal units as well since this is more common. Further, I've made the distinction between "GB" for decimal units and "GiB" for binary units.

Ly-sec:
Looks good to me @notiant , thank you very much!

PR #2002, opened 2026-02-28
feat(theming): add qimgv template

HANDZCZ:
# Pull Request

## Motivation

Adds theme template for [qimgv](https://github.com/easymodo/qimgv).

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring

## Testing

Tested with multiple predefined color schemes.

- [x] Tested on niri
- [ ] Tested on Hyprland
- [ ] Tested on sway

## Screenshots / Videos

Screenshot from 2026-02-28 21-14-57
Screenshot from 2026-02-28 21-21-38

## Checklist
- [x] Self-reviewed my code

## Additional Notes

If you can, please check that the theme looks somewhat decent and is similar to other apps.
Also, I couldn't find anything about hot-reloading the theme, so it only updates when relaunching the app.

### DIFF
--- a/Assets/Templates/qimgv.conf
+++ b/Assets/Templates/qimgv.conf
@@ -1,0 +1,14 @@
+[Colors]
+accent={{colors.on_surface_variant.default.hex}}
+background={{colors.surface.default.hex}}
+background_fullscreen={{colors.surface.default.hex}}
+folderview={{colors.surface.default.hex}}
+folderview_topbar={{colors.surface_variant.default.hex}}
+icons={{colors.primary_container.default.hex | auto_lightness 15}}
+overlay={{colors.surface_variant.default.hex}}
+overlay_text={{colors.on_surface.default.hex}}
+scrollbar={{colors.surface_bright.default.hex}}
+text={{colors.on_surface.default.hex}}
+tid=3
+widget={{colors.surface_variant.default.hex}}
+widget_border={{colors.primary.default.hex}}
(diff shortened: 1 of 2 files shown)


ItsLemmy:
This seems to be Qt app, how was it looking without a custom template? In theory it should follow Qt theming.
We only want to offer the common apps as built int templates the other more niche one should be user-templates to avoid cluttering the UI and codebase.

HANDZCZ:
This is on Eldritch color scheme when "use system colors" is on:

Scroll level is not visible, has wrong text color and background color:
image

Background is inconsistent, black when viewing the picture, and the right color when browsing folders, outline for the popups is missing (wrong color):
image
Forgot to add how it looks when themed...
image

Folders have this whiteish color, which is really bright on a dark theme:
image

ItsLemmy:
have you tried to see if this was fixable on the main Qt template ?

HANDZCZ:
~I can try~, but I thought that the Qt template was "final" and qimgv is using it wrong.

Ok, I just looked, and qimgv is not even setting the background color or overlay text/background color, so it is 0x000000.
It uses the same color for the popup and its border.
Here is the color definition part: https://github.com/easymodo/qimgv/blob/master/qimgv/themestore.cpp#L9-L18
Basically, the entire "use system theme" option is almost unusable for actual theming.

It also doesn't help that I can't find the format for the Qt ColorScheme option and any Qt color scheme testing app.
So I can't even fix qimgv and the Qt template (if needed).

ItsLemmy:
It's most likely  true: qimgv does not use Qt colors as it should, and they should fix it. Maybe open a bug report there.

I recommend you keep this as a user template for now, which work just fine.
Unless there is a ton of traction for that PR, I do not plan to merge this as it's a pretty small app.

PR #2021, opened 2026-03-01
Idle settings: update idle time only every 5s

notiant:
I think it's better if the current idle time in the settings is only updated every 5s since it's too distracting otherwise. I also changed the settings title to a new key since "common.idle" doesn't work in every language (already added on the i18n website) and changed NSpinBox to also accept the Enter key to save the current value.

### DIFF
--- a/Modules/Panels/Settings/Tabs/Idle/BehaviorSubTab.qml
+++ b/Modules/Panels/Settings/Tabs/Idle/BehaviorSubTab.qml
@@ -37,10 +37,10 @@ ColumnLayout {
 
     NText {
       Layout.alignment: Qt.AlignBottom | Qt.AlignRight
-      text: IdleService.idleSeconds >= 5 ? I18n.trp("common.second", Math.floor(IdleService.idleSeconds / 5) * 5) : I18n.tr("common.active")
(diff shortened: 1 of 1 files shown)


ItsLemmy:
the idle time is for debugging  and not to stare at it, I don't see a good reason to update it every 5s.

notiant:
I can remove that change, but I just found the constant change between "active" and "1 second" when editing settings on the page a bit annoying.

ItsLemmy:
yes please

Modules/Panels/Settings/Tabs/Idle/BehaviorSubTab.qml
```diff
@@ -37,7 +37,7 @@ ColumnLayout {
 
     NText {
       Layout.alignment: Qt.AlignBottom | Qt.AlignRight
-      text: IdleService.idleSeconds > 0 ? I18n.trp("common.second", IdleService.idleSeconds) : I18n.tr("common.active")
+      text: IdleService.idleSeconds > 0 ? I18n.trp("common.second", IdleService.idleSeconds : I18n.tr("common.active")
```
ItsLemmy, line 40: this is gonna break

PR #2027, opened 2026-03-01
Launcher: reduce label font for grid view

notiant:
The label font for launcher entries in grid view is quite large, I have reduced it a bit.
Also fixing missing default values for some idle settings.


ItsLemmy:
Thanks

PR #2039, opened 2026-03-02
feat(wallpapers): implement themed wallpapers feature

### LINKED ISSUE #1612: Switch wallpaper according to light/dark theme changes
### Feature Description

I’d like Noctalia to support binding a wallpaper to each theme variant:

* **Light theme → Wallpaper A**
* **Dark theme → Wallpaper B**

When the theme switches (manually or automatically), Noctalia should switch to the corresponding wallpaper automatically (ideally using Noctalia’s existing wallpaper transition animation).

### Why Is This Needed?

Noctalia already supports automatically switching between light/dark themes (e.g., based on location/time). In practice, users often want the wallpaper to follow the theme as well.

Right now this can be approximated with **Hooks** (Theme Changed → call `ipc wallpaper set ...`), but it can lead to a race/competition at login:

Example scenario:

1. Before shutdown, the system is in **dark theme**.
2. After reboot/login, the automatic theme logic decides it should be **light theme** (e.g., morning).
3. During session startup, Noctalia may first restore the previous **dark theme** and trigger the hook → wallpaper switches to **dark wallpaper**.
4. Immediately after, the system switches to **light theme** and triggers the hook again → wallpaper switches again to **light wallpaper**.

Result: the wallpaper transitions **twice** during login, and the first transition may still be in progress when the second starts.
Expected behavior: wallpaper should switch **only once**, based on the final theme state after startup has stabilized.

This happens frequently in a real workflow: using the PC at night (dark theme), shutting down, and then booting in the morning (light theme).

This feels like something Noctalia should handle internally, because Noctalia has the best knowledge of:

* when the theme state is “final” vs “intermediate” during startup,
* when wallpaper transitions are in progress,
* and how to avoid duplicate transitions.

### Suggested Solutions

One possible implementation approach:

1. **Add per-theme wallpaper settings**

   * A “Light wallpaper” and “Dark wallpaper” entry in wallpaper settings (or theme settings).
   * Optional per-monitor overrides if Noctalia already supports per-monitor wallpaper.

2. **Internal debounce / “final state wins” logic**

   * If multiple theme changes happen within a short window during startup (or while restoring session), only apply the wallpaper change once, based on the final theme state.
   * Alternatively: queue wallpaper changes and cancel any in-progress/queued transition when a newer theme state arrives.

3. **Startup alignment**

   * After Noctalia finishes loading (or after the auto-theme decision is applied), sync wallpaper to the current theme exactly once.
   * This would prevent the “restore old theme then immediately switch” issue.

4. **Expose as a first-class feature**

   * This would remove the need for user-side hook scripts and avoid race conditions caused by external commands.

### Additional Context

* Current workaround: use “Theme Changed” hook to run `ipc call wallpaper set ...`, but it’s susceptible to the startup race described above.
* I’m using **niri** (single monitor `eDP-1`), but I believe this can affect any setup where theme is automatically changed on login.

kaeeraa:
# Pull Request

## Motivation
Implement themed wallpapers option that automatically switch between light and dark variants based on the current color scheme.
This addresses a common use case where users want wallpapers to stay visually consistent with light/dark mode without relying on random rotation or manual switching.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring

## Related Issue
- Closes #1612 

## Testing

- [ ] Tested on niri
- [ ] Tested on Hyprland
- [ ] Tested on sway
- [x] **Tested on mangowc**

## Screenshots / Videos
showcase

## Checklist
- [x] Self-reviewed my code

## Additional Notes
I could not find any `qmlfmt` for NixOS, only `qmlformat`. And seems that `qmlformat` is not the same thing because it's changes all whitespaces..

**Testers needed!!!**

Feel free to report your feedback about this implementation


ItsLemmy:
Have yet to look into the details, but seems a bit redundant with @tuibird  favorite system

wxmup:
Looking forward to the merging of this feature!

notiant:
This feature might be better integrated with the favorite system and I think @tuibird is currently working on the wallpaper panel and might integrate this functionality?
Apart from that, the current version has some very obvious bug:
Screenshot_2026-03-03 15-43-04

fractal-l:
> Apart from that, the current version has some very obvious bug:
> Screenshot_2026-03-03 15-43-04

Hmm? Can't reproduce..
image

---

> This feature might be better integrated with the favorite system

Honestly, i have no idea what could be integrated, lgtm right now.. Any ideas?

notiant:
The visual bug only happens on Niri cause there are extra settings below. But I thought it might be better to implement the dark/light mode specific wallpaper by making it a favorite and clicking on the already existing moon/sun icon in the wallpaper panel instead of adding more visual clutter and extra settings.

fractal-l:
> The visual bug only happens on Niri cause there are extra settings below. But I thought it might be better to implement the dark/light mode specific wallpaper by making it a favorite and clicking on the already existing moon/sun icon in the wallpaper panel instead of adding more visual clutter and extra settings.

Okay! I'm going to improve my feature, also there's some bugs i noticed while using it

fractal-l:
> The visual bug only happens on Niri cause there are extra settings below. But I thought it might be better to implement the dark/light mode specific wallpaper by making it a favorite and clicking on the already existing moon/sun icon in the wallpaper panel instead of adding more visual clutter and extra settings.

But about extra settings, does `themed wallpapers` toggle counts too? I don't see another way to make this functional toggleable

fractal-l:
Also i will integrate this into `Automation` by allowing to choose multiple wallpapers for same theme variant

fractal-l:
Okay, need feedback for this design. Integrated with "favorites" system and removed toggle as it's not needed anymore.
But maybe it became a little bit less intuitive..

fractal-l:
I'll squash these commits into single one after everything will be completed

Modules/Panels/Settings/Tabs/Wallpaper/GeneralSubTab.qml
```diff
@@ -13,6 +13,9 @@ ColumnLayout {
   Layout.fillWidth: true
 
   property var screen
+  function openWallpaperPanel() {
```
notiant, line 16: This function seems unused and probably can be removed

Modules/Panels/Wallpaper/WallpaperPanel.qml
```diff
@@ -852,11 +853,14 @@ SmartPanel {
...
       if (isDirectory) {
         WallpaperService.setBrowsePath(targetScreen.name, path);
-      } else {
-        var screen = Settings.data.wallpaper.setWallpaperOnAllMonitors ? undefined : targetScreen.name;
-        WallpaperService.changeWallpaper(path, screen);
-        WallpaperService.applyFavoriteTheme(path, screen);
+        return;
       }
+      var screen = Settings.data.wallpaper.setWallpaperOnAllMonitors ? undefined : targetScreen.name;
+      if (Settings.data.wallpaper.themedWallpapers.enabled) {
```
notiant, line 859: This resets the dark/light mode selection when clicking on another wallpaper. How can you select a wallpaper for dark and light mode this way?

Modules/Panels/Wallpaper/WallpaperPanel.qml
```diff
@@ -1149,6 +1175,12 @@ SmartPanel {
                 imagePath: wallpaperItem.cachedPath
                 radius: Style.radiusM
                 borderColor: {
+                  if (wallpaperItem.darkMode) {
+                    return Color.mSecondary;
+                  }
+                  if (wallpaperItem.lightMode) {
```
notiant, line 1181: Not a fan of this, just leave the color as mSecondary when selected

notiant:
The functionality for having specific wallpapers for light and dark mode has now been implemented into Noctalia by @Ly-sec so I think you can close this PR.

Ly-sec:
Indeed, I sadly totally missed that PR and implemented the whole light/dark split wallpaper stuff already. So I'll be closing this one.

fractal-l:
Np! I just needed this function in Noctalia. Implementation doesn't matter

wxmup:
I am glad to hear that noctalia now supports automatic wallpaper switching. However, where can I enable this option? I have been looking for it for a long time and haven't been able to find it.

图片

fractal-l:
> I am glad to hear that noctalia now supports automatic wallpaper switching. However, where can I enable this option? I have been looking for it for a long time and haven't been able to find it.
> 图片

image

You can toggle it in wallpaper selector with these two tabs

wxmup:
> > I am glad to hear that noctalia now supports automatic wallpaper switching. However, where can I enable this option? I have been looking for it for a long time and haven't been able to find it.
> > 图片
> 
> image
> 
> You can toggle it in wallpaper selector with these two tabs

Thanks! There is deeply hide.

PR #2042, opened 2026-03-02
Auto hide panels

tmarti2:
Proof of concept : Auto hide SmartPanels when leaving them.
 
### New settings in `User Interface > Panels`
- "Never" (by default) current behavior
- "Once entered" auto-hide a panel when mouse leaves it
- "Always" also auto-hide the panel if we never enter it
  
Delays when entering and before entering can be changed in settings.

#### Components
This was really tricky, I settled on the version I came up with the less invasive changes : each component that could trigger the auto-hide feature needs to be handled separately, I identified few components : 
- Toast & Notification
- Popup
- Tray Menu
- SmartPanel

For `Toast` and `Notification`, I just used the existing `HoverHandler`. 
For `SmartPanel` I added a `HoverHandler` in the `contentLoader`. 
For `TrayMenu` I used the visibility
For `Popup` I also used visibility, but to avoid duplicating the code everywhere I created a `NPopup` widget which just overloads `Popup` with a simple binding.

#### Logic

The logic is quite simple, I have three properties:
- `isDirectlyHovered`: set by the handler of `SmartPanel`
- `isChildrenHovered`: used by Notification/Toast handlers
- `isChildrenVisible`: It was quite tedious to have handlers everywhere, notably I was not able to factorize the code easily as I did with `NPopup`, so I thought it was a bad idea to use hovering, so I choose visibility instead. If a popup is opened, we do not close panels (for example combo boxes, and many other widgets).
- `isHovered`: disjunction of all the above

And so I have a timer:
- Setting any of the above to `true` stops the timer
- Setting any of the above to `false` will recompute `isHovered` and start the timer if it is false.
- Once the timer triggers, if `isHovered` is still false, the setting is enabled and the panel is still open, we close it. These checks are mainly here to avoid race condition, in theory setting `isHovered` to true stops the timer, and closing the panel too.

TODO: Take into account the hovering of the bar widget corresponding to the opened panel

Known limitation: Well, as mentioned it requires to identify everything that should stop auto-hide. For instance It won't work on plugins that have contextual menus/popups in their panels, since we need to replace `Popup` for example, etc.

@ItsLemmy I'm curious to get your opinion on this one at this point, even if it's still in progress. This part must be done before the "open panel when hovering widget" feature.

Edit: Also I came up with (imo) bad names, so I'll probably find better ones. This also includes setting labels/descriptions

### DIFF
--- a/Assets/Translations/en.json
+++ b/Assets/Translations/en.json
@@ -696,6 +696,11 @@
       "centered": "Centered panel",
       "window": "Separate window"
     },
+    "settings-panel-auto-hide": {
+      "never": "Never",
+      "once-entered": "Once entered",
+      "always": "Always"
+    },
     "visualizer-types": {
       "linear": "Linear",
       "mirrored": "Mirrored",
@@ -1684,6 +1689,12 @@
       "panel-background-opacity-label": "Panel background opacity",
       "panels-attached-to-bar-description": "Panels lock to the bar and screen edges, creating a seamless look with stylish inverted corners.",
       "panels-attached-to-bar-label": "Snap panels to edges",
+      "panels-auto-hide-description": "Auto-hide a panel after mouse leaves",
+      "panels-auto-hide-label": "Auto-hide",
+      "panels-auto-hide-delay-description": "Time before panel hides automatically after mouse leaves.",
+      "panels-auto-hide-delay-label": "Auto hide delay",
+      "panels-auto-hide-delay-before-enter-description": "Time before panel hides automatically before mouse enters.",
+      "panels-auto-hide-delay-before-enter-label": "Auto hide delay before entering",
       "panels-overlay-description": "Ensures panels and the bar remain visible, even over fullscreen applications.",
       "panels-overlay-label": "Keep panels & bar on top",
       "scaling-description": "Changes the size of the general user interface, excluding the bar.",
--- a/Assets/settings-default.json
+++ b/Assets/settings-default.json
@@ -155,6 +155,9 @@
     "panelBackgroundOpacity": 0.93,
     "translucentWidgets": false,
     "panelsAttachedToBar": true,
+    "panelsAutoHide": "never",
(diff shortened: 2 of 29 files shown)


turannul:
Alongside can you get rid of all (x) close buttons please?- its annoying fk out of me since day one.

tmarti2:
I can do that I guess

tmarti2:
@turannul well maybe it should in fact be done in a separate MR, there is a few things to considere and we probably should avoid adding things to this one.

- Close button are not only on smart panels, but also settings popups, notifications, etc. - Escape does not work to close plugin settings, maybe it should be fixed along side this new setting.
- Maybe the auto-hide feature should be added to popups (like widget settings) ?

tmarti2:
My last commit simplifies the logic a but by using Qt bindings. So we only start the timer when `isHovered` is false now.

Also I forgot about the static dock, which handles it's how hide feature. I'll need to look into that to either turn what I did off for the static dock or use it to replace what's currently done in it.

tmarti2:
@tibssy You did the static dock right ? if you have an opinion on this (last comment before this one) do not hesitate

tmarti2:
I've tested and there only 2 changes to do to turn this off for the dock : set `isHovered` to `true` and stop the SmartPanel timer with the `onOpened` action. The last commit does that.

tibssy:
Hey @tmarti2 Without looking at the code just yet, my first thought is to see if we can reuse the same listener/trigger logic used for the Bar's 'Display mode' auto-hide to close the panels. I’ll check the implementation details later this afternoon and get back to you with more feedback!

tmarti2:
My solution is similar to the bar one, but simpler (BarService is quite complicated at first glance). Like the bar, I use hovering status, but I had to go deeper since `PopupMenuWindow` and `openedPanel` could not be used in my case.

ItsLemmy:

Cheers!

PR #2044, opened 2026-03-03
Fix launcher opening delay

eric-handley:
# Pull Request

## Motivation

After updating recently I noticed a new significant delay between IPC `toggle launcher` and the launcher actually opening, especially with animations disabled. This is sometimes as long as ~1s, which is irritating. 

This PR introduces a couple small fixes to make opening the launcher feel more responsive again

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring

## Related Issue

## Testing

- [ ] Tested on niri
- [x] Tested on Hyprland
- [ ] Tested on sway

## Screenshots / Videos

## Checklist
- [x] Self-reviewed my code

## Additional Notes

### DIFF
--- a/Modules/MainScreen/SmartPanel.qml
+++ b/Modules/MainScreen/SmartPanel.qml
@@ -72,6 +72,8 @@ Item {
   property bool cachedAnimateFromBottom: false
   property bool cachedAnimateFromLeft: false
   property bool cachedAnimateFromRight: false
+
+  readonly property bool animationsDisabled: Settings.data.general.animationDisabled
   property bool cachedShouldAnimateWidth: false
   property bool cachedShouldAnimateHeight: false
 
@@ -1313,7 +1315,7 @@ Item {
           // Make panel visible, now only the intended dimension will animate
           root.isPanelVisible = true;
 
-          if (Style.animationNormal === 0) {
+          if (root.animationsDisabled) {
             // Skip delay when animations are disabled
             root.sizeAnimationComplete = true;
           } else {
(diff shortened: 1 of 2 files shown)


ItsLemmy:
Thanks for the PR, I'm not sure sure why some of you have this lag on opening the Launcher, most of us don't have any problem with it even on older hardware.

Do you start noctalia via your compositor or systemd (we have had many report of weird slowness with systemd)

About the PR, overall the changes look good, but the way to test for animations disabled  needs to be changed.  We dont want to rely on Style.animationNormal for that as its unclear, please use ```Settings.data.general.animationDisabled```  you can have that in a readonly property to make the code cleaner if needed.

eric-handley:
@ItsLemmy Hi! Fixed the animations check. 

I'm definitely starting qs from hyprland. My only other guess is that maybe I have a lot of desktop entries? I'm also using the unicode and kamoji plugins which might be adding some delay.

ItsLemmy:
Yeah probably the large amount of desktop entries then.  I made minor changes, can you try it out and see if it's still fast for you? thanks

eric-handley:
@ItsLemmy it's maybe a little bit faster than before after updating, but strangely it's still a lot slower than when I run the dev version. This made me think it might be because I have different settings between the two versions, so I copied my settings.json to `~/.config/noctalia-dev` and got the same slowdown.

So it seems like this PR does increase opening speed a tiny bit, but the main reason for the delay is something else in my settings

ItsLemmy:
interesting, feel free to share you settings if you feel like it.  apart from animations enable/speed I have a hard time which settings could cause this.

eric-handley:
@ItsLemmy I've identified the difference between settings causing the issue, see #2073

PR #2050, opened 2026-03-03
Add some missing translations

notiant:
I have added the new keys on the web platform.
Also removing panels.idle.custom-entry-timeout-format and panels.idle.custom-entry-timeout-format-plural since common.second already exists.


Ly-sec:
Hey @notiant thank you very much for the PR! Going to merge it (and also push i18n afterwards).

PR #2052, opened 2026-03-03
Proof of concept of a simple isNew feature

tmarti2:
This is just a test of how this could be implemented. But this solution requires to actually add the version when creating a setting, which can be forgotten. I don't know if it is possible to have a cleaner method.


tmarti2:
I'll close this, it's is a bad solution I think and we should probably declare missing settings (between saved settings and new settings) as new, but I have no idea how this part of noctalia work and I don't want to explore that (for now)

PR #2055, opened 2026-03-03
feat(desktop-widgets): Custom sticker desktop widget

spiros132:
# Pull Request
## Motivation
This PR adds a new desktop widget called "Custom Sticker" (name could be changed if that seems weird). This widget allows for displaying custom images on the desktop. This would allow people to customize their desktop even more and make it even more personal.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring

## Related Issue
No issue on github, but there was a feature request from discord. 

## Testing
I tested the widget by going into the desktop edit mode, creating a new Custom Sticker desktop widget, adding an image on the widget, and then re-scaling to see that everything works correctly. 

- [x] Tested on niri
- [ ] Tested on Hyprland
- [ ] Tested on sway

## Screenshots / Videos
An example screenshot of an image on my desktop (don't mind the skull image, it was the first thing I found on the internet xD).
image

## Checklist
- [x] Self-reviewed my code

## Additional Notes
The feature request was for a separate plugin to be made, but in my opinion I believe this fits better in the shell itself and not specifically as a plugin.

One thing I was unsure about was the translation. Some of the other desktop widgets use each others translations, I wasn't sure if that's how it is preferred; having a central type of translation for all desktop widgets, or creating different translations for different widgets. For example the weather widget, uses the clocks translation for the "panels.desktop-widgets.clock-show-background-label".

### DIFF
--- a/Assets/Templates/cava.ini
+++ b/Assets/Templates/cava.ini
@@ -1,5 +1,4 @@
 [color]
-background = '{{colors.surface.default.hex}}'
 foreground = '{{colors.primary.default.hex}}'
 
 gradient = 1
--- a/Assets/Templates/yazi.toml
+++ b/Assets/Templates/yazi.toml
@@ -76,9 +76,9 @@ selected = { reversed = true }
 # : Tabs [[[
 
 [tabs]
-active = { fg = "{{colors.primary.default.hex}}", bold = true, bg = "{{colors.surface.default.hex}}" }
+active = { fg = "{{colors.surface.default.hex}}", bold = true, bg = "{{colors.primary.default.hex}}" }
 inactive = { fg = "{{colors.secondary.default.hex}}", bg = "{{colors.surface.default.hex}}" }
-sep_inner = { open = "[", close = "]" }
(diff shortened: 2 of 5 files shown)


Ly-sec:
Hey @spiros132 this is quite the cool desktop widget but we decided that this should rather be a plugin given it is not something that would fit into noctalia-shell itself. So feel free to create a PR in `noctalia-plugins` instead.

For the translations: if some translation fits 100% we can reuse those instead of polluting the translation files with duplicate keys :).

I'll close this one, thank you very much though!

spiros132:
Hej @Ly-sec. Alright, I'll create this on the noctalia-plugins instead then. Just a question tho, how do I implement a widget setting for the desktop widget? Is that supposed to be the plugin setting itself? Because in that case you won't be able to have multiple desktop widgets with different images on your desktop.

ItsLemmy:
DesktopWidgets created via a plugin directly use the plugin settings. You can check my fancy audio visualizer.  But yeah this would mean a single sticker for now.

PR #2068, opened 2026-03-04
feat(plugins): Added a way for plugins to define their own desktop widget settings

spiros132:
# Pull Request
This PR contains a change in the desktop widget settings dialog that allows plugins to define their own desktopWidgetSettings entrypoint instead of relying on the global plugin settings.

## Motivation
This is needed since some of plugins might want to create multiple desktop widgets where each widget has its own settings instead of using the global plugin settings. For example allowing two clock widgets with different settings to exist at the same time. 

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring

## Related Issue

## Testing
The only plugin that I have installed that includes a desktop widget is the catwalk plugin, by changing a tiny bit of the code and adding the DesktopWidgetSettings file, I correctly saw in the "~/.config/noctalia/settings.json" that the a random value I set got changed correctly.  

- [x] Tested on niri
- [ ] Tested on Hyprland
- [ ] Tested on sway

## Screenshots / Videos
I'm not sure what kind of screenshots / videos to put here since this is mostly a non-visual thing. 

## Checklist
- [x] Self-reviewed my code

## Additional Notes
I'll try to update the documentation as well if this PR goes through. I'll try to give some more ideas and what I've implemented in this plugin below.

In the entrypoints there is now a new field called "desktopWidgetSettings" as follows:
```json
{
  ...
  "entryPoints": {
    ...
    "desktopWidgetSettings": "DesktopWidgetSettings.qml",
    ...
  }
}
```

The "DesktopWidgetSettings.qml" file has three properties as of now that get injected:
```qml
...

ColumnLayout {
  // As in the regular "plugin code"
  property var pluginApi
  
  // Taken from the Settings of the other desktop widget settings.
  property var widgetData
  property var widgetMetadata

  // Signal to tell the dialog we want to save the settings
  signal settingsSaved(var settings)
}
```

Since these settings use the same save functionality the other desktop widget settings, there needs to be a signal called settingsSaved that includes a settings variable. I think that this might look weird to some people since in the plugins you usually save with pluginApi.saveSettings(). 

Would it make sense to instead inject a new variable with all the functionality that the plugin desktop widget needs instead? For example a variable called "widget" where this contains a field called "data" which is equal to the "widgetData" and / or a method called saveSettings() that saves the changed widgetData settings?

### DIFF
--- a/Services/UI/DesktopWidgetRegistry.qml
+++ b/Services/UI/DesktopWidgetRegistry.qml
@@ -248,7 +248,7 @@ Singleton {
     if (root.isPluginWidget(widgetId)) {
       var pluginId = widgetId.replace("plugin:", "");
       var manifest = PluginRegistry.getPluginManifest(pluginId);
-      if (manifest && manifest.entryPoints && manifest.entryPoints.settings) {
+      if (manifest && manifest.entryPoints && (manifest.entryPoints.desktopWidgetSettings || manifest.entryPoints.settings)) {
         hasSettings = true;
       }
     } else {


spiros132:
> Since these settings use the same save functionality the other desktop widget settings, there needs to be a signal called settingsSaved that includes a settings variable. I think that this might look weird to some people since in the plugins you usually save with pluginApi.saveSettings().
> 
> Would it make sense to instead inject a new variable with all the functionality that the plugin desktop widget needs instead? For example a variable called "widget" where this contains a field called "data" which is equal to the "widgetData" and / or a method called saveSettings() that saves the changed widgetData settings?

@ItsLemmy

I'm curious about what you think about this and if I should start working towards implementing a better desktop widget settings way before doing the documentation.

PR #2078, opened 2026-03-05
Add {{mode}} tag to template renderer

notiant:
In order to make themed icons like [here](https://github.com/mots/Noctalia-Dynamic-Icons) working properly I've found the need for accessing the current theme mode (Light/Dark) through the template processor. This PR adds {{mode}} as a tag that resolves to the current theme.

Also removing the Bluetooth auto-connect notification while I have a PR open.


Ly-sec:
Hey @notiant thank you very much for the PR. I'll be merging it.

PR #2092, opened 2026-03-06
Add queue for Bluetooth auto-connect

notiant:
This PR adds a queue for the Bluetooth auto-connect feature so Noctalia won't fire connection attempts for all paired devices at the same time.


ItsLemmy:
Thank you

PR #2094, opened 2026-03-06
fix(hyprland): sort toplevels by x/y coordinates to fix workspace widget ordering in dynamic layouts

rivera-diego:
Bug In Hyprland, when using dynamic tiling layouts that aren't purely chronological (like Master Layout or the new Scrolling Layout), the Workspaces widget displays the application icons out of order compared to their physical placement on the screen. This happens because Hyprland.toplevels.values returns the windows in an unstructured order, usually chronologically based on focus history.

Fix This PR modifies 

HyprlandService.qml
 to:

Safely extract the geometric x and y coordinates directly from the raw IPC data (lastIpcObject.at), falling back to standard variables.
Introduce a toSortedWindowList manual fallback mapping function right before the return iteration.
Sort the matrix based on Spatial representation: workspace first, then X coordinates (left to right), then Y coordinates (top to bottom).
The new automatic parsing in v4.6.0 handles the raw events gorgeously, but chaining it with this geometric sorting creates a bullet-proof, 1:1 screen representation on the Workspaces widget.

### DIFF
--- a/Services/Compositor/HyprlandService.qml
+++ b/Services/Compositor/HyprlandService.qml
@@ -281,7 +281,7 @@ Item {
         }
       }
 
-      windows = windowsList;
+      windows = toSortedWindowList(windowsList);
 
       if (newFocusedIndex !== focusedWindowIndex) {
         focusedWindowIndex = newFocusedIndex;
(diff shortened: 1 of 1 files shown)


Ly-sec:
Hey @rivera-diego , I just tested it and it's seemingly working fine. Thank you very much!

PR #2117, opened 2026-03-08
Grid snapping for scale component in desktop widgets

spiros132:
# Pull Request
Added a grid snapping support for the scale in the DraggableDesktopWidget.qml component

## Motivation
When using the grid to position and scale desktop widgets, I would have expected both position and scale to snap to the grid. This PR makes it so that the scale can also snap to the grid. 

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring

## Related Issue
No related issue found.

## Testing
By using the desktop widgets, scaling them and seeing that they scale correctly with their background on, on the desktop. 

- [x] Tested on niri
- [ ] Tested on Hyprland
- [ ] Tested on sway

## Screenshots / Videos
Here's a video of the grid scale snapping working, it also showcases (not the best example), of how this can be used to scale the desktop widgets, so that they have a more "correct" size compared to each other. 

https://github.com/user-attachments/assets/fa40e5c7-6dad-4e25-b783-36327d3fd7ca

## Checklist
- [x] Self-reviewed my code

## Additional Notes
One final note that I was thinking about is that I used the internal.initialWidth. It hasn't been used in any other part of the file as far as I could see, if it does have a function that I can't see / notice, please do let me know so that I can update the code.

### DIFF
--- a/Assets/Translations/en.json
+++ b/Assets/Translations/en.json
@@ -1020,6 +1020,7 @@
       "edit-mode-description": "Enable edit mode to move and reposition desktop widgets. When enabled, widgets show a drag outline and can be repositioned.",
       "edit-mode-exit-button": "Exit edit mode",
       "edit-mode-grid-snap-label": "Grid snap",
+      "edit-mode-grid-snap-scale-label": "Snap scale",
       "edit-mode-label": "Edit mode",
       "enabled-description": "Enable or disable desktop widgets entirely.",
       "enabled-label": "Enable desktop widgets",
--- a/Assets/settings-default.json
+++ b/Assets/settings-default.json
@@ -539,6 +539,7 @@
     "enabled": false,
     "overviewEnabled": true,
     "gridSnap": false,
+    "gridSnapScale": false,
     "monitorWidgets": []
   }
 }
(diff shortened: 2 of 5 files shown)


Ly-sec:
Hey @spiros132 , I kind of think this should be a toggle in the edit mode (transient perhaps) no?

spiros132:
> Hey @spiros132 , I kind of think this should be a toggle in the edit mode (transient perhaps) no?

Tbh yes, I was thinking the exact same thing. Since people might be wondering why it's snapping now instead of the usual. 

In my opinion I think it makes more sense for both the scale and position to snap in "grid mode" and to stop the snap if you turn "grid mode" off. 

I'll try to implement a toggle in the following days and tell you when it's done :) 

P.S. also just a question, what do you mean by transient?

Ly-sec:
> P.S. also just a question, what do you mean by transient?
Basically that it should reset - aka not get saved in the settings. So it's (let's say) disabled by default but when you want it, you can toggle it on for the current "session" but now that I think of it, maybe it could just be saved in the settings either way. @spiros132

spiros132:
I've added a toggle now, the scale snap is only enabled if the whole grid system is enabled, I think that makes more sense to avoid clutter and confusion about the two buttons. 

@Ly-sec ready for review :)

Ly-sec:
Awesome @spiros132 this is a good implementation :)! I'll be merging it, thank you very much!

PR #2118, opened 2026-03-08
fix(plugins): preserve settings.json during plugin update

DzmingLi:
# Pull Request

## Motivation
When updating a plugin, `installPlugin()` uses `cp -r` to copy downloaded plugin files into the plugin directory, which overwrites the user's `settings.json`. This means any user customizations to plugin settings are lost on every update.

This adds a backup/restore of `settings.json` around the `cp -r` call, preserving user customizations across plugin updates. For fresh installs, the backup step is a no-op (`[ -f ... ]` fails, `|| true` continues).

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring

## Related Issue
- N/A

## Testing
- [x] Tested on niri
- [ ] Tested on Hyprland
- [ ] Tested on sway

Install a plugin via UI, modify its `settings.json`, trigger an update, confirm settings are preserved. Fresh install a new plugin, confirm it works normally (no stale backup interferes).

## Screenshots / Videos
N/A

## Checklist
- [x] Self-reviewed my code

## Additional Notes
Only `PluginService.qml` is changed. The existing `updatePlugin()` backup logic for bar/desktop widgets is unaffected.

### DIFF
--- a/Services/Noctalia/PluginService.qml
+++ b/Services/Noctalia/PluginService.qml
@@ -441,8 +441,11 @@ Singleton {
     // Use git sparse-checkout to clone only the plugin subfolder
     // GIT_TERMINAL_PROMPT=0 prevents hanging on private repos that need auth
     // Note: We download from the original pluginId folder in the repo, but save to compositeKey folder
-    var downloadCmd = "temp_dir=$(mktemp -d) && GIT_TERMINAL_PROMPT=0 git clone --filter=blob:none --sparse --depth=1 --quiet '" + repoUrl + "' \"$temp_dir\" 2>/dev/null && cd \"$temp_dir\" && git sparse-checkout set '" + pluginId + "' 2>/dev/null && mkdir -p '" + pluginDir + "' && cp -r \"$temp_dir/" + pluginId + "/.\" '" + pluginDir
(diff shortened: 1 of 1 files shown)


ItsLemmy:
But there should not be any settings.json versionned on git...

DzmingLi:
My change doesn't touch the git/versioning side at all — I only added a backup/restore step for the user's local settings.json around the existing cp -r call.

ItsLemmy:
My comment still stand, plugins are not supposed to ship settings.json  as they are user settings (and not default settings). So there is no need to backup and restore settings.json on update - as the copy will not overwrite the user file.

We did numerous plugins update in the last months and no one opened any bug report about disappearing settings.

That being said I just reviewed the plugins repo and found 3 plugins that did ship settings.json, I fixed those...

About your PR, I'm fine adding a bit more safety, but maybe we could just   rm -f ```"$temp_dir/$pluginId/settings.json"```   before the cp -r ?

This way we are 100% certain we dont overwrite any user file - if let's say someone is using a plugin from an non official repo that decided to version settings.json.

DzmingLi:
Makes sense, closing.

PR #2119, opened 2026-03-08
feat(nix): add pluginPackages for declarative plugin management

DzmingLi:
# Pull Request

## Motivation
The home-manager module manages `plugins.json` and `pluginSettings`, but actual plugin files (QML, assets) still require runtime git download. Users want declarative management of plugin files via nix derivations, so that plugins can be pinned and reproduced across systems.

This adds a `pluginPackages` option (`attrsOf (either package path)`) that symlinks plugin derivations into `~/.config/noctalia/plugins/{name}/`, with `settings.json` excluded to keep runtime settings mutable. Plugins with a valid `manifest.json` are auto-discovered by `PluginRegistry.scanPluginFolder()` — no registration logic needed.

Example usage:
```nix
programs.noctalia-shell = {
  pluginPackages = {
    catwalk = pkgs.callPackage ./plugins/catwalk { };
  };
};
```

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring

## Related Issue
- N/A

## Testing
- [x] Tested on niri
- [ ] Tested on Hyprland
- [ ] Tested on sway

Create a test derivation with plugin files (including `manifest.json`), add to `pluginPackages`, run `home-manager switch`. Verify symlinks appear in `~/.config/noctalia/plugins/{name}/`, `settings.json` is not symlinked, and the plugin is discoverable by the shell.

## Screenshots / Videos
N/A

## Checklist
- [x] Self-reviewed my code

## Additional Notes
- Only `nix/home-module.nix` is changed
- `pluginSettings` for the same plugin still works (mutable `settings.json`)
- Systemd restart triggers are added so the shell restarts when plugin packages change

### DIFF
--- a/nix/home-module.nix
+++ b/nix/home-module.nix
@@ -179,6 +179,27 @@ in
         or filepath, to be written to ~/.config/noctalia/plugins/plugin-name/settings.json.
       '';
     };
+
+    pluginPackages = lib.mkOption {
+      type = with lib.types; attrsOf (either package path);
+      default = { };
+      example = lib.literalExpression ''
+        {
+          catwalk = pkgs.callPackage ./plugins/catwalk { };
+        }
+      '';
+      description = ''
+        Declarative plugin packages. Maps plugin names to derivations or paths
+        containing plugin files (QML, assets, manifest.json).
+
+        Each entry is symlinked into ~/.config/noctalia/plugins/{name}/ with
+        settings.json excluded so that runtime settings remain mutable.
+
+        Plugins are discovered automatically by the shell's PluginRegistry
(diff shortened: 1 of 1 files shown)


Ly-sec:
CC: @bokicoder

bokicoder:
So the user needs to create a derivation for the plugin first, and then use this option to symlink the plugin into `~/.config/noctalia/plugins/`.

Not every user can create a derivation, which means packaging. Even if they could, why don't users just symlink the plugin files into `~/.config/noctalia/plugins/` instead?

DzmingLi:
Makes sense. Closing.

Issue #1893, opened 2026-02-19
App2unit launch method issue


AuTune:
### Description
The app2unit toggle in Noctalia seems to be running non-terminal applications as direct commands from the exec line of the .desktop file, but that is causing mislabeling of the units.

### Steps to Reproduce
My example is with tidal-hifi, the music app I use. On lines 567-572 from ApplicationProvider.qml, I found that non-terminal applications are having just their exec lines directly executed when launched. The equivalent app2unit command would be:

```
app2unit -- env -u GIO_LAUNCHED_DESKTOP_FILE /opt/tidal-hifi/tidal-hifi
```

And when I look for the units, they look like:

```
systemctl --user list-units 'app-Hyprland-*' --no-pager | grep -E 'env|tidal'
app-Hyprland-env@4bb2b10d.service        loaded active running env
app-Hyprland-env@54e6bd6e.service        loaded active running env
```

The processes are labelled both as "env" with no distinguishing between them. However, when running the app2unit command structure from Noctalia intended for *non-terminal* applications on tidal-hifi works much better:

```
app2unit -- tidal-hifi.desktop
```
```
systemctl --user list-units 'app-Hyprland-*' --no-pager | grep -E 'env|tidal'
app-Hyprland-env@4bb2b10d.service           loaded active running env
app-Hyprland-tidal\x2dhifi@f2459938.service loaded active running TIDAL Hi-Fi - TIDAL Hi-Fi
```

The units are much more clearly labelled now.  It seems much better to just handle all app2unit launches as the entire .desktop file instead of the current method of delineating between terminal and non-terminal applications like this. My current workaround is a custom launch command, but perhaps this revision could be considered for the default behavior of the app2unit toggle.

- Distro: CachyOS
- Compositor: Hyprland
- Noctalia-shell Version: v4.5.0
- Monitor resolution: 5120x1440@240hz

notiant:
If it works better if you use "app2unit --" as a custom launch prefix, it might be better to remove the app2unit toggle completely from Noctalia. The recommended method in the docs is "systemd-run --user --scope --collect" anyway.

Ly-sec:
> If it works better if you use "app2unit --" as a custom launch prefix, it might be better to remove the app2unit toggle completely from Noctalia. The recommended method in the docs is "systemd-run --user --scope --collect" anyway.

If that's the case then yes, we can gladly remove the app2unit toggle completely. Let's see what @AuTune says :).

[...]

PR #2163, opened 2026-03-13
Remove app2unit in launcher settings

notiant:
This removes the app2unit toggle from the launcher settings since there are problems with it (#1893) and it has been made redundant by the custom launch prefix option.

2 i18n entries also can be removed.


Ly-sec:
Hey @notiant , thank you veryuch! I totally forgot to check it yesterday, sorry about that.
I'll be merging it

PR #2176, opened 2026-03-14
Respect lockOnSuspend in all suspend cases

### LINKED ISSUE #2036: Lock on suspend does not seem to work
### Description
When I press the power button or close my laptop lid, the laptop goes on sleep but does not lock the screen. Opening the lid or pressing a key gives the previous unlocked session

### Steps to Reproduce
1. Go to Settings > Lock Screen > Behaviour 
2. Turn on 'Lock on suspend'
3. Suspend the computer
4. Resume the computer: the computer is not locked

### Expected Behavior
I should have a lock screen on resuming

### Environment
- Distro:  Arch
- Compositor: niri
- Noctalia-shell Version: 4.6.1
- Monitor resolution: 1920x1080

### Additional Context
noctalia shell is run through config.kdl with 'spawn-at-startup "qs" "-c" "noctalia-shell"'

Aitor42:
# Pull Request

## Motivation
Fix lock-on-suspend not working when suspend is triggered externally (power button, lid close). Previously, only user-initiated suspend from the session menu would properly lock the screen before suspending. External suspend events bypassed the lock on suspend setting.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring

## Related Issue
- Closes #2036

## Testing

- [x] Tested on niri
- [ ] Tested on Hyprland
- [ ] Tested on sway

## Screenshots / Videos

## Checklist
- [x] Self-reviewed my code

## Additional Notes

### DIFF
--- a/Services/Compositor/CompositorService.qml
+++ b/Services/Compositor/CompositorService.qml
@@ -562,6 +562,50 @@ Singleton {
     }
   }
 
+  // Lock screen before external suspend (power button, lid close) by
+  // monitoring systemd-logind's PrepareForSleep D-Bus signal.
+  Process {
+    id: sleepMonitor
+    running: true
+    command: [
+      "gdbus", "monitor", "--system",
+      "--dest", "org.freedesktop.login1",
+      "--object-path", "/org/freedesktop/login1"
+    ]
+
+    stdout: SplitParser {
+      onRead: line => {
+        if (line.indexOf("PrepareForSleep") !== -1 && line.indexOf("true") !== -1)
+          root._lockIfNeeded("PrepareForSleep");
+      }
+    }
+
+    onExited: () => sleepMonitorRestart.start()
+  }
+
+  Timer {
+    id: sleepMonitorRestart
+    interval: 2000
+    onTriggered: sleepMonitor.running = true
+  }
+
+  // Fallback: lock on resume via time-jump detection if D-Bus signal was missed.
+  Connections {
+    target: Time
+    function onResumed() {
+      root._lockIfNeeded("resume fallback");
+    }
+  }
+
+  function _lockIfNeeded(source) {
+    if (!Settings.data.general.lockOnSuspend)
+      return;
+    if (!PanelService || !PanelService.lockScreen || PanelService.lockScreen.active)
+      return;
+    Logger.i("Compositor", "Locking screen (" + source + ")");
+    root.lock();
+  }
+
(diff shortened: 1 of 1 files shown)


Ly-sec:
Hey @Aitor42 ,this sounds more like a thing that should be handled by your compositor.
See (for niri):
https://github.com/niri-wm/niri/wiki/Configuration:-Input#disable-power-key-handling and https://github.com/niri-wm/niri/wiki/Configuration:-Switch-Events

As far as I'm aware hyprland does support the exact same thing.

Aitor42:
You are right this can be configure manually, but seeing that many people think this is more like a "bug" I decided to give it a try and apply it in directly in the shell-level. 
Also the PrepareForSleep D-Bus signal is compositor-agnostic and catches all suspend sources (power key, lid, systemctl, idle, etc.), which is the same approach GNOME/KDE use at the DE level. This way the existing lock on suspend toggle works on all cases without the need of changing the configuration manualy.

Ly-sec:
> You are right this can be configure manually, but seeing that many people think this is more like a "bug" I decided to give it a try and apply it in directly in the shell-level. Also the PrepareForSleep D-Bus signal is compositor-agnostic and catches all suspend sources (power key, lid, systemctl, idle, etc.), which is the same approach GNOME/KDE use at the DE level. This way the existing lock on suspend toggle works on all cases without the need of changing the configuration manualy.

It only catches things that actually rely on it - which is a few compositors. This is not a one solution rule all kinda thing sadly enough. If every compositor would follow wayland protocols I'd agree but that's just not the case.

Aitor42:
> If every compositor would follow wayland protocols

Correct me if I'm wrong, but noctalia requires a wayland composition, right?

> It only catches things that actually rely on it

Yes, it only catches some cases, that's why it's an addition to the current functionality, the goal is to detect and manage this extra cases that are not contemplated in the current funcionality. If the user has configure the compositor to do this funcionality this is complementary and they can work together.

But I see your point, it's okay if you see it not neccesary to implement it and keep this simple.

urwrstkn8mare:
@Aitor42 is it possible for this to be made into a plugin? This would be very useful if I want to put my laptop to sleep via something other than Niri or Noctalia (i.e. directly via systemd).

gammons:
I think I found the bug, and now lock + suspend works for me when I applied my change locally - #2309

PR #2183, opened 2026-03-15
Fix kitty predefined selection colors

DFSko:
## Summary
Fix the predefined Kitty template to use terminal selection tokens instead of foreground and cursor text tokens.

## Problem
For predefined color schemes, kitty-predefined.conf exported selection_foreground from terminal_cursor_text and selection_background from terminal_foreground.

This produced incorrect selection colors. For example, with the ADW scheme the terminal selection background is defined as blue (#3584e4), but the generated Kitty config used white.

## Fix
Use the correct predefined terminal tokens:
- terminal_selection_fg
- terminal_selection_bg

## Verification
Reproduced with ADW.json where terminal.selectionBackground is #3584e4 and confirmed the old Kitty export incorrectly generated selection_background #ffffff.


ItsLemmy:
thanks!

PR #2199, opened 2026-03-17
feat(AudioPanel): volume change on mouse wheel

koffydrop:
# Pull Request

## Motivation
Previously you could not adjust volumes with the mouse wheel on the volume widget, despite being able to in the control center, which annoyed me to no end.
This works on inputs, outputs and, per application audio.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring

## Testing
Just regular day-to-day use.

- [x] Tested on niri
- [ ] Tested on Hyprland
- [ ] Tested on sway

## Screenshots / Videos

sliders

## Checklist

- [x] Self-reviewed my code

## Additional Notes

I just copied and adapted the code from [AudioCard](https://github.com/noctalia-dev/noctalia-shell/blob/f6d398eccc7a5e7588d10a2e99bd51cb8def8a1f/Modules/Cards/AudioCard.qml#L185). Not sure if there's more that needs to be considered.

I'm also testing it on version 4.5.0 which is the one packaged in nixpkgs at the moment, but I made sure to check it does work on main with noctalia-qs.

This introduces this warning, which I'm not sure how to fix without breaking functionality.
```
WARN scene: QML MouseArea at @Modules/Panels/Audio/AudioPanel.qml[674:25]: Detected anchors on an item that is managed by a layout. This is undefined behavior; use Layout.alignment instead.
```

### DIFF
--- a/Modules/Panels/Audio/AudioPanel.qml
+++ b/Modules/Panels/Audio/AudioPanel.qml
@@ -288,6 +288,24 @@ SmartPanel {
                     onPressedChanged: function (pressed) {
                       localOutputVolumeChanging = pressed;
                     }
+
+                    MouseArea {
+                      anchors.fill: parent
+                      hoverEnabled: true
+                      acceptedButtons: Qt.NoButton
+                      propagateComposedEvents: true
+
+                      onWheel: wheel => {
+                                 if (AudioService.sink) {
+                                   const delta = wheel.angleDelta.y || wheel.angleDelta.x;
+                                   const step = Settings.data.audio.volumeStep / 100.0; // Convert percentage to 0-1 range
+                                   const increment = delta > 0 ? step : -step;
+                                   const maxVolume = Settings.data.audio.volumeOverdrive ? 1.5 : 1.0;
+                                   const newValue = Math.max(0, Math.min(maxVolume, localOutputVolume + increment));
+                                   localOutputVolume = newValue;
+                                 }
+                               }
+                    }
                   }
 
                   NText {
@@ -363,6 +381,24 @@ SmartPanel {
                     }
                     onPressedChanged: function (pressed) {
                       localInputVolumeChanging = pressed;
+                    }
+
+                    MouseArea {
+                      anchors.fill: parent
+                      hoverEnabled: true
+                      acceptedButtons: Qt.NoButton
+                      propagateComposedEvents: true
+
+                      onWheel: wheel => {
+                                 if (AudioService.source) {
+                                   const delta = wheel.angleDelta.y || wheel.angleDelta.x;
+                                   const step = Settings.data.audio.volumeStep / 100.0; // Convert percentage to 0-1 range
+                                   const increment = delta > 0 ? step : -step;
+                                   const maxVolume = Settings.data.audio.volumeOverdrive ? 1.5 : 1.0;
+                                   const newValue = Math.max(0, Math.min(maxVolume, localInputVolume + increment));
+                                   localInputVolume = newValue;
+                                 }
+                               }
                     }
                   }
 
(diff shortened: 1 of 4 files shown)


ItsLemmy:
This is not quite right, the right approach would be to add a onWheel signal in NSlider and use that appropriately.
Let me know if you wanna tackle this, or I'll do it when I got some time

koffydrop:
I'll give it shot, if I can't figure it out I'll let you know

koffydrop:
@ItsLemmy how does this look?

I'm not sure why jumping from 4.5.0 to 4.6.8 makes the sink and source volumes not follow the volume step if it changes too fast, but application audio works as expected
sliders

ItsLemmy:
I don't know either, all I know is PW does not like to receive a bunch of volume changes in a small time frame, as they have some quick fade implemented somewhere. That's why we usually debounce calls.

koffydrop:
Gotcha. In that case this should be good for review, unless you prefer if I squash into a single commit

Modules/Panels/Audio/AudioPanel.qml
```diff
@@ -634,6 +654,16 @@ SmartPanel {
...
                         }
+                        onWheel: function (event) {
+                          if (appBox.nodeAudio && appBox.modelData && appBox.modelData.ready === true) {
+                            const delta = event.angleDelta.y || event.angleDelta.x;
+                            const step = Settings.data.audio.volumeStep / 100.0; // Convert percentage to 0-1 range
+                            const increment = delta > 0 ? step : -step;
+                            const maxVolume = Settings.data.audio.volumeOverdrive ? 1.5 : 1.0;
+                            const newValue = Math.max(0, Math.min(maxVolume, appBox.nodeAudio.volume + increment));
+                            appBox.nodeAudio.volume = newValue;
+                          }
```
ItsLemmy, line 737: In onMoved we also call setAppStreamVolume. Would need to also call here

```  if (key)
    AudioService.setAppStreamVolume(key, value);```

ItsLemmy:
- Would be better to move the MouseArea to be a top-level child of the Slider root (after the handle), not inside background.  When on the background  the knob wont react to wheel events.

- For the Layout warning, you can try  ```y: (parent.height - height) / 2``` instead of  anchors.verticalCenter:
  parent.verticalCenter.
  
- Would be awesome if you could also edit AudioCard  to use that new feature, so we avoid keeping a  separate custom implementations.

koffydrop:
The layout warning actually goes away after moving the MouseArea to NSlider, and `anchors.fill: parent` now makes it work like I originally intended.
Should I still prefer using `y: (parent.height - height) / 2` or remove the more manual anchor and size properties in favor of `anchors.fill: parent`?

ItsLemmy:
No this was only a workaround to avoid the warning, if it's gone keep it simple with the original syntax.

Modules/Panels/Audio/AudioPanel.qml
koffydrop: Fixed in https://github.com/noctalia-dev/noctalia-shell/pull/2199/commits/f0a26b7d489dbe13a70d1b89f7eb1bc45600b3d7

koffydrop:
> I'm not sure why jumping from 4.5.0 to 4.6.8 makes the sink and source volumes not follow the volume step if it changes too fast, but application audio works as expected

I used the local*VolumeChanging variables and a timer to reset them after 100ms without wheel input, so now the text should display the correct volume before syncing to pipewire.

Also with a volume step of 1 sometimes the sliders don't register a change due to float precision, and makes it impossible to stop at certain values without closing their respective panels first. I rounded the checks to account for that, hope that's alright.

If you can think of other sliders that could use this I can look into them as well.

koffydrop:
Testing now on 4.7.1 day-to-day, no apparent issues so far

koffydrop:
Rebased with main

ItsLemmy:

Cheers!

koffydrop:
I'll have a look once it lands for me ^^

PR #2205, opened 2026-03-17
Added min/max events in audio service to notify subscribers

nZo-sp:
# Pull Request
Added two new events in AudioService `volumeAtMaximum` and `volumeAtMinimum` to notify subscriber when audio limit is reached. In this way, the user has a visual feedback even when there are no changes in the audio volume.

## Motivation
When using hotkeys to change volume audio and the volume is already at its maximum or minimum, the user does not have any clue if the command has executed or not. With this PR, the OSD with the current volume value pops out.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring

## Related Issue
- None

## Testing
- [x] Tested on niri
- [ ] Tested on Hyprland
- [ ] Tested on sway

## Checklist
- [x] Self-reviewed my code

### DIFF
--- a/Modules/Bar/Widgets/Volume.qml
+++ b/Modules/Bar/Widgets/Volume.qml
@@ -73,6 +73,28 @@ Item {
         externalHideTimer.restart();
       }
     }
+
+    function onVolumeAtMaximum() {
+      if (!firstVolumeReceived) {
+       firstVolumeReceived = true;
+      } else {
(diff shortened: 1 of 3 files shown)


Ly-sec:
Hey @nZo-sp , that's very reasonable, thank you for the PR. I'll be merging this one.

PR #2221, opened 2026-03-19
Added scroll with keyboard in searchable dropdown

nZo-sp:
# Pull Request

Added scroll with keyboard on dropdown

## Motivation
Searchable dropdown are usually used when there are a lot of items. If the user want to try differente values currently he has to:
- open the dropdown
- search the value
- select
- repeat again

With this PR the user is able to change values using the keyboard selecting the highlighted value

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring

## Testing

- [x] Tested on niri
- [ ] Tested on Hyprland
- [ ] Tested on sway

## Screenshots / Videos

## Checklist
- [x] Self-reviewed my code

### DIFF
--- a/Widgets/NSearchableComboBox.qml
+++ b/Widgets/NSearchableComboBox.qml
@@ -241,67 +241,40 @@ RowLayout {
           onTextChanged: root.searchText = text
           fontSize: Style.fontSizeS
 
-          Keys.onUpPressed: {
-            if (listView.currentIndex > 0) {
-              listView.currentIndex--;
-              listView.positionViewAtIndex(listView.currentIndex, ListView.Contain);
-
-              if (root.selectOnNavigation) {
-                selectHighlighted()
-              }
-            }
-          }
-
-          Keys.onDownPressed: {
-            if (listView.currentIndex < listView.count - 1) {
-              listView.currentIndex++;
-              listView.positionViewAtIndex(listView.currentIndex, ListView.Contain);
-
-              if (root.selectOnNavigation) {
-                selectHighlighted()
-              }
-            }
-          }
-
-          Keys.onReturnPressed: {
-            selectHighlighted()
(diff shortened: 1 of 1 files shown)


ItsLemmy:
Cool wanted to that for a while.

But we should no longer hardcode Up and Down like that. We gotta use the user's preferred keybinds via Settings.data.general.keybinds.

And would need to be tested when the settings are shown in a separate window, which adds some complexity.

nZo-sp:
> Cool wanted to that for a while.
> 
> But we should no longer hardcode Up and Down like that. We gotta use the user's preferred keybinds via Settings.data.general.keybinds.
> 
> And would need to be tested when the settings are shown in a separate window, which adds some complexity.

Using as example LockScreen i used this `Keybinds.checkKey` with `enter | escape | up | down`. There are no settings for page-up and down so i entirely removed the section.
About the "separate window" thing, how can I test it? Is there a place somewhere that opens a separate window with a SearchableBox or I have to create a sandbox to test?

ItsLemmy:
image

Then close and reopen settings.

nZo-sp:
Testing performed
- Added secondary keybindings for "Up" and "Down" navigation.
- Verified that both primary and secondary keybindings work correctly.
- Confirmed consistent behavior in both attachedmode and detachedmode.

Observations
- The "autoselect" mode is intentionally limited to explicit keyboard navigation. It does not trigger when a user types in the search box. Enabling this for every keystroke would cause excessive, unnecessary changes for each character entered.

- In the font selection search box, selected items are correctly moved into the viewable area. However, a minor layout issue persists: if a specific font selection reduces the overall size of the scroll area, the item may occasionally fall outside the visible bounds.

ItsLemmy:
Thanks!

Issue #2216, opened 2026-03-19
MainScreen panels and desktop widgets always damage entire screen


ghost:
### Description
As hinted at in #2215…
Every time panel content is changed, the *entire* screen is damaged. Same for the desktop widgets. This is bad because especially on bandwidth-starved laptop GPUs, that makes rendering the shell extremely expensive. On my older laptop, I can see this in the form of both the compositor and Quickshell almost maxing out the processor/iGPU power budget while redrawing part of a panel at screen refresh rate, and `perf top` showing the kernel’s `copy_page_range` pinned to the first position. It also means Performance Mode has almost no effect.

### Steps to Reproduce
Run noctalia-shell and interact with a panel. Niri has a debug shortcut that paints damaged regions with a red tint.

### Expected Behavior
Damaged screen area limited to changed region.

### Environment
- Distro: openSUSE Tumbleweed
- Compositor: Niri (Git snapshot newer than 25.11)
- Noctalia-shell Version: 5547a2716
- Noctalia QS Version: fd50060 (2 commits after 0.0.9)
- Monitor resolution: 1920x1080

ItsLemmy:
I'm aware of the full screen thing and it's implication, the problem is the unified/attached design we have going on.
I'll see if I can refactor and split the panel content in separate smaller PanelWindow but it's medium/large undertaking.

ghost:
Another problem is that QS isn’t treated as a candidate for direct scanout, but I guess that may be a problem with Smithay/Niri and QtWayland, and possibly the GPU driver.

ghost:
In fact, the shell permanently covering the entire screen area completely prevents direct scanout even for clients where it normally works! Oops…

ghost:
#2225 addresses that last concern at least.

ghost:
Don’t take this the wrong way but was putting everything on a single surface as an “optimization” an LLM suggestion? It seems like something it’d do.

Because so far it:

* prevents direct scanout
* prevents caching
* prevents damage tracking
* negates performance benefits of a threaded scene graph renderer that uses independent surfaces for each element

ItsLemmy:
Nope, it wasn't. I arrived at it while developing the unified background/shadow system after a few failed attempts,  it was the best approach I could come up with at the time. I had no prior Wayland or Qt experience before Noctalia.    I always knew it would come back and haunt me :P

[...]

PR #2225, opened 2026-03-19
MainScreen/PanelWindow: do not put on top layer

ghost:
# Pull Request

## Motivation
Having a part of the shell that is always in visible/rendered state sitting on the top layer completely prevents direct scanout of other Wayland clients, resulting in expensive copies and increased latency.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring

## Related Issue
- #2216

## Testing

- [x] Tested on niri
- [ ] Tested on Hyprland
- [ ] Tested on sway

One client that should always have direct scanout is mpv with the gpu-next video output.

Items that *should* go on top/overlay if configured that way, such as OSD and Launcher, still behave as expected. Bar still covers non-fullscreen windows.

## Checklist
- [x] Self-reviewed my code

### DIFF
--- a/Modules/MainScreen/MainScreen.qml
+++ b/Modules/MainScreen/MainScreen.qml
@@ -44,7 +44,9 @@ PanelWindow {
   }
 
   // Wayland
-  WlrLayershell.layer: WlrLayer.Top
+  // As this is always rendered, putting it on the top layer
+  // would prevent direct scanout for other Wayland clients.
+  // WlrLayershell.layer: WlrLayer.Top
   WlrLayershell.namespace: "noctalia-background-" + (screen?.name || "unknown")
   WlrLayershell.exclusionMode: ExclusionMode.Ignore // Don't reserve space - BarExclusionZone handles that
   WlrLayershell.keyboardFocus: {


ashley-paul-12345:
@mia-0 @ItsLemmy
excuse me if i misunderstood this, but won't commenting `WlrLayershell.layer: WlrLayer.Top` have no effect as the layer is by default Top as in https://quickshell.org/docs/master/types/Quickshell.Wayland/WlrLayershell/#layer?

a better a approach would be to completely unload the `PanelWindow` using `LazyLoader` when there's a fullscreen toplevel in the focused workspace, for example in Hyprland something like this:
```qml
LazyLoader {
  active: !Hyprland.focusedWorkspace?.toplevels.values.some(t => t.wayland?.fullscreen)

  component: PanelWindow {
  ...
}
```
I don't use Niri or Sway, so I honestly don't know how can this be done for them
This will also stop the full layer damage in case there's a fullscreen toplevel in the focused workspace, resulting in freeing more of the gpu, and I think that should fix the `direct_scanout` in Hyprland

ghost:
I may have disabled the OSD when I made the PR so it looked to me like it was doing something. Oops! The later revert of the change that kept the OSD “visible” at all times is what actually helped here.

Thanks for the suggestion though. There are many places that could use some lazy loading/unloading in general; a lot of them will need overhauls of surrounding markup to work…

PR #2230, opened 2026-03-20
Play volume feedback sound when changing volume externally

shouya:
# Pull Request

## Motivation

I have key bindings to change volumes using `wpctl` directly. However, when the volume gets changed this way, the feedback sound does not play despite the volume change OSD shows up. I checked the code and found that the volume change feedback sound is only played when users change the volume through noctalia.

I think this behavior is inconsistent because if noctalia is expected to pick up external volume change, it should playback the feedback sound to indicate the volume change audibly in the same logic that it pops up an OSD as to show the event.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring

## Testing

- [x] Tested on niri
- [ ] Tested on Hyprland
- [ ] Tested on sway

## Screenshots / Videos

## Checklist
- [x] Self-reviewed my code

## Additional Notes

To test it out, run `wpctl set-volume @DEFAULT_AUDIO_SINK@ 0.05+`.

### DIFF
--- a/Services/Media/AudioService.qml
+++ b/Services/Media/AudioService.qml
@@ -14,6 +14,10 @@ Singleton {
   property var lastVolumeFeedbackTime: 0
   readonly property int minVolumeFeedbackInterval: 100
 
+  // Track the last sink that produced volume feedback so we can suppress the
+  // initial onVolumeChanged that fires on startup and device switches.
+  property PwNode _lastFeedbackSink: null
+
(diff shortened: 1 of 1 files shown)


Ly-sec:
Hey @shouya , I just tested it and it seems like it does what it's supposed to. Thank you very much, I'll be merging it.

PR #2238, opened 2026-03-20
LockKeysService: use Qml file system model

ghost:
# Pull Request

Forking shell processes is very very expensive. Replace with a more efficient Qml model.

This should automatically instantiate watchers for hotplugged devices, as FolderListModel uses a file system watcher.

Quickshell’s FileView does not get notified of changes because that’s how sysfs works, but it does handle the content conparison for us. Convenient!

## Motivation
Having the service enabled pinned my laptop’s CPU usage to a constant 15% (single core) and consumed a lot of memory bandwidth due to all the process forking.

## Type of Change
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Refactoring

## Testing
Looked at the logs and pushed some keys.

## Checklist
- [x] Self-reviewed my code

### DIFF
--- a/Services/Keyboard/LockKeysService.qml
+++ b/Services/Keyboard/LockKeysService.qml
@@ -1,6 +1,8 @@
 pragma Singleton
 
 import QtQuick
+import QtQml.Models
+import Qt.labs.folderlistmodel 2.10
 import Quickshell
 import Quickshell.Io
 import qs.Commons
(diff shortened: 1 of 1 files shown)


Ly-sec:
Hey @mia-0 , this is indeed a much better solution. We'll be merging it, thank you once again for another well done PR!

ghost:
Filed a related upstream issue: https://github.com/quickshell-mirror/quickshell/issues/655

PR #2253, opened 2026-03-21
fix(NFilePicker): Added multiSelection choice

spiros132:
# Pull Request
Added the choice to have a `NFilePicker` where you can choose multiple items at the same time

## Motivation
The `NFilePicker` component currently only supports choosing one item but I believe it would be good for the developers to have the choice for multiple items to be chosen at the same time. The boolean to enable multiple choices at the same time is false by default since that's how it has been working so far. Right now you can select multiple items with the Shift button.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Refactoring

## Related Issue
No related issue found.

## Testing
The testing was done by using it, opening multiple file pickers and seeing that it works. 

- [x] Tested on niri
- [ ] Tested on Hyprland
- [ ] Tested on sway

## Screenshots / Videos
A screenshot showing multiple folders being selected at the same time. 
image

## Checklist
- [x] Self-reviewed my code

## Additional Notes
This PR only adds support for selecting multiple files / folders with holding the shift button. Some people are used to having the shift button select from the first item in a list to the last item in the list, so in that case maybe ctrl might have been better but I'm not really sure tbh.

Also the code is very repetitive, in that case I can refactor it to put everything under one function instead to make it less repetitive if that's wished for.

### DIFF
--- a/Widgets/NFilePicker.qml
+++ b/Widgets/NFilePicker.qml
@@ -168,6 +168,55 @@ Popup {
                       }
                     }
 
+    // Function when an item is clicked, either a folder or a file since both have the same functionality, reduces repetitive code
+    function itemClicked(modifiers, path) {
+      var list = currentSelection.slice();
+      const index = list.indexOf(path);
+      if (root.allowMultiSelection && (modifiers & Qt.ShiftModifier) && list.length > 0) {
+        if (index > -1) {
+          list.splice(index, 1);
+          currentSelection = list;
+        } else {
+          var i = 0;
+          var toggle = false;
+          const firstItemPath = list[0];
+          while (i < filteredModel.count) {
+            const itemPath = filteredModel.get(i).filePath;
+
+            // This should be called twice, when we get to the item selected and when we get to the starting item.
+            if (itemPath === firstItemPath || itemPath === path) {
+              toggle = !toggle;
+            }
+
+            // Add all files between the starting item and the item selected.
+            if (toggle) {
+              if (!list.includes(itemPath)) {
+                list.push(itemPath);
+              }
+            }
+
+            i++;
+          }
+
+          // Add the path selected as well since it's skipped in the while loop
+          if (!list.includes(path)) {
+            list.push(path);
+          }
+          currentSelection = list;
+        }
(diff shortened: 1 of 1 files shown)


Widgets/NFilePicker.qml
```diff
@@ -677,13 +704,39 @@ Popup {
                              if (model.fileIsDir) {
                                // In folder mode, single click selects the folder
                                if (root.selectionMode === "folders") {
-                                 filePickerPanel.currentSelection = [model.filePath];
+                                 // Deep copy otherwise the isSelected might not get updated.
+                                 if (root.allowMultiSelection && (mouse.modifiers & Qt.ShiftModifier)) {
+                                   var list = JSON.parse(JSON.stringify(filePickerPanel.currentSelection));
```
Ly-sec: We think it would be better to replace this with something like `var list =  filePickerPanel.currentSelection.slice() ` for performance reasons, less json, less computing and all that.

Widgets/NFilePicker.qml
```diff
@@ -677,13 +704,39 @@ Popup {
...
+                                 }
                                }
                                // In file mode, single click on folder does nothing (must double-click to enter)
                              } else {
                                // Single click on file selects it (only in file mode)
                                if (root.selectionMode === "files") {
-                                 filePickerPanel.currentSelection = [model.filePath];
+                                 if (root.allowMultiSelection && (mouse.modifiers & Qt.ShiftModifier)) {
+                                   // Deep copy otherwise the isSelected might not get updated.
+                                   var list = JSON.parse(JSON.stringify(filePickerPanel.currentSelection));
```
Ly-sec: same thing here

Widgets/NFilePicker.qml
spiros132: Thank you, I didn't know that was an option!

spiros132:
@Ly-sec I've reworked it a bit to make it easier to read, reduce the repetitive code and added so that shift can select multiple files at the same time.

Ly-sec:
Awesome @spiros132 this is much nicer now. Going to merge it, thank you very much!

PR #2271, opened 2026-03-23
i18n: add Czech (cs) translation

navik61:
# Pull Request
Adds Czech (cs) translation and registers it in I18n.qml availableLanguages list.


ItsLemmy:
Thanks, if you ever need to fix/add new translations please join https://i18n.noctalia.dev/

PR #2298, opened 2026-03-25
Dismiss toast ipc call

tmarti2:
This is a simple MR which adds an IPC call to dismiss toasts. In my case I used the same keybind than notifications dismiss and it seems to work: `spawn-sh "noctalia notifications dismissAll && noctalia toast dismiss";`

I noticed that using IPC call to dismiss notifications does not trigger the `animateOut` animation from `Notification.qml`, I'm not sure it is intended (?) For now I did not used `hideImmediately` for toast because it looks better to have the animation, but maybe I should for better consistency with notifications ?

This PR updates the doc for toast IPC calls: https://github.com/noctalia-dev/noctalia-docs/pull/106

### DIFF
--- a/Modules/Toast/ToastScreen.qml
+++ b/Modules/Toast/ToastScreen.qml
@@ -32,6 +32,10 @@ Item {
                           "timestamp": Date.now()
                         });
     }
+
+    function onDismiss() {
+      root.dismissToast();
+    }
   }
 
   // Clear queue on component destruction to prevent orphaned toasts
@@ -40,6 +44,13 @@ Item {
     isShowingToast = false;
     hideTimer.stop();
     quickSwitchTimer.stop();
+  }
+
+  function dismissToast(){
+    if (windowLoader.item) {
+      hideTimer.stop();
+      windowLoader.item.hideToast();
+    }
   }
 
   function enqueueToast(toastData) {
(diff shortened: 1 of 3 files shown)


Ly-sec:
Hey @tmarti2 this PR (and also the docs one) looks good to me. Now that you mentioned that notifications use `hideImmediately`... maybe we should remove that one too so it looks coherent to the normal look & feel. Would you mind changing that (obviously gotta test if it causes some issues)?

tmarti2:
@Ly-sec I wanted to do that but I'm not sure it is as simple as toasts. For the oldest ipc call we just do `popupModel.remove(index);` and for dismissAll we call a dismiss function (which probably comes from quickshell ?) and then destroy/clear notifications lists.

Edit: Maybe I did not understood what you meant, do you mean using immediately for toasts or normal hide for notifications ?

Ly-sec:
@tmarti2 no I do think you understood it correctly - not using hide mmediately for notifications but the more I think of it, we can just use hide immediately for toast & notifications, should be fine for an IPC call :)

tmarti2:
@Ly-sec Ok, I changed back to immediate for toasts, it should be good.

Ly-sec:
Awesome @tmarti2 , I'll merge it. Thank you very much :heart: !

PR #2312, opened 2026-03-26
Rework Airplane Mode handling & remove bluetoothctl fallback

notiant:
1. This PR improves the Airplane Mode handling which has been made possible by using wifiEnabled from Quickshell.Networking. `airplaneModeEnabled` is now not longer stored in the settings file.
2. The bluetoothctl fallback process has been removed since it's unnecessary. The fallback is only called if `!adapter` but in this case Noctalia doesn't show any devices, so the logic isn't very sound. This has been tested by @turannul and doesn't seem to cause any problems. This also removes the constantly polling `ctlShowProcess`.
3. Some reordering in BluetoothService to create a better structure.

### DIFF
--- a/Services/Networking/NetworkService.qml
+++ b/Services/Networking/NetworkService.qml
@@ -678,7 +678,7 @@ Singleton {
           }
 
           let enhancedBand = band;
-          if (channel && width) {
+          if (channel && width && width !== "0 MHz") {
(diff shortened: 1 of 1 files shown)


Services/Networking/NetworkService.qml
turannul: && width !== "0 MHz" Additional condition to avoid displaying invalid data

PR #2315, opened 2026-03-27
fix(audio): filter out intermediate stream nodes

ern775:
# Pull Request
## Motivation
Previously, custom pipewire filter streams were being treated as application streams. This meant that only the filter stream would be shown in the volume widget since there weren't any checks for streams connected to streams.

This pr adds a check for whether the output stream is virtual or not, and if it is, treats is as an intermediate node.

The main use case for the pipewire filters are sound effects, mixing audio and in my use case, equalizers. This pr should fix this for every filter as they are all always virtual.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring

## Related Issue

## Testing

- [ ] Tested on niri
- [x] Tested on Hyprland
- [ ] Tested on sway

## Screenshots / Videos
My filter setup:
image

Before changes:
image

After changes:
image

## Checklist
- [x] Self-reviewed my code

## Additional Notes

### DIFF
--- a/Services/Media/AudioService.qml
+++ b/Services/Media/AudioService.qml
@@ -189,6 +189,9 @@ Singleton {
     node: root.sink
   }
 
+  // Track all streams globally to prevent binding loops for filtered out streams
+  readonly property var streamNodes: Pipewire.ready ? Pipewire.nodes.values.filter(n => n && n.isStream) : []
+
   // Find application streams that are connected to the default sink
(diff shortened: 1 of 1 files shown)


Ly-sec:
Hey @ern775 , I just did some testing and it seems to work fine (from my limited testing that is, no regressions to be seen).
I'll be merging this one, thank you very much!

PR #2340, opened 2026-03-28
Apply reduced light mode opacity also to dock, notifications & OSD

notiant:
This PR extends the reduced opacity in light mode that applies to panels also to the dock, notifications, toasts and OSD so that this behavior is more consistent. Also changing the light mode opacity from `Math.pow(baseOpacity, 2)` to `Math.pow(baseOpacity, 1.5)`.

### DIFF
--- a/Commons/Color.qml
+++ b/Commons/Color.qml
@@ -324,21 +324,20 @@ Singleton {
     }
   }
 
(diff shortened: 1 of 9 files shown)


Ly-sec:
Hey @notiant that looks good to me, thank you very much!

PR #2343, opened 2026-03-29
fix(plugins): cd into mktemp before git clone to survive dead cwd

Mic92:
## Motivation
If `noctalia-shell`'s working directory is removed while it runs (started from a temp dir, dotfiles manager relinks the launch dir, `cd` into a since-deleted build tree, ...), the plugin fetch/install subprocesses inherit that dead cwd and `git` fails with:

```
shell-init: error retrieving current directory: getcwd: cannot access parent directories: No such file or directory
```

With stderr redirected to `/dev/null` this surfaced only as `Empty response from <source>` — no hint at the actual cause, which made it look like a network/manifest problem.

This PR:
- extracts a small `sparseCloneCmd()` helper shared by fetch and install that `cd`s into the `mktemp` dir before running `git`, so the inherited cwd no longer matters;
- stops discarding stderr and includes it in the error log/callback so future failures are diagnosable.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring

## Related Issue
- n/a

## Testing
- [x] Tested on niri
- [ ] Tested on Hyprland
- [ ] Tested on sway

Reproduced by launching `noctalia-shell` from a directory, removing that directory, then triggering a plugin fetch from the UI. Before: `Empty response from <source>`. After: fetch succeeds; if git itself fails, its stderr is shown in the log/toast.

## Screenshots / Videos
n/a (no visual change)

## Checklist
- [x] Self-reviewed my code

## Additional Notes
Happy to split the stderr-surfacing into a separate PR if you'd prefer the cwd fix on its own.

### DIFF
--- a/Services/Noctalia/PluginService.qml
+++ b/Services/Noctalia/PluginService.qml
@@ -246,6 +246,18 @@ Singleton {
     }
   }
 
+  // Build a shell command that sparse-clones repoUrl into a fresh temp dir,
+  // runs `body` there, then cleans up. We cd into the temp dir immediately so
+  // a stale/deleted process cwd can't break git; stderr is left to the caller
+  // so failures are diagnosable. GIT_TERMINAL_PROMPT=0 prevents hanging on
+  // private repos that would otherwise prompt for credentials.
+  function sparseCloneCmd(repoUrl, body) {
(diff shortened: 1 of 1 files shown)


ItsLemmy:

Cheers!

PR #2380, opened 2026-04-01
MediaPlayerPanel: fix layout issues

notiant:
I noticed some problems with the margins in compact mode in the MediaPlayerPanel, also there is a layout problem in normal mode if the album cover is disabled.


Ly-sec:
Looks good to me, thank you very much @notiant !

PR #2397, opened 2026-04-04
feat(idle): add lock-screen-off timeout setting

kevclark:
# Summary
- Adds a new lockScreenOffTimeout setting that turns off monitors after a configurable period of inactivity while the lock screen is active
- Defaults to 0 (disabled) so existing users are unaffected
- Adds a dedicated IdleMonitor that is created when the lock screen activates and destroyed when it deactivates
- New setting appears in the Idle → Behaviour tab between Lock screen and Suspend rows (no settings gear needed for this stage)

# Motivation
When a laptop or desktop locks, the monitors remain on indefinitely unless a separate screen-off timeout was already reached. This adds a dedicated post-lock screen-off stage, useful for privacy (screen blanked quickly after locking) and power saving.

Before moving to Noctalia I used swayidle to handle the post-lock screen-off:
```
  spawn-at-startup "swayidle" "-w" \
      "timeout" "600" "niri msg action power-off-monitors" \
      "timeout" "610" "locker.sh" \
      "timeout" "10" "if pgrep -x gtklock; then niri msg action power-off-monitors; fi" \
      "lock" "locker.sh" \
      "before-sleep" "locker.sh"
```
  Where locker.sh is a small guard script to prevent double-invocation:
```
  #!/usr/bin/env sh
  if ! pgrep -x gtklock; then
      gtklock -d
  fi
```
The timeout 10 stage — checking if the locker is active and blanking the screen — is exactly what this feature replaces natively within noctalia's idle management.

An example use case with a lock timeout of 600s and a screen-off timeout of 610s. The lock screen activates at 600s; the monitors turn off 10s later. If the user wakes the screen from the lock screen (e.g. moves the mouse) but does not unlock the session, the screen-off timeout resets and it takes a full 610s for the monitors to turn off again. With lockScreenOffTimeout set to, say,  60s, the monitors will turn off after just 60s of inactivity on the  lock screen — regardless of how many times the user wakes them.

# Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring

# Testing
- Enable idle management and set a short lockScreenOffTimeout (e.g. 10s) — monitors turn off after that period while locked
- Set lockScreenOffTimeout to 0 — monitors stay on while locked (feature disabled)
- Monitors restore correctly on unlock
- Setting change while lock screen is active updates the monitor immediately
- No regression on existing screen-off, lock, and suspend timeout behaviour
 
## Tested on
- [x] Tested on niri
- [ ] Tested on Hyprland
- [ ] Tested on sway

## Screenshot
noctalia-lock-screen-off

## Checklist
- [x] Self-reviewed my code


dinhokusanagi:
Really nice work on this! The changes are clean and definitely improve the overall experience.

As a suggestion, it would be great to have the option to enable or disable these features individually, instead of only being able to toggle all of them at once. This would add a lot more flexibility for users.

kevclark:
@dinhokusanagi I hear what you're saying, I've just kept the controls the same as @ItsLemmy originally implemented.  Setting any of these to 0 will disable them. Individual toggles is more intuitive but setting such fields to 0 is still fairly clear imho.

dinhokusanagi:
> @dinhokusanagiEu ouço o que você está dizendo, eu apenas mantive os controles iguais aos @ItsLemmyimplementado originalmente. Definir qualquer um deles como 0 irá desativá-los. Alternar individualmente é mais intuitivo, mas definir esses campos como 0 ainda é bastante claro na minha humilde opinião.

I understand, thank you.

ItsLemmy:

Cheers!

Issue #2357, opened 2026-03-30
`Sync System Theme` doesn't follow the current theme.


not-reed:
### Description
`Sync System Theme` doesn't always follow the currently selected theme.

### Steps to Reproduce
1. Go to Settings -> Color Scheme
2. Disable `Dark Mode`, Disable `Sync System Theme`.
3. Open Firefox (as an example, other apps work too) It is in light mode.
4. Enable `System Sync Theme` (expecting this will make the system theme follow the current theme)
5. Click `Dark Mode` to enable dark mode
6. Firefox is still in light mode! (some apps, i.e.e ghostty are now in dark mode)
7. Disable `Sync System Theme`
8. Enable `Sync System Theme` again
9. Firefox is now in Dark Mode too

### Expected Behavior
with `Sync System Theme` being a toggle that I enable, I would expect any changes to light/dark mode to also make changes to the system theme in the future. Current behavior seems to be a one time sync at the time its enabled, not an ongoing hook.

### Environment
- Distro:  NixOS
- Compositor: Niri
- Noctalia-shell Version: 4.7.2-git

Ly-sec:
Hey @not-reed , I just pushed a fix (which solves it on my end), can you please update and let me know if that solves it for you?

not-reed:
Hmm, I am now on v4.7.4-git ([75d67b3](https://github.com/noctalia-dev/noctalia-shell/commit/75d67b3)) and it would seem that the toggle no longer does anything for me, and firefox is forever in dark mode 😄 

Noctalia theme switches, and ghostty theme switches, but not firefox or chrome

Ly-sec:
Hey @not-reed mind giving it another try?

not-reed:
@Ly-sec thanks for the fix, works perfectly!

PR #2406, opened 2026-04-04
GTK refresh fix

notiant:
1. This fixes the inverted color scheme for GTK apps when changing the Noctalia theme. I've looked through the version history and the reason for #2357 was the `force` variable in `pushSystemColorScheme()`, so this should work now.
2. Some margin corrections for panels.


Ly-sec:
Thank you very much @notiant , I'll merge it :).

PR #2412, opened 2026-04-06
Use general reverse scrolling setting for bar

notiant:
I don't see a reason why there should be a separate setting for reverse scrolling for the bar. It makes more sense that the scrolling behavior of the workspace widget and the empty areas of the bar are always the same.

### DIFF
--- a/Assets/settings-default.json
+++ b/Assets/settings-default.json
@@ -70,7 +70,6 @@
       ]
     },
     "mouseWheelAction": "none",
-    "reverseScroll": false,
     "mouseWheelWrap": true,
     "middleClickAction": "none",
     "middleClickFollowMouse": false,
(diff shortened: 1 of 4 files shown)


tibssy:
I originally added the separate reverse scrolling option for the Bar Behavior specifically because of Niri's horizontal content/column scrolling. While standard workspace scrolling usually feels intuitive in one direction, some users might prefer to reverse the scroll direction specifically when navigating columns horizontally in Niri, without affecting how the actual workspace widget behaves.

notiant:
> I originally added the separate reverse scrolling option for the Bar Behavior specifically because of Niri's horizontal content/column scrolling. While standard workspace scrolling usually feels intuitive in one direction, some users might prefer to reverse the scroll direction specifically when navigating columns horizontally in Niri, without affecting how the actual workspace widget behaves.

Yeah, in that case it probably makes sense.

PR #2414, opened 2026-04-06
NComboBox: remove hover animation

notiant:
This removes the fade animation for the hovered item in NComboBox. The current behavior creates a weird effect when moving fast between list items (e.g. in the font selection).


Ly-sec:
Hey @notiant , thanks for the PR. I'll merge this one :).

PR #2422, opened 2026-04-08
feat: add Starship prompt template

adriamartin91:
## Summary

Adds built-in Starship prompt template support to Noctalia, allowing the Starship color palette to update dynamically when switching color schemes.

**New files:**
- `Assets/Templates/terminal/starship.toml` — Wallpaper-based template (MD3 color tokens)
- `Assets/Templates/terminal/starship-predefined.toml` — Predefined colorscheme template (terminal color tokens)

**Modified files:**
- `Services/Theming/TemplateRegistry.qml` — Registered as terminal with dual-mode support
- `Scripts/bash/template-apply.sh` — Post-hook for palette injection

### How it works

Starship doesn't support config includes, so the template uses a two-step approach:

1. Noctalia generates a palette-only TOML to  `~/.cache/noctalia/starship-palette.toml`
2. The post-hook injects it into the user's `starship.toml` between delimiters (`# >>> NOCTALIA STARSHIP PALETTE >>>` /  `# <<< ... <<<`), sets `palette = "noctalia"`, and preserves all other config

The palette provides both standard color names (`blue`, `red`, `green`) and Catppuccin-compatible names (`mauve`, `teal`, `peach`, `lavender`) for broad compatibility with existing Starship configurations.

### Note on color updates

Starship re-reads its config on every prompt render, so no reload signal is needed. After switching color schemes, the new colors will appear on the next prompt (pressing Enter). Previously rendered prompts in the terminal scrollback will keep their original colors — this is a terminal limitation, not specific to this template.


Ly-sec:
Hey @adriamartin91 , thank you very much for the PR, I'll be merging this one :).

PR #2425, opened 2026-04-08
fix(Brightness): Fixes a bug with the bar widget, and also with the minimum enforce

spiros132:
# Pull Request

Fixes a bug with the bar widget where the scroll wheel suddenly stopped working when apply on all monitors was turned on. And also fixed a bug where the minimum for ddc monitors weren't 0.01 correctly. 

## Motivation
As before, this should fix a problem where the ddc monitors could go lower than 1% even if enforce minimum was toggled on. Also refactored to use less repetitive code, and to fix that the bar widget wasn't working with the scroll wheel, when apply on all was turned on.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Refactoring

## Related Issue
No related issues as far as I could see. 

## Testing
On my computer checking that the bar widget worked correctly and that the minimum was enforced correctly as well.

- [x] Tested on niri
- [ ] Tested on Hyprland
- [ ] Tested on sway

## Screenshots / Videos
No screenshots needed here as far as I could tell.

## Checklist
- [x] Self-reviewed my code

## Additional Notes
No additional notes.

### DIFF
--- a/Services/Hardware/BrightnessService.qml
+++ b/Services/Hardware/BrightnessService.qml
@@ -478,7 +478,7 @@ Singleton {
     }
 
     function setBrightness(value: real): void {
-      var min = Settings.data.brightness.enforceMinimum ? 0.01 : 0;
+      var min = Settings.data.brightness.enforceMinimum && isDdc ? 0.01 : 0;
       value = Math.max(min, Math.min(1, value));
       var rounded = Math.round(value * 100);
 


Ly-sec:
Hey @spiros132 thank you very much for the PR, this seems to work :). I'll be merging it!

PR #2426, opened 2026-04-08
fix: show ethernet connection in network widget

### LINKED ISSUE #2385: Ethernet not showing in network widget after latest update
After updating Noctalia Shell to the latest version, the network widget no longer displays Ethernet connections. The widget now only shows Wi-Fi, even when a wired connection is active and working normally.

Expected behavior:
The network widget should detect and display Ethernet connections alongside Wi-Fi, or at least show the active wired connection when connected.

Current behavior:

Only Wi-Fi is shown
Ethernet is completely missing from the widget
System is connected via cable and working normally outside the widget

Additional info:
This appears to be a regression after the recent networking overhaul, not just a missing feature, since Ethernet support was expected or partially present before.

Environment:

Distro: NixOS
Compositor: Hyprland / Wayland
Network backend: NetworkManager
Noctalia-shell Version: v4.7.3

Connect Ethernet cable
Open network widget
Observe that only Wi-Fi is displayed

Image

ultherego:
# Pull Request

## Motivation
Fix Ethernet connection not being displayed in the network widget after recent networking changes. The issue occurs even when a wired connection is active and working correctly.

## Type of Change
- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring

## Related Issue
- Closes #2385

## Testing

- [x] Tested on niri
- [ ] Tested on Hyprland
- [ ] Tested on sway

## Screenshots / Videos
obraz
obraz
obraz

## Checklist
- [x] Self-reviewed my code

## Additional Notes
This change restores Ethernet (wired) connection visibility in the network widget while keeping existing Wi-Fi behavior unchanged.
Handles wired connections from NetworkManager and ensures they are properly reflected in the widget state.

### DIFF
--- a/Modules/Panels/Network/NetworkPanel.qml
+++ b/Modules/Panels/Network/NetworkPanel.qml
@@ -1144,6 +1144,7 @@ SmartPanel {
 
                         NButton {
                           text: I18n.tr("common.disconnect")
+                          visible: VPNService.disconnectingUuid !== modelData.uuid
                           fontSize: Style.fontSizeS
                           backgroundColor: Color.mSurfaceVariant
                           textColor: Color.mOnSurface
@@ -1246,6 +1247,7 @@ SmartPanel {
 
                         NButton {
                           text: I18n.tr("common.connect")
+                          visible: VPNService.connectingUuid !== modelData.uuid
                           fontSize: Style.fontSizeS
                           backgroundColor: Color.mPrimary
                           textColor: Color.mOnPrimary
--- a/Modules/Panels/Network/NetworkPanel.qml
+++ b/Modules/Panels/Network/NetworkPanel.qml
@@ -27,13 +27,13 @@ SmartPanel {
   property string panelViewMode: "wifi"
   property bool panelViewPersistEnabled: false
 
-  // Whether VPN configurations exist (tab only shown when true)
-  readonly property bool vpnAvailable: Object.keys(VPNService.connections).length > 0
-
   // If VPN configs vanish while on VPN tab, switch away
-  onVpnAvailableChanged: {
-    if (!vpnAvailable && panelViewMode === "vpn") {
-      panelViewMode = NetworkService.wifiAvailable ? "wifi" : "ethernet";
+  Connections {
+    target: VPNService
+    function onHasConnectionsChanged() {
+      if (!VPNService.hasConnections && root.panelViewMode === "vpn") {
+        root.panelViewMode = NetworkService.wifiAvailable ? "wifi" : "ethernet";
+      }
     }
   }
 
@@ -42,12 +42,12 @@ SmartPanel {
     target: NetworkService
     function onEthernetAvailableChanged() {
       if (!NetworkService.ethernetAvailable && root.panelViewMode === "ethernet") {
-        root.panelViewMode = NetworkService.wifiAvailable ? "wifi" : root.vpnAvailable ? "vpn" : "wifi";
+        root.panelViewMode = NetworkService.wifiAvailable ? "wifi" : VPNService.hasConnections ? "vpn" : "wifi";
       }
     }
     function onWifiAvailableChanged() {
       if (!NetworkService.wifiAvailable && root.panelViewMode === "wifi") {
-        root.panelViewMode = NetworkService.ethernetAvailable ? "ethernet" : root.vpnAvailable ? "vpn" : "ethernet";
+        root.panelViewMode = NetworkService.ethernetAvailable ? "ethernet" : VPNService.hasConnections ? "vpn" : "ethernet";
       }
     }
   }
@@ -171,7 +171,7 @@ SmartPanel {
       if (NetworkService.ethernetConnected) {
         NetworkService.refreshActiveEthernetDetails();
       }
-      if (vpnAvailable) {
+      if (VPNService.hasConnections) {
         VPNService.refresh();
       }
     } else {
@@ -184,10 +184,10 @@ SmartPanel {
       panelViewMode = "wifi";
     } else if (NetworkService.ethernetAvailable) {
       panelViewMode = "ethernet";
-    } else if (vpnAvailable) {
+    } else if (VPNService.hasConnections) {
       panelViewMode = "vpn";
     } else {
-      panelViewMode = "wifi"; // fallback (shows empty/disabled state)
+      panelViewMode = "ethernet"; // fallback
     }
     panelViewPersistEnabled = true;
   }
@@ -302,7 +302,7 @@ SmartPanel {
               let count = 0;
               if (NetworkService.wifiAvailable) count++;
               if (NetworkService.ethernetAvailable) count++;
-              if (root.vpnAvailable) count++;
+              if (VPNService.hasConnections) count++;
               return count > 1;
             }
             margins: Style.marginS
@@ -332,7 +332,7 @@ SmartPanel {
               text: "VPN"
               tabIndex: 2
               checked: modeTabBar.currentIndex === 2
-              visible: root.vpnAvailable
+              visible: VPNService.hasConnections
             }
           }
         }
@@ -1308,42 +1308,6 @@ SmartPanel {
                       }
                     }
                   }
-                }
-              }
-            }
-
-            // VPN empty state
-            NBox {
-              visible: panelViewMode === "vpn" && Object.keys(VPNService.connections).length === 0
-              Layout.fillWidth: true
-              Layout.preferredHeight: emptyVpnColumn.implicitHeight + Style.margin2M
-
-              ColumnLayout {
-                id: emptyVpnColumn
-                anchors.fill: parent
-                anchors.margins: Style.marginM
-                spacing: Style.marginL
-
-                Item {
-                  Layout.fillHeight: true
-                }
-
-                NIcon {
-                  icon: "shield-off"
-                  pointSize: 48
-                  color: Color.mOnSurfaceVariant
-                  Layout.alignment: Qt.AlignHCenter
-                }
-
-                NText {
-                  text: I18n.tr("tooltips.manage-vpn")
-                  pointSize: Style.fontSizeL
-                  color: Color.mOnSurfaceVariant
-                  Layout.alignment: Qt.AlignHCenter
-                }
-
-                Item {
-                  Layout.fillHeight: true
                 }
               }
             }
--- a/Modules/Panels/Network/NetworkPanel.qml
+++ b/Modules/Panels/Network/NetworkPanel.qml
@@ -37,12 +37,17 @@ SmartPanel {
     }
   }
 
-  // If Ethernet adapter disappears while on Ethernet tab, switch to WiFi
+  // If an adapter disappears while its tab is active, switch to the best available
   Connections {
     target: NetworkService
     function onEthernetAvailableChanged() {
       if (!NetworkService.ethernetAvailable && root.panelViewMode === "ethernet") {
-        root.panelViewMode = "wifi";
+        root.panelViewMode = NetworkService.wifiAvailable ? "wifi" : root.vpnAvailable ? "vpn" : "wifi";
+      }
+    }
+    function onWifiAvailableChanged() {
+      if (!NetworkService.wifiAvailable && root.panelViewMode === "wifi") {
+        root.panelViewMode = NetworkService.ethernetAvailable ? "ethernet" : root.vpnAvailable ? "vpn" : "ethernet";
       }
     }
   }
@@ -171,10 +176,14 @@ SmartPanel {
   }
 
   onOpened: {
-    if (NetworkService.ethernetAvailable && !NetworkService.wifiEnabled) {
-      panelViewMode = "ethernet";
-    } else {
+    if (NetworkService.wifiAvailable && NetworkService.wifiEnabled) {
       panelViewMode = "wifi";
+    } else if (NetworkService.ethernetAvailable) {
+      panelViewMode = "ethernet";
+    } else if (vpnAvailable) {
+      panelViewMode = "vpn";
+    } else {
+      panelViewMode = "wifi"; // fallback (shows empty/disabled state)
     }
     panelViewPersistEnabled = true;
   }
@@ -307,6 +316,7 @@ SmartPanel {
               text: I18n.tr("common.wifi")
               tabIndex: 0
               checked: modeTabBar.currentIndex === 0
+              visible: NetworkService.wifiAvailable
             }
 
             NTabButton {
--- a/Modules/Panels/Network/NetworkPanel.qml
+++ b/Modules/Panels/Network/NetworkPanel.qml
@@ -25,7 +25,6 @@ SmartPanel {
 
   // Unified panel view mode: "wifi" | "ethernet" (persisted)
   property string panelViewMode: "wifi"
-  property bool panelViewPersistEnabled: false
 
   // If VPN configs vanish while on VPN tab, switch away
   Connections {
@@ -133,14 +132,10 @@ SmartPanel {
     } else if (panelViewMode === "vpn") {
       _tabIdx = 2;
     }
-    if (modeTabBar.currentIndex !== _tabIdx) {
+    if (modeTabBar && modeTabBar.currentIndex !== _tabIdx) {
       modeTabBar.currentIndex = _tabIdx;
     }
 
-    // Persist last view (only after restored the initial value)
-    if (panelViewPersistEnabled) {
-      Settings.data.network.networkPanelView = panelViewMode;
-    }
     if (panelViewMode === "wifi") {
       ethernetInfoExpanded = false;
       if (NetworkService.wifiEnabled && !NetworkService.scanningActive) {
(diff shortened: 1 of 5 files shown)


Modules/Panels/Network/NetworkPanel.qml
```diff
@@ -942,9 +950,258 @@ SmartPanel {
                 }
               }
             }
+
+            NBox {
```
notiant, line 1059: The layout of the box needs some rework:

Image

should have the same margins and centering as ethernet:

Image

The left and right margins of the VPN connections in you screenshot are also different from wifi/ethernet

Modules/Panels/Network/NetworkPanel.qml
```diff
@@ -265,9 +318,7 @@ SmartPanel {
                 anchors.margins: Style.marginM
                 spacing: Style.marginL
 
-                Item {
-                  Layout.fillHeight: true
-                }
+                Item { Layout.fillHeight: true }
```
notiant: This format change is out of style with the rest of the code. Also not sure why you are deleting so many comments

Modules/Panels/Network/NetworkPanel.qml
```diff
@@ -164,17 +178,17 @@ SmartPanel {
             }
           }
 
-          // Mode switch (Wi‑Fi / Ethernet)
           NTabBar {
             id: modeTabBar
-            visible: NetworkService.ethernetAvailable && NetworkService.wifiAvailable
```
notiant, line 170: I'm not a fan of having the tab bar always visible. I only use the widget for wifi and have no ethernet port, so this seems like unnecessary clutter.

Modules/Panels/Network/NetworkPanel.qml
```diff
@@ -58,27 +59,23 @@ SmartPanel {
...
-      if (last === "ethernet" && NetworkService.ethernetAvailable) {
-        panelViewMode = "ethernet";
-      } else {
-        panelViewMode = "wifi";
-      }
+    // Default opening logic:
+    // - If Wi-Fi is actively connected, open Wi-Fi first
+    // - Otherwise always open Ethernet first
+    // VPN is manual-only and should not be auto-selected
+    if (NetworkService.wifiConnected) {
```
notiant: This doesn't make sense on devices that have no ethernet port. There is a problem in the current opening logic that the ethernet tab is not shown if ethernet is connected but the last view was Wi-Fi and Wi-Fi is disabled, but this can be solved with this:

```
  onOpened: {
    // Restore last view if valid, otherwise choose what's available (prefer Wi‑Fi when both exist)
    const last = Settings.data.network.networkPanelView;
    if (NetworkService.ethernetAvailable && (last === "ethernet" || !NetworkService.wifiEnabled)) {
      panelViewMode = "ethernet";
    } else {
      panelViewMode = "wifi";
    }
    panelViewPersistEnabled = true;
  }
```

Modules/Panels/Network/NetworkPanel.qml
dinhokusanagi: I agree with that, I don't have Wi-Fi and it's visible in the widget. In this case, anything not in use should be hidden.

turannul: > I agree with that, I don't have Wi-Fi and it's visible in the widget. In this case, anything not in use should be hidden. 

It was the default before i even involved in it, i kept it as is. 
not my skill issue this time 😃 

Though this tab bar can be better

maybe (wifi || eth) && vpn

ultherego:
How it currently works:

1. The Wi-Fi tab is visible when `NetworkService.wifiAvailable is true` — that is, when a Wi-Fi adapter/interface exists in the system (e.g., wlan0). This property comes from `NetworkService.qml, which sets _wifiAvailable = true` when nmcli device finds a device of type wifi.

The tab is always primary.

2. The Ethernet tab is visible when `NetworkService.ethernetAvailable is true` — that is, when a physical Ethernet interface exists in the system (e.g., eth0, enp3s0). Similar to Wi-Fi, it is detected via nmcli device as a device of type ethernet.

The tab is visible, but launching the widget automatically sets Wi-Fi as primary.
If Wi-Fi is OFF (via the switch), it sets Ethernet as primary.


3. The VPN tab is visible when `root.vpnAvailable is true` — that is, when `Object.keys(VPNService.connections).length > 0`, in other words, when at least one VPN connection is configured in NetworkManager (e.g., WireGuard, OpenVPN).

This is a different condition than Wi-Fi/Ethernet — those depend on the presence of a physical adapter, while VPN depends on the presence of a configuration (connection profile).

Modules/Panels/Network/NetworkPanel.qml
```diff
@@ -27,7 +27,116 @@ SmartPanel {
...
+  // Whether VPN configurations exist (tab only shown when true)
+  readonly property bool vpnAvailable: Object.keys(VPNService.connections).length > 0
+
+  // If VPN configs vanish while on VPN tab, switch away
+  onVpnAvailableChanged: {
+    if (!vpnAvailable && panelViewMode === "vpn") {
+      panelViewMode = NetworkService.wifiAvailable ? "wifi" : "ethernet";
+    }
+  }
+
```
turannul: This would supposed to be in VPNService?

Retrieved/used within connections block or if (VPNService.x)

Modules/Panels/Network/NetworkPanel.qml
```diff
@@ -58,26 +171,23 @@ SmartPanel {
...
+      panelViewMode = "ethernet";
+    } else if (vpnAvailable) {
+      panelViewMode = "vpn";
     } else {
-      if (!NetworkService.wifiEnabled && NetworkService.ethernetAvailable) {
-        panelViewMode = "ethernet";
-      } else {
-        panelViewMode = "wifi";
-      }
+      panelViewMode = "wifi"; // fallback (shows empty/disabled state)
```
turannul: On a second thought wifi as fallback questionable choice. 
Do you think ethernet fits better?

Modules/Panels/Network/NetworkPanel.qml
```diff
@@ -942,9 +1088,269 @@ SmartPanel {
...
+                      }
+                    }
+                  }
+                }
+              }
+            }
+
+            // VPN empty state
+            NBox {
+              visible: panelViewMode === "vpn" && Object.keys(VPNService.connections).length === 0
```
turannul: I cannot try atm and rest of the day.

But have a question, empty view of vpn has no purpose i can think of...

turannul (changes requested):

turannul:
> How it currently works:
> 
> 
> 
> 1. The Wi-Fi tab is visible when `NetworkService.wifiAvailable is true` — that is, when a Wi-Fi adapter/interface exists in the system (e.g., wlan0). This property comes from `NetworkService.qml, which sets _wifiAvailable = true` when nmcli device finds a device of type wifi.
> 
> 
> 
> The tab is always primary.
> 
> 
> 
> 2. The Ethernet tab is visible when `NetworkService.ethernetAvailable is true` — that is, when a physical Ethernet interface exists in the system (e.g., eth0, enp3s0). Similar to Wi-Fi, it is detected via nmcli device as a device of type ethernet.
> 
> 
> 
> The tab is visible, but launching the widget automatically sets Wi-Fi as primary.
> 
> If Wi-Fi is OFF (via the switch), it sets Ethernet as primary.
> 
> 
> 
> 
> 
> 3. The VPN tab is visible when `root.vpnAvailable is true` — that is, when `Object.keys(VPNService.connections).length > 0`, in other words, when at least one VPN connection is configured in NetworkManager (e.g., WireGuard, OpenVPN).
> 
> 
> 
> This is a different condition than Wi-Fi/Ethernet — those depend on the presence of a physical adapter, while VPN depends on the presence of a configuration (connection profile).

If you want to dig further i do not tried vpns on linux/network manager
but macOS threated like virtual interface only if thats the case here.

we may can use deviceStatusProcess would super-seed entire VPNService :P 
with none to minimum effort

Honestly l don't know why i didn't do/tried doing this before.

turannul (approved):
Looks good to me

ultherego:
> > How it currently works:
> > 
> > 1. The Wi-Fi tab is visible when `NetworkService.wifiAvailable is true` — that is, when a Wi-Fi adapter/interface exists in the system (e.g., wlan0). This property comes from `NetworkService.qml, which sets _wifiAvailable = true` when nmcli device finds a device of type wifi.
> > 
> > The tab is always primary.
> > 
> > 2. The Ethernet tab is visible when `NetworkService.ethernetAvailable is true` — that is, when a physical Ethernet interface exists in the system (e.g., eth0, enp3s0). Similar to Wi-Fi, it is detected via nmcli device as a device of type ethernet.
> > 
> > The tab is visible, but launching the widget automatically sets Wi-Fi as primary.
> > If Wi-Fi is OFF (via the switch), it sets Ethernet as primary.
> > 
> > 3. The VPN tab is visible when `root.vpnAvailable is true` — that is, when `Object.keys(VPNService.connections).length > 0`, in other words, when at least one VPN connection is configured in NetworkManager (e.g., WireGuard, OpenVPN).
> > 
> > This is a different condition than Wi-Fi/Ethernet — those depend on the presence of a physical adapter, while VPN depends on the presence of a configuration (connection profile).
> 
> If you want to dig further i do not tried vpns on linux/network manager but macOS threated like virtual interface only if thats the case here.
> 
> we may can use deviceStatusProcess would super-seed entire VPNService :P with none to minimum effort
> 
> Honestly l don't know why i didn't do/tried doing this before.



Tested it — NM does create virtual interfaces (tun0 for OpenVPN, wg0 for WireGuard), and deviceStatusProcess sees them. However it can't replace VPNService because the interface only exists while VPN is active. Inactive VPN configs are invisible to `nmcli device show`:

```
# VPN inactive — device show has no trace of VPN:
❯ nmcli -t -f GENERAL.DEVICE,GENERAL.TYPE device show
GENERAL.DEVICE:enp5s0
GENERAL.TYPE:ethernet

# VPN active — tun0 appears but name is "tun0", not the VPN name:
❯ nmcli -t -f GENERAL.DEVICE,GENERAL.TYPE,GENERAL.CONNECTION device show tun0
GENERAL.DEVICE:tun0
GENERAL.TYPE:tun
GENERAL.CONNECTION:tun0

# Only connection show knows VPN configs and names:
❯ nmcli -t -f NAME,UUID,TYPE connection show | grep vpn
My VPN 1:xxxxxxxx-...:vpn:
My VPN 2:yyyyyyyy-...:vpn:
```

So deviceStatusProcess could complement VPNService (detect active VPN state, IP, DNS for bar icon etc.) but can't supersede it — we still need `nmcli connection show` for the list of available VPNs and connect/disconnect.

Modules/Panels/Network/NetworkPanel.qml
```diff
@@ -58,26 +171,23 @@ SmartPanel {
...
-      if (last === "ethernet" && NetworkService.ethernetAvailable) {
-        panelViewMode = "ethernet";
-      } else {
-        panelViewMode = "wifi";
-      }
+    if (NetworkService.wifiAvailable && NetworkService.wifiEnabled) {
+      panelViewMode = "wifi";
+    } else if (NetworkService.ethernetAvailable) {
+      panelViewMode = "ethernet";
+    } else if (VPNService.hasConnections) {
```
notiant, line 109: There needs to be an additional `NetworkService.wifiAvailable` check after the VPN check; if I disable wifi it opens the ethernet panel for me and i can't use the panel to enable wifi again!
Also, if you are removing the use of `Settings.data.network.networkPanelView` you should delete all references to it and `panelViewPersistEnabled` in the code. Settings have to be removed from settings.qml and settings-default.json as well.

ultherego: Good points, thanks for checking. I’ll push the fixes over the weekend and try to analyze it.

Additionally, I encountered an issue myself:

The widget was constantly enabled
I unplugged the Wi-Fi USB
I received a notification about it being unavailable
I plugged the Wi-Fi USB back in — the Wi-Fi unavailable view remained; only reopening the widget updated it

Services/Networking/NetworkService.qml
```diff
@@ -1151,11 +1158,26 @@ Singleton {
...
+          // Adapter added/removed: debounce to avoid rapid-fire refreshes
+          Logger.d("Network", "Device event: " + data);
+          _monitorDebounce.restart();
         }
       }
     }
   }
+
+  Timer {
+    id: _monitorDebounce
```
turannul: We do need this? onRead only triggers when nmcli prints

Services/Networking/NetworkService.qml
```diff
@@ -1151,11 +1158,26 @@ Singleton {
     stdout: SplitParser {
       onRead: data => {
         if (data.endsWith(": connected") || data.endsWith(": disconnected")) {
+          // High-priority state changes: refresh immediately
           Logger.d("Network", "State changed: " + data);
+          _monitorDebounce.stop();
           deviceStatusProcess.running = true;
           connectivityCheckProcess.running = true;
+        } else if (data.endsWith(": unavailable") || data.endsWith(": unmanaged") || data.indexOf(": device removed") !== -1 || data.indexOf(": device created") !== -1) {
```
turannul: We should not bother with unmanaged devices.

ultherego: I was in the process of checking what else was left to remove, etc.

turannul: looking for something to remove? - [See this](https://github.com/noctalia-dev/noctalia-shell/compare/main...turannul:noctalia-shell:pr/networking-qs-migration)

Was replacing Processes with quickshell API/Methods. - Whats left atm connectProcess

turannul (approved):

dinhokusanagi (approved):
Tested this PR locally using Nix.

Environment:
- Wayland (Hyprland)

Results:
- Build: OK
- Runs: OK
- No crashes observed
- Feature behaves as expected

Notes:
- Tested basic usage and interactions
- No regressions observed

Screenshot attached.

2026-04-18-190227_hyprshot

notiant:
There are some layout problems with the current version, in part because you copied from the ethernet layout which also has problems (the reference layout at the moment is the Wi-Fi view). Since there are a lot of changes, I've edited the current NetworkPanel.qml in your PR and pasted it here: https://pastebin.com/bmQdvaJP

~~Another issue is the refreshTimer in VPNService. I don't think it's necessary to run this every 5s as a background process and would remove the timer completely.~~
Just seeing that this wasn't added by you and is probably necessary for the VPN widget. Not sure if it's the best implementation though.

ultherego:
Thank you for taking a look at this.

~~Polling the VPN every 5 seconds doesn’t make sense — that’s what the refresh button is for.~~
I’ll take another look at the 5s interval, because from what I remember and from reviewing the code, the refreshTimer is only triggered in the following situations:
* Component.onCompleted — once at startup
* NetworkPanel.onPanelViewModeChanged — when the user switches to the VPN tab
* NetworkPanel.onEffectivelyVisibleChanged — when the panel is opened
* scheduleRefresh() — after connect/disconnect/refresh (1–2s delay)
* Refresh button in the panel


Another thing: I see you changed the logic of the connect/disconnect button in the VPN. I’m not sure if that’s a good idea, because they disappear for all items while they’re being used. Sometimes someone connects to several at once and just clicks connect/disconnect on selected ones simultaneously (it’s more about convenience than necessity).

The rest of the changes seem reasonable to me, thanks a lot 👍

notiant:
> Thank you for taking a look at this.
> 
> Polling the VPN every 5 seconds doesn’t make sense — that’s what the refresh button is for. Another thing: I see you changed the logic of the connect/disconnect button in the VPN. I’m not sure if that’s a good idea, because they disappear for all items while they’re being used. Sometimes someone connects to several at once and just clicks connect/disconnect on selected ones simultaneously (it’s more about convenience than necessity).
> 
> The rest of the changes seem reasonable to me, thanks a lot 👍

The connect/disconnect button only disappear for the VPN connection that's currently connecting/disconnecting. I've just copied the logic from the Wi-Fi tab. You can keep the `enabled` property if you want to prevent that people can connect to another VPN while there is a connect process ongoing.


For the VPN refresh: There exists a VPN widget for the bar and I think the refreshTimer might be necessary to keep it updated.

ultherego:
`visible: !VPNService.disconnecting && VPNService.disconnectingUuid !== modelData.uuid`
`!disconnecting` is `false` globally — so both buttons disappear regardless of the UUID.

I changed it to `enabled: !VPNService.disconnecting` it does the same thing (blocks both), but at least the buttons remain visible and the user can see they’re greyed out instead of disappearing without a trace.

notiant:
You're right, the visibility condition was wrong. I think the correct one should be `modelData.active && VPNService.disconnectingUuid !== modelData.uuid` for the disconnect button and `!modelData.active && VPNService.connectingUuid !== modelData.uuid` for the connect button? Still would change this to mirror the behavior in the Wi-Fi tab (and keep the enabled condition as it is).

ultherego:
You're right, updated. Used the same pattern as WifiSubTab — visible per-item (uuid check) + enabled as a global guard.
```
// disconnect
visible: VPNService.disconnectingUuid !== modelData.uuid
enabled: !VPNService.disconnecting

// connect
visible: VPNService.connectingUuid !== modelData.uuid
enabled: !VPNService.connecting
```

notiant (approved):

Issue #2431, opened 2026-04-10
[Feature] Battery toast should display which device has low battery


mikolajmikolajczyk:
## Summary

When a battery warning toast is shown, it does not indicate **which device** triggered the warning. This makes the notification ambiguous and unhelpful when multiple battery-powered devices are connected (e.g. laptop + wireless peripherals).

## Current behavior

A generic low battery toast appears with no indication of which device is affected.

## Expected behavior

The toast should include the device name (e.g. _"Logitech MX Vertical: 10% battery remaining"_), so the user knows which device to charge.

## Details

upower exposes the device model name alongside the battery level and warning level. For example:

```
model:          Logitech MX Vertical
battery-level:  low
warning-level:  low
percentage:     10%
```

This information is already available at the time the toast is triggered — it just needs to be surfaced in the notification message.

## Context

- Noctalia settings involved: `enableBatteryToast`, `batteryWarningThreshold`, `batteryCriticalThreshold`
- upower reports `warning-level: low` on peripheral devices (mice, keyboards, etc.), not just the laptop battery
- Without a device name, the user may assume it's the laptop battery and ignore the warning

mshiyaf:
Why was that PR #2446 closed ?
I had this issue too, I  was frequently getting this warning. I had no clue whether it was my laptop's battery got damaged or something else.
Only when i checked using upower, I found out the culprit, it was my MX Master mouse.

Better to have a proper toast mentioning the device name or model ?

turannul:
> Why was that PR #2446 closed ?
> I had this issue too, I  was frequently getting this warning. I had no clue whether it was my laptop's battery got damaged or something else.
> Only when i checked using upower, I found out the culprit, it was my MX Master mouse.
> 
> Better to have a proper toast mentioning the device name or model ?

Why closed? - maintainers decided to not to bother "clicking a single button to accept..." in favor of v5. 

I have no time to waste here anymore.

turannul:
[Replace files - profit!](https://drive.google.com/file/d/1jIgvFoinzutuwnGtZ5wzsjR9hF6MNCdt)
@mshiyaf

Ly-sec:
Thanks so much for reaching out and sharing this with us!

Right now, our team is entirely focused on building v5. Since the v4 lifecycle has officially come to an end, we are wrapping up old issues and won’t be shipping any further updates or fixes to that codebase.

[...]

PR #2360, opened 2026-03-30
Implement new view explicitly present upower requirement

turannul:
# Pull Request

## Motivation
Clearly show whenever upower is not installed.
https://discord.com/channels/1401598189823590460/1488171651550937248

## Type of Change
Mark the relevant option with an "x".
- [ ] Bug fix
- [X] New feature
- [ ] Breaking change
- [ ] Refactoring

## Related Issue
N/A

## Testing
Describe how you tested your changes and mark the relevant items.

- [X] Tested on niri
- [ ] Tested on Hyprland
- [ ] Tested on sway
- [ ] Tested with different bar positions and density settings
- [ ] Tested at different interface scaling values
- [ ] Tested with multiple monitors (if applicable)

## Screenshots / Videos
Screenshot from 2026-03-30 23-48-32

## Checklist
- [X] Code follows project style guidelines
- [X] Self-reviewed my code
- [X] No new warnings or errors
- [ ] Documentation or comments updated (if relevant)

## Additional Notes
Add any additional context or follow-up notes for reviewers.

PR #2446, opened 2026-04-12
Rework BatteryService - include [all/upower] devices

turannul:
# Pull Request

## Motivation
Fixes an issue

## Type of Change
- [X] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Refactoring

## Related Issue
#2431 

## Testing
Previously it doesn't showed my airpods name in the toast now it does.

- [X] Tested on niri
- [ ] Tested on Hyprland
- [ ] Tested on sway

## Screenshots / Videos
 (This image taken during testing) - Ignore %0 it was hardcoded variable
 ++ Style changed since then not using - in between warning and dev name
Screenshot from 2026-04-12 02-44-31

## Checklist
- [X] Self-reviewed my code

## Additional Notes
I don't mind more testers with niche devices, or previously effected by this issue.


notiant:
This won't fix the naming problem with Logitech mice that are connected with a unifying receiver. These are not Bluetooth devices. I'm also not sure why we kept this distinction between UPower and Bluetooth devices in the refactor. I think it might make more sense to treat every device as a UPower device and filter Bluetooth devices with `(device.nativePath && (device.nativePath.includes("bluez") || device.nativePath.includes("bluetooth")))`, so that we don't have to deal with devices that have different properties. The panel also needs to be changed so that a Logitech mouse can be displayed there (a UPower device that is not a laptop battery).

turannul:
> This won't fix the naming problem with Logitech mice that are connected with a unifying receiver. These are not Bluetooth devices. I'm also not sure why we kept this distinction between UPower and Bluetooth devices in the refactor. I think it might make more sense to treat every device as a UPower device and filter Bluetooth devices with `(device.nativePath && (device.nativePath.includes("bluez") || device.nativePath.includes("bluetooth")))`, so that we don't have to deal with devices that have different properties. The panel also needs to be changed so that a Logitech mouse can be displayed there (a UPower device that is not a laptop battery).

Bluetooth batteries does work (without UPower), just in case for some people who don't have UPower installed for some reason. Having dual-stack was Lemmy's suggestion.

notiant:
Yes, but I think it's cleaner to not have this distinction and instead rely on your PR #2360 to tell people they need to install upower.
Actually, I'm not even sure if we need to have Bluetooth devices as a separate model in the first place. It might make more sense to treat externally connected devices like Logitech mice (with a unifying receiver) exactly the same way as a Bluetooth device.

turannul:
b4890e0 - Prevents this from happening
Screenshot from 2026-04-13 01-42-20

Services/Hardware/BatteryService.qml
```diff
@@ -98,46 +87,48 @@ Singleton {
       return false;
     }
 
-    // Handle Bluetooth devices (identified by having batteryAvailable property)
-    if (device.batteryAvailable !== undefined) {
-      return device.connected === true;
+    if (!device.isLaptopBattery && device.type !== UPowerDeviceType.LinePower) {
```
notiant, line 90: Is this check really necessary? I don't think Bluetooth devices that aren't connected show up as UPower batteries.

Services/Hardware/BatteryService.qml
```diff
@@ -164,31 +154,24 @@ Singleton {
...
+    return !device.isLaptopBattery && device.type !== UPowerDeviceType.LinePower;
   }
 
   function getDeviceName(device) {
     if (!isDeviceReady(device)) {
       return "";
     }
 
-    if (!isBluetoothDevice(device) && device.isLaptopBattery) {
+    if (!isPeripheral(device) && device.isLaptopBattery) {
```
notiant: I think the check for `isPeripheral` is redundant here

Services/Hardware/BatteryService.qml
```diff
@@ -200,16 +183,7 @@ Singleton {
...
-    if (isBluetoothDevice(device) && device.name) {
-      return device.name;
-    }
-
-    if (device.model) {
-      return device.model;
-    }
-
-    return "";
+    return device.name || device.deviceName || device.model || "";
```
notiant, line 186: UPower devices don't have a `name` or a `deviceName` property

Services/Hardware/BatteryService.qml
```diff
@@ -336,14 +310,81 @@ Singleton {
...
+    if (!device) {
+      return "bt-device-generic";
+    }
+
+    const name = (device.model || device.deviceName || device.name || "").toLowerCase();
+    const nativePath = (device.nativePath || "").toLowerCase();
+    const iconHint = (device.icon || "").toLowerCase();
+
+    // 1. High-precision sub-types
+    if (name.includes("pod") || name.includes("bud") || name.includes("minor") || iconHint.includes("earbud"))
```
notiant: `name.includes("pod") || name.includes("bud")` look like they could easily lead to mismatches

Services/Hardware/BatteryService.qml
turannul: that is there because disconnected bluetooth devices are still preserved in `upower -d`

Services/Hardware/BatteryService.qml
turannul: hints are used secondary source - without name/hint it would return generic headphone icon.

notiant: what about a speaker with the name "SoundPod", wouldn't this device show the earbud icon although it's a speaker?
`device.name` and `device.deviceName`  should be removed here as well btw

turannul: i'll be improving this now.

turannul: I did something, somehow we still need deviceName it used as hint but no-one fool it this time

PR #2465, opened 2026-04-15
Make UPower only battery provider. - Replaces #2446

turannul:
I renamed branches that close Active Pull Requests this reopens it.

- [X]  Bug Fix
- [X] New Feature


turannul:
With 81ff96e I added changes in #2360 #2466

Services/Hardware/BatteryService.qml
```diff
@@ -6,49 +6,29 @@ import Quickshell
...
-import qs.Services.Networking
+import qs.Services.Networking // For Bluetooth device presence check
+import qs.Services.System // For UPower availability
 import qs.Services.UI
 
 Singleton {
   id: root
 
-  readonly property var primaryDevice: _laptopBattery || _bluetoothBattery || null // Primary battery device (prioritizes laptop over Bluetooth)
-  readonly property real batteryPercentage: getPercentage(primaryDevice)
```
notiant, line 16: Those properties are used by the Lock Screen, you shouldn't delete them

turannul: Thanks, showhow i didn't noticed this.
Search in vscode must be broken - searched before removing it...
We do not need to define it here to use it somewhere we can define/use anywhere within shell.
That was why built it like i don't remember

turannul:
The lock screen battery improved:

Laptop users will be seeing their peripheral batteries whenever below a threshold.

Desktop people will be seeing their primary device icon (A keyboard or mouse maybe headphones idk?)

Screenshot from 2026-04-24 08-36-48

Test case single battery no mockery/driver tricks but bluetooth devices.

Screenshot from 2026-04-24 08-35-08

eed0a3f - commit message way better than this anyway: simplified/fixed/improved my logic even more.

Screenshot from 2026-04-24 08-34-25

Services/Hardware/BatteryService.qml
```diff
@@ -6,49 +6,32 @@ import Quickshell
...
-    return list;
+  readonly property var laptopBatteries: {
+    if (!upowerInstalled) {
+      return [];
+    }
+    let laptopBatteriesList = (UPower.devices?.values ?? []).filter(d => d && d.isLaptopBattery).sort((x, y) => x.nativePath.localeCompare(y.nativePath, undefined, {
+                                                                                                                                             numeric: true
+                                                                                                                                           }));
+    if (UPower.displayDevice.isPresent && laptopBatteriesList.length > 0) {
+      return [UPower.displayDevice].concat(laptopBatteriesList.length > 1 ? laptopBatteriesList : []);
```
notiant: This was better before your recent change. At the moment, if the laptop only has 1 battery, then `laptopBatteries` will only contain the DisplayDevice and battery health won't work.

turannul: Fixed now thanks for heads up.
Screenshot from 2026-04-25 11-50-26
Screenshot from 2026-04-25 11-50-56

Services/Hardware/BatteryService.qml
```diff
@@ -6,49 +6,35 @@ import Quickshell
...
-    return list;
-  }
+    let physicalBatteries = (UPower.devices?.values ?? []).filter(d => d && d.isLaptopBattery && !isDisplayDevice(d));
+    physicalBatteries.sort((x, y) => x.nativePath.localeCompare(y.nativePath, undefined, {
+                                                                  numeric: true
+                                                                }));
 
-  readonly property var _laptopBattery: UPower.displayDevice.isPresent ? UPower.displayDevice : (laptopBatteries.length > 0 ? laptopBatteries[0] : null)
-  readonly property var _bluetoothBattery: bluetoothBatteries.length > 0 ? bluetoothBatteries[0] : null
+    if (UPower.displayDevice.isPresent && physicalBatteries.length > 0) {
```
notiant: This can be simplified:
```
return (UPower.displayDevice.isPresent && physicalBatteries.length > 1) ? [UPower.displayDevice].concat(physicalBatteries) : physicalBatteries;
```

Modules/LockScreen/LockScreen.qml
```diff
@@ -105,7 +105,13 @@ Loader {
             Item {
               Item {
                 id: batteryIndicator
-                readonly property var battery: BatteryService.primaryDevice
+                readonly property var battery: {
```
notiant, line 108: Not sure about this. I would either leave it just at `BatteryService.primaryDevice` or restrict it to just the critical battery level and look at all peripheral batteries instead of just the first.

turannul: Im not sure why you would want to restrict with Critical only. Doing so wouldn't gain meaningful value, but functionality loss

Services/Hardware/BatteryService.qml
turannul: Done.

Modules/LockScreen/LockScreen.qml
notiant: On a laptop it's supposed to show the laptop battery, so I would only show another device if it's really urgent (you really rely here on the fact that the icon that is shown is correct and people notice it; could be confusing)

turannul: Im confident connected devices are defined by UPower | where device.type should be accurate.

I don't think that ever would happen within consumer grade devices, to be clarify if it doesn't its on UPower.

I have verified it works as expected multiple times, AirPods and mouse.

turannul: I'd like to mention I am actively using this for over 2 (and half) weeks now.
Adding more whenever find some free time to improve it.

notiant: I would use this (if this PR ever gets merged :)):
```
readonly property var battery: {
  const pb = BatteryService.peripheralBatteries;
  if (pb.length > 0) {
    const critical = pb.find(b => BatteryService.isCriticalBattery(b));
    if (critical) {
      return critical;
    }
    const low = pb.find(b => BatteryService.isLowBattery(b));
    if (low) {
      return low;
    }
  }
  return BatteryService.primaryDevice;
}
```

turannul: honestly i don't think its going to be

turannul: I've used your way, but squashed a bit.

notiant (approved):

Modules/LockScreen/LockScreen.qml
turannul: Guess what i forgot :P 
```
                readonly property var battery: {
                  const lb = BatteryService.laptopBatteries[0]; // BAT0 or DD
                  if (BatteryService.isLowBattery(lb) || BatteryService.isCriticalBattery(lb)) {
                    return lb;
                  }
                  const pb = BatteryService.peripheralBatteries; // Peripherals
                  return (pb.length > 0 && (pb.find(b => BatteryService.isLowBattery(b)) || pb.find(b => BatteryService.isCriticalBattery(b))));
                }
```

PR #2475, opened 2026-04-16
Add option to not mark notifications as seen when opening the history

NicolaiRuckel:
## Motivation

I like to leave notifications on do not disturb for the most time. I still want to see when there's a new notification with the little dot on the bar icon but I don't want any pop-up notifications. When I check the new notifications by opening the history I may see some that are not actionable yet and that I want to deal with later. However, now the bar icon won't show that there are any new notifications (of course they are still in the history). This merge request adds an option to not mark notifications as seen anymore when opening the history.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring

## Testing

- [x] Tested on niri
- [ ] Tested on Hyprland
- [ ] Tested on sway

This is probably independent from all of the above.

## Checklist
- [x] Self-reviewed my code

## Additional Notes

There are two points that are worth discussing in my opinion:

* I could also see giving the option a name that describes that the dot on the bar icon shows that there are notifications in the history instead of unseen ones instead.
* I could also create a new plugin instead but that would be a lot of code duplication for a single added if clause and a feature toggle.

### DIFF
--- a/Assets/Translations/en.json
+++ b/Assets/Translations/en.json
@@ -1507,6 +1507,8 @@
       "settings-enabled-description": "Enable or disable the notification daemon, requires a restart of Noctalia shell.",
       "settings-enabled-label": "Enable notifications",
       "settings-location-description": "Where notifications appear on screen.",
+      "settings-mark-seen-on-open-label": "Mark notifications seen when opening history",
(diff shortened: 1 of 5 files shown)


ItsLemmy:

Cheers!

PR #2480, opened 2026-04-16
fix: stale data on wifi widget

### LINKED ISSUE #2184: Wi-Fi ui refresh promlem
### Description
After closing and reopening the laptop lid, the Wi-Fi icon/signal bars in the status bar do not reflect the currently active network. If the device automatically connects to a different access point while suspended (e.g., after moving to a new location), the indicator continues showing the previous network's state. The icon only updates after manually opening the quick settings menu and waiting ~3–4 seconds for it to re-query the network.

### Steps to Reproduce
1. Connect to a Wi-Fi network (e.g., `WiFi-A`).
2. Close the laptop lid (device suspends).
3. Move to a different location.
4. Open the laptop lid — NetworkManager automatically connects to a nearby saved network (e.g., `WiFi-B`).
5. Observe the status bar Wi-Fi icon — it still shows `WiFi-A` or the previous state.
6. Open the quick wifi settings menu — after ~3–4 seconds the icon updates to reflect `WiFi-B`.

### Expected Behavior
The Wi-Fi icon/signal indicator should automatically refresh upon lid open and immediately reflect the network that NetworkManager has connected to, without requiring any manual interaction.

### Environment
| Property | Value |
|---|---|
| Distro | CachyOS |
| Kernel | 6.19.7-1-cachyos |
| WM / Compositor | Hyprland |
| Noctalia Shell Version | v4.6.8-git |
| Noctalia QS Version | v0.0.8 |
| Host | VivoBook_ASUSLaptop M1603QA |
| CPU | AMD Ryzen 7 5800HS |
| GPU | Radeon Vega Series / Radeon Vega Mobile Series |
| RAM | 15.0 GB |

### Additional Context
- NetworkManager handles connection switching correctly in the background — this is purely a UI/indicator refresh issue.

danielepintore:
# Pull Request

## Motivation
closes #2184 
I have noticed that when I am connected to a network and suspend my laptop, the Wi-Fi widget displays stale SSID data from the previous connection upon resuming in a new location. A similar issue occurs during automatic failover: if the primary connection drops and the system automatically connects to a saved hotspot, the SSID name in the status bar remains unchanged despite the successful reconnection.

This pull requests forces a scan to update the network list and reflect the latest changes when those events occur. I've also implemented a mechanism to start again `nmcli -t monitor` if the process for some reason hangs or exits

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring

## Related Issue

## Testing

- [ ] Tested on niri
- [x] Tested on Hyprland
- [ ] Tested on sway

## Screenshots / Videos

## Checklist
- [x] Self-reviewed my code

### DIFF
--- a/Services/Networking/NetworkService.qml
+++ b/Services/Networking/NetworkService.qml
@@ -1171,8 +1171,10 @@ Singleton {
       onRead: data => {
         if (data.endsWith(": connected") || data.endsWith(": disconnected")) {
           Logger.d("Network", "State changed: " + data);
+          // Refresh status and trigger a scan to update SSID
           deviceStatusProcess.running = true;
           connectivityCheckProcess.running = true;
+          root.scan();
         }
       }
     }
--- a/Services/Networking/NetworkService.qml
+++ b/Services/Networking/NetworkService.qml
@@ -101,6 +101,17 @@ Singleton {
     }
   }
 
+  // Handle system resume to refresh network status and trigger a scan
+  Connections {
+    target: Time
+    function onResumed() {
+      Logger.i("Network", "System resumed - refreshing state");
+      deviceStatusProcess.running = true;
+      connectivityCheckProcess.running = true;
+      scan();
+    }
+  }
+
   // Start initial checks when nmcli becomes available
   Connections {
     target: ProgramCheckerService
@@ -196,6 +207,14 @@ Singleton {
     id: delayedScanTimer
     interval: 7000
     onTriggered: scan()
+  }
+
+  // Timer to restart the monitor process if it exits (e.g. after sleep or NM restart)
+  Timer {
+    id: monitorRestartTimer
+    interval: 2000
+    repeat: false
+    onTriggered: networkMonitorProcess.running = true
   }
 
   // Core functions
@@ -1156,6 +1175,10 @@ Singleton {
           connectivityCheckProcess.running = true;
         }
       }
+    }
+    onExited: {
+      Logger.w("Network", "Monitor process exited. Restarting in 2s...");
+      monitorRestartTimer.start();
     }
   }
 }


notiant:
The onResumed() function is unnecessary, the monitor process already triggers in this case. I think adding the scan there is enough to fix the problem with the stale SSID.

turannul:
The scan is taking so much time due to `--rescan yes` - without that it doesn't work as expected.
UI has to get new scan results to refresh.
That is not the solution.

danielepintore:
So if I've understood correctly I've to find a way to update the data without calling `scan()` or without the option `--rescan yes` right? I mean I've tested it and while the scan takes some time as you said I didn't see any problem on the bar

turannul:
> So if I've understood correctly I've to find a way to update the data without calling `scan()` or without the option `--rescan yes` right? I mean I've tested it and while the scan takes some time as you said I didn't see any problem on the bar

Correct.

In a perfect world, we shouldn't have store networks, instead read from dbus/iwd/wpa_supplicant/nm directly, (which one is most convenient) but that's wasn't possible at the time due to lack of API or dbus interface (within shell) didn't existed.

I wasn't allowed to write a noctalia-networkd and offload all the work and still not allowed to.

3-4 seconds actually acceptable tbh, i've saw worse (20 seconds sometimes or more)

edit: whenever you run scan in terminal, it has no effect to shell.

turannul:
> > So if I've understood correctly I've to find a way to update the data without calling `scan()` or without the option `--rescan yes` right? I mean I've tested it and while the scan takes some time as you said I didn't see any problem on the bar
> 
> Correct.
> 
> In a perfect world, we shouldn't have store networks, instead read from dbus/iwd/wpa_supplicant/nm directly, (which one is most convenient) but that's wasn't possible at the time due to lack of API or dbus interface (within shell) didn't existed.
> 
> I wasn't allowed to write a noctalia-networkd and offload all the work and still not allowed to.
> 
> 3-4 seconds actually good tbh, i saw 20s + 
> 

From what i remember without `--rescan yes` it only returned the network i've already connected to

adding secondary scan_fast() is not worth (to me) worth the complexity it brings.

I've tried making the interface in python it didn't bring 'expected' improvement either.

What you're seen it was best effort, without changing how things work.

notiant:
I think to address the problem with stale data the current PR is fine given that this will be reworked anyway in v5? I would only drop the `onResumed()` function.

turannul:
> I think to address the problem with stale data the current PR is fine given that this will be reworked anyway in v5? I would only drop the `onResumed()` function.

data is stale because fetching new one taking (I already mentioned... There is no good way to fix it, at this point not worth bothering) long, and yes within v5 this should not be an issue.

notiant:
> > I think to address the problem with stale data the current PR is fine given that this will be reworked anyway in v5? I would only drop the `onResumed()` function.
> 
> data is stale because fetching new one taking (I already mentioned... There is no good way to fix it, at this point not worth bothering) long, and yes within v5 this should not be an issue.

But the current problem is that the scan is only triggered when the network panel is opened and would otherwise indefinitely show the stale data, so this PR's band-aid fix to trigger a scan through `networkMonitorProcess` seems fine to me.

turannul:
> > > I think to address the problem with stale data the current PR is fine given that this will be reworked anyway in v5? I would only drop the `onResumed()` function.
> 
> > 
> 
> > data is stale because fetching new one taking (I already mentioned... There is no good way to fix it, at this point not worth bothering) long, and yes within v5 this should not be an issue.
> 
> 
> 
> But the current problem is that the scan is only triggered when the network panel is opened and would otherwise indefinitely show the stale data, so this PR's band-aid fix to trigger a scan through `networkMonitorProcess` seems fine to me.

Im not sure how though, nmcli -t monitor - should've triggered the update in bg without any interaction.

At least that's what it supposed to do

ItsLemmy:

Cheers!

Issue #2435, opened 2026-04-11
custom launch prefix ending with a space causes apps to fail to start


hachimagi:
### Description
1. enable custom launch prefix
2. set custom launch prefix to ``app2unit -- `` (space at the end)
3. apps launched from launcher dont launch, and i get a notification saying 
<img width="440" height="100" alt="Image" src="https://github.com/user-attachments/assets/f82476e8-9851-4260-879d-7b4cd3075dc5" />

### Environment
- Distro:  CachyOS
- Compositor: Hyprland
- Noctalia Shell: v4.7.5
- Noctalia QS: v0.0.12 (dbc319db0)  (0.0.12-1 in cachyos repo)

### Additional Context
multiple spaces in a row work fine in the terminal and in hyprland keybinds so not sure what's going on
i just put it there because I wasn't sure if the launcher adds a space but then  i spent like 40min trying to figure out why my apps aren't launching with a prefix 😭

automatically removing trailing spaces from the set custom prefix would prolly be easiest

Ly-sec:
Thanks so much for reaching out and sharing this with us!

Right now, our team is entirely focused on building v5. Since the v4 lifecycle has officially come to an end, we are wrapping up old issues and won’t be shipping any further updates or fixes to that codebase.

We are closing this issue for now, but we truly appreciate you taking the time to report it. We aren’t quite ready to open up the v5 tracker for public bug reports or feature requests just yet, but we’ll definitely keep your feedback in mind as we develop the new version!

Cheers!

PR #2499, opened 2026-04-19
Safeguard custom commands from trailing whitespaces

notiant:
This PR removes trailing whitespaces from custom commands in the launcher settings since they can cause errors (see issue #2435)


Ly-sec:
Hey @notiant thank you for the PR ♥️ I'll be merging it.

PR #2501, opened 2026-04-20
fix(tray): change the order in which the identifier for `blacklist` and `pinned` items is taken

lombervid:
# Pull Request

## Motivation
Currently the identifier for the `blacklist` and `pinned` items in the tray widget is being taken from the properties `tooltipTitle`/`name`/`id`, in that order of precedence. Taking `tooltipTitle` as the first option is not ideal since that property seems to be dynamic and changes depending on certain conditions.

For example:
- Blueman - changes depending on the status of the bluetooth (`Bluetooth Disabled` / `Bluetooth Enabled`)
- KeepassXC - changes based on the database focused (`[DATABASE_NAME] - KeePassXC`)

Because of this:
- It is difficult to know the correct name to use in the `blacklist` rules
- It is possible that you set a rule for an app but when the app changes status the rule doesn't match anymore

So, the order was changed to `id`/`title`/`tooltipTitle`.
`name` property was replaced with `title` since it [doesn't seem to be a valid property](https://quickshell.org/docs/master/types/Quickshell.Services.SystemTray/SystemTrayItem/).
`wildCardMatch` function was added in `TrayMenu.qml` to keep consistency on how pinned items are verified (since in both `Tray.qml` and `TrayDrawerPanel.qml` a `wildCardMatch` function is used to check for matches with regex patterns).

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring

## Testing
- Add rules in the `blacklist` field and verify the apps are omitted
- Verify the `Pin/Unpin Application` is working correctly
- Manually add a rule (that does not exactly match the `id` value of the app, i.e, `keepassxc`/`*KeePass*` for `KeePassXC`) in the `pinned` items in the settings
    - Verify it is pinned
    - Verify the menu show the `Unpin Application` button instead of `Pin Application`
    - Verify the `Unpin Application` button removes the entry from the settings

---
- [x] Tested on niri
- [x] Tested on Hyprland
- [ ] Tested on sway

## Screenshots / Videos
https://github.com/user-attachments/assets/24488cec-9fde-40fe-a976-5d9314b5869e

## Checklist
- [x] Self-reviewed my code

## Additional Notes
The only app that I've seen so far that could have a problem trying to blacklist with this approach is Discord, whose ID is `chrome_status_icon_1` in my case. (Pinning it using the `Pin Application` button works fine) .

### DIFF
--- a/Modules/Bar/Extras/TrayMenu.qml
+++ b/Modules/Bar/Extras/TrayMenu.qml
@@ -22,6 +22,23 @@ PopupWindow {
   // Derive menu from trayItem (only used for non-submenus)
   readonly property QsMenuHandle menu: isSubMenu ? null : (trayItem ? trayItem.menu : null)
 
+  // Helper functions
+  function wildCardMatch(str, rule) {
+    if (!str || !rule)
+      return false;
+    if (str === rule)
+      return true;
+    const placeholder = '\uE000';
+    let processedRule = rule.replace(/\*/g, placeholder);
+    let escaped = processedRule.replace(/[.+?^${}()|[\]\\]/g, '\\$&');
+    let pattern = '^' + escaped.replace(new RegExp(placeholder, 'g'), '.*') + '$';
+    try {
+      return new RegExp(pattern, 'i').test(str);
+    } catch (e) {
+      return false;
+    }
+  }
+
   // Compute if current tray item is pinned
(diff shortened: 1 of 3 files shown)


ItsLemmy:

Cheers!

PR #2507, opened 2026-04-20
fix(code): make markdown preview readable in light mode

kohanyirobert:
## Motivation

Markdown preview in VSCode/VSCodium is hard to read when the Noctalia palette is in **light mode**: the theme template hardcodes `"type": "dark"`, so VSCode uses dark-theme fallbacks for UI elements the theme doesn't explicitly color (including blockquotes and inline code backgrounds). The result is dark-on-dark or low-contrast text inside blockquote / code-block regions even when the rest of the editor is on a light surface.

This PR makes the VSCode template honor the current Noctalia mode and explicitly colors the markdown-preview surfaces from the palette.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring

## Related Issue
- Closes # (none)

## Changes

`Assets/Templates/code.json`:

- `"type": "dark"` → `"type": "{{mode}}"` — the template renderer already supports `{{mode}}` (resolved to `"dark"` / `"light"` via `default_mode`, see `Scripts/python/src/theming/lib/renderer.py`), so VSCode now picks the correct light/dark defaults.
- Added four palette-aware keys so blockquote / inline-code / preformatted blocks match the rest of the theme instead of falling back to VSCode defaults:
  - `textBlockQuote.background` → `surface_container_low`
  - `textBlockQuote.border` → `outline`
  - `textCodeBlock.background` → `surface_container`
  - `textPreformat.foreground` → `on_surface`

## Testing
- [x] Tested on niri
- [ ] Tested on Hyprland
- [ ] Tested on sway

Verified locally by applying the equivalent patch via a user template post-hook against both a wallpaper-derived light palette and the default Noctalia light scheme, then reloading the VSCode window and opening a markdown preview.

## Screenshots / Videos

Annotated `Markdown Kitchen Sink` preview — green highlights inline blockquote, red highlights a long blockquote. All screenshots are **light mode**; dark mode is unchanged.

**Before — wallpaper-derived light palette:**
before-wallpaper-light

**Before — default Noctalia light scheme:**
before-default-light

**After — wallpaper-derived light palette:**
after-wallpaper-light

**After — default Noctalia light scheme:**
after-default-light

## Checklist
- [x] Self-reviewed my code
- [ ] Documentation or comments updated (if relevant) — N/A

## Additional Notes
Dark mode is unchanged (since `{{mode}}` resolves to `"dark"` there). Only the four new color keys and the `type` field differ; no token colors or other UI surfaces are touched.

### DIFF
--- a/Assets/Templates/code.json
+++ b/Assets/Templates/code.json
@@ -1,6 +1,6 @@
 {
     "name": "NoctaliaTheme",
-    "type": "dark",
+    "type": "{{mode}}",
     "semanticHighlighting": true,
     "semanticTokenColors": {
       "enumMember": {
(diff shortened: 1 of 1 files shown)


ItsLemmy:

Cheers!

We now have a community-templates repo for v5, with the vscode one, you can PR that over there if you'd like

kohanyirobert:
Alrighty! Thanks for the update!

PR #2512, opened 2026-04-21
fix(ipc): added missing entry to Control settingsTabMap

Mondotosz:
# Pull Request

## Motivation

I was using the ipc calls to open different settings tabs and I couldn't open the idle tab.

```
qs -c noctalia-shell ipc call settings openTab idle
```

It looked like a simple fix so I went down the openTab function and found that `idle` isn't in the map
used to resolve the arguments so I simply added it like the other tabs.

I checked if it worked on my system and I could open the tab without issue.
I didn't check anything other than opening the tab since the change is so small and shouldn't have any impact.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring

## Testing

- [ ] Tested on niri
- [x] Tested on Hyprland
- [ ] Tested on sway

## Checklist
- [x] Self-reviewed my code

### DIFF
--- a/Services/Control/IPCService.qml
+++ b/Services/Control/IPCService.qml
@@ -104,7 +104,8 @@ Singleton {
                                             "system": SettingsPanel.Tab.System,
                                             "systemmonitor": SettingsPanel.Tab.System,
                                             "userinterface": SettingsPanel.Tab.UserInterface,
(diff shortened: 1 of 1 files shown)


Ly-sec:
Hey @Mondotosz , thank you for the PR :) we must've forgotten that one. Merging it.

PR #2513, opened 2026-04-21
NBusyIndicator: follow color change

notiant:
I noticed that the spinning circle in the network and Bluetooth panel (when connecting/disconnecting) doesn't use the proper color since it doesn't update its color when the connection status changes. Luckily, this can be fixed by just adding a single line.


Ly-sec:
Thank you very much @notiant :) merging it.

Issue #1957, opened 2026-02-24
Dynamic tray menu options do not update correctly


AlexAntonik:
### Description
When an application has a tray menu with dynamic options, they don’t work properly. For example, if I click Connect in the Amnezia VPN tray menu and the connection is established, the Disconnect option does not become active. The same happens with Tailscale-systray.
### Steps to Reproduce
All on video.

### Expected Behavior
Other option become active when state changes/only one option present(Connect or disconnect).

### Environment
- Distro:  NixOS
- Compositor:  Hyprland
- Noctalia-shell Version: 4.5.1-git
- Monitor resolution: 2880x1620

https://github.com/user-attachments/assets/e18c5755-09d8-488b-9475-a2b3ce010e0b

Ly-sec:
Hey @AlexAntonik , I just pushed a potential fix, can you please test it and let me know if that works for you?

AlexAntonik:
Still not updating

ItsLemmy:
I tried with nm-applet and it works fine. I can see the items dynamically refreshed in the tray menu.
Might be an issue with the app itself not broadcasting the changes. 
Did you try other apps ?

AlexAntonik:
Same issue with tailscale-systray and amnezia-vpn apps both from nixpkgs. On waybar this apps worked fine.

CubeTheThird:
I experience this with a couple of tools. One is [twitch-indicator](https://github.com/fldc/twitch-indicator), where the menu lists twitch streamers who are live (the list doesn't get updated).

The other is [Liferea](https://github.com/lwindolf/liferea) where the tray icon is supposed to update to reflect the number of updated RSS feeds.

Ly-sec:
Thanks so much for reaching out and sharing this with us!

Right now, our team is entirely focused on building v5. Since the v4 lifecycle has officially come to an end, we are wrapping up old issues and won’t be shipping any further updates or fixes to that codebase.

We are closing this issue for now, but we truly appreciate you taking the time to report it. We aren’t quite ready to open up the v5 tracker for public bug reports or feature requests just yet, but we’ll definitely keep your feedback in mind as we develop the new version!

Cheers!

PR #2515, opened 2026-04-21
Fix tray menu not closing / not updating

FUFSoB:
# Pull Request

## Motivation

This PR aims to fix two bugs with tray menus:
- They are not closing properly after making a click on item in the menu. The focus remains on the closed layer instead of going to windows, forcing to make an extra click to return back to normal.
- They are not updating changed text properly. For example, after enabling some option in the tray menu and then reopening that menu, it says it's still disabled, even though in the program itself it's actually enabled as expected.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring

## Related Issue
- Should probably solve #1957, though I did not test this with Amnezia.

## Testing

Did test with my own project with a tray icon, these changes make it work as expected.

Also tested with Telegram Desktop's "Enable/Disable Notifications", now works as expected.

- [x] Tested on niri
- [ ] Tested on Hyprland
- [ ] Tested on sway

## Screenshots / Videos
None

## Checklist
- [x] Self-reviewed my code

## Additional Notes
None

### DIFF
--- a/Modules/Bar/Extras/TrayMenu.qml
+++ b/Modules/Bar/Extras/TrayMenu.qml
@@ -462,15 +462,16 @@ PopupWindow {
                              } else {
                                // Click on regular items triggers them
                                modelData.triggered();
-                               root.hideMenu();
-
-                               // Close the drawer if it's open
-                               if (root.screen) {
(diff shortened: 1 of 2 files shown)


ItsLemmy:

Cheers!

PR #2520, opened 2026-04-21
fix(bluetooth): unblock rfkill when enabling a blocked adapter

Mic92:
## Motivation
quickshell's `BluetoothAdapter.setEnabled()` early-returns with `Cannot enable adapter because it is blocked by rfkill` and never writes the `Powered` property when the adapter is in the `Blocked` state. That leaves the bluetooth toggle dead after a partial airplane-mode exit — e.g. the laptop's airplane Fn key (kernel `rfkill block all`), then wifi is re-enabled via NetworkManager while bluetooth stays soft-blocked. `systemd-rfkill` persists that block across reboots, so the toggle stays broken until the user runs `rfkill unblock bluetooth` by hand.

`logind` grants the active seat user `rw` on `/dev/rfkill` (uaccess), so the shell can lift the soft block itself — same mechanism `setAirplaneMode()` already relies on. When the adapter is blocked and the user asks to enable it, run `rfkill unblock bluetooth && bluetoothctl power on` and let the resulting state change propagate back through quickshell.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring

## Related Issue
- n/a

## Testing
- [x] Tested on niri
- [ ] Tested on Hyprland
- [ ] Tested on sway

Steps:
1. `rfkill block bluetooth` → `bluetoothctl show` reports `PowerState: off-blocked`
2. Click the bluetooth toggle in the bar / control center
3. Before: log spams `Cannot enable adapter because it is blocked by rfkill`, toggle does nothing
4. After: log shows `Bluetooth Adapter rfkill-blocked, unblocking`, adapter ends up `Powered: yes`

## Screenshots / Videos
n/a (no visual change)

## Checklist
- [x] Self-reviewed my code

## Additional Notes
`bluetoothctl power on` is chained after the unblock because quickshell's `setEnabled` would still see the stale `Blocked` state if called synchronously; going through `bluetoothctl` avoids the race without adding a timer.

### DIFF
--- a/Services/Networking/BluetoothService.qml
+++ b/Services/Networking/BluetoothService.qml
@@ -152,6 +152,16 @@ Singleton {
       Logger.d("Bluetooth", "Enable/Disable skipped: no adapter");
       return;
     }
+    // quickshell's BluetoothAdapter.setEnabled() refuses to write Powered
+    // while the adapter is rfkill-blocked, so a partial airplane-mode exit
+    // (wifi unblocked, bluetooth still blocked) leaves the toggle dead.
+    // Lift the soft block ourselves; logind grants the seat user rw on
+    // /dev/rfkill so no privilege escalation is needed.
+    if (state && adapter.state === BluetoothAdapter.Blocked) {
+      Logger.i("Bluetooth", "Adapter rfkill-blocked, unblocking");
+      Quickshell.execDetached(["sh", "-c", "rfkill unblock bluetooth && bluetoothctl power on"]);
+      return;
+    }
     try {
       adapter.enabled = state;
       Logger.i("Bluetooth", "SetBluetoothEnabled", state);


notiant:
I've changed the Bluetooth toggle to be disabled when Bluetooth is blocked a couple of days ago, so it would be great if you can revert this change in your PR (in BluetoothSubTab).
What do you mean by "quickshell's setEnabled would still see the stale Blocked state if called synchronously"? I can't see any difference in behavior when testing with and without `bluetoothctl power on` (unblocking alone powers on Bluetooth for me).

Mic92:
The actual bug is in quickshell actually... Let me move the patch to this layer. Here is the quickshell upstream project: https://github.com/quickshell-mirror/quickshell/pull/765

will apply the same to the noctalia fork.

Mic92:
That's the better fix: https://github.com/noctalia-dev/noctalia-qs/pull/37

Issue #2541, opened 2026-04-24
BluetoothService: adapter.enabled stuck at false when shell starts before bluetooth daemon


Okydragon:
## Description

On a fresh session startup (no suspend involved), the Bluetooth widget in noctalia-shell permanently shows as **disabled**, even though the adapter is fully powered and operational at the system level.

This is a related but **distinct scenario** from #1750  — that issue covers the suspend/resume case. This one occurs on the very first startup when the shell initializes before the Bluetooth daemon is ready.

---

## Steps to Reproduce

1. Have noctalia-shell set to autostart on login.
2. Ensure the Bluetooth daemon (`bluetoothd`) is managed by systemd but starts slightly after the shell (common with USB-based adapters or manual service enablement).
3. Log in to the session.
4. Open the Bluetooth widget/panel in noctalia-shell.

**Expected:** Widget reflects the real adapter state (enabled).
**Actual:** Widget shows Bluetooth as disabled and remains stuck, even after the daemon is fully active.

---

## Evidence

While the widget shows disabled, the system state is fully healthy:

```bash
# Adapter is powered on at the DBus level
$ busctl get-property org.bluez /org/bluez/hci0 org.bluez.Adapter1 Powered
b true

# Daemon is active
$ systemctl status bluetooth
● bluetooth.service - Bluetooth service
     Active: active (running)

# No radio blocks
$ rfkill list
# All: Software: unblocked, Hardware: unblocked
```

Journal logs confirm the race condition — the shell started **19 minutes before** Bluetooth was enabled:

```
20:11:32 → noctalia-shell (qs) started
20:30:39 → bluetoothd enabled via systemctl
```

---

## Root Cause

In `Services/Networking/BluetoothService.qml`:

```qml
readonly property BluetoothAdapter adapter: Bluetooth.defaultAdapter
readonly property bool enabled: adapter?.enabled ?? false
```

When the shell starts before BlueZ is active, `Bluetooth.defaultAdapter` returns `null`. The `enabled` property resolves to `false` and **never updates** because the existing `ctlPollTimer` workaround only fires on `onResumed` (after suspend), never on initial adapter availability.

---

## Proposed Fix

Adding an `onAdapterChanged` handler that triggers `ctlPollTimer.restart()` whenever the adapter object becomes available for the first time:

```qml
onAdapterChanged: {
    if (adapter) {
        ctlPollTimer.restart();
    }
}
```

[...]

Issue #2533, opened 2026-04-23
Bar Icon Position Problem


Lighthrough0:
### Feature Description
The calendar widget ico position is a little bit lower，also when choosing Arch logo in Control center and launcher

### Why Is This Needed?
As a beautifier-software, more beautiful is necessary

### Suggested Solutions
to move the icon to the center 

### Additional Context
When choosing Arch logo，the resolution is not high enough

<img width="94" height="92" alt="Image" src="https://github.com/user-attachments/assets/072a88a7-82d6-426f-b99f-764b6f1035d5" />
<img width="484" height="102" alt="Image" src="https://github.com/user-attachments/assets/c5767a47-52f2-44fb-9195-e1cc05824aab" />

Lighthrough0:
The calendar widget icon position will sometimes get displayed properly

nanobreaker:
I have same issue on nixos, disto logo is slightly off towards right-bottom corner.

<img width="140" height="67" alt="Image" src="https://github.com/user-attachments/assets/0c7366d7-457c-4a5f-9b03-4fedf4f0415c" />

Ly-sec:
Thanks so much for reaching out and sharing this with us!

Right now, our team is entirely focused on building v5. Since the v4 lifecycle has officially come to an end, we are wrapping up old issues and won’t be shipping any further updates or fixes to that codebase.

We are closing this issue for now, but we truly appreciate you taking the time to report it. We aren’t quite ready to open up the v5 tracker for public bug reports or feature requests just yet, but we’ll definitely keep your feedback in mind as we develop the new version!

Cheers!

Issue #2461, opened 2026-04-14
Arch icon not centered in Waybar


Kwmee:
Hey, I know this is a minor issue and I could fix it myself, but it gets misconfigured with every update. It's a quick fix and it would be much appreciated!

<img width="1917" height="44" alt="Image" src="https://github.com/user-attachments/assets/96ee9156-ae70-440a-b0ab-7672d05372d2" />

<img width="61" height="42" alt="Image" src="https://github.com/user-attachments/assets/b473af1e-46a9-4832-ae9d-f7df84630b2b" />

Ly-sec:
Thanks so much for reaching out and sharing this with us!

Right now, our team is entirely focused on building v5. Since the v4 lifecycle has officially come to an end, we are wrapping up old issues and won’t be shipping any further updates or fixes to that codebase.

We are closing this issue for now, but we truly appreciate you taking the time to report it. We aren’t quite ready to open up the v5 tracker for public bug reports or feature requests just yet, but we’ll definitely keep your feedback in mind as we develop the new version!

Cheers!

PR #2544, opened 2026-04-24
Fix potential Bluetooth problem on boot

notiant:
1. Adding a fix for a potential problem with BluetoothService's `adapter.enabled` described here: #2541
2. Slightly reducing the distro logo for the launcher and control center widget to make the Arch icon fit better (some people complained about this: #2533 & #2461)

Before:
Screenshot_2026-04-24 12-17-22
After:
Screenshot_2026-04-24 12-17-42

### DIFF
--- a/Services/Networking/BluetoothService.qml
+++ b/Services/Networking/BluetoothService.qml
@@ -71,10 +71,16 @@ Singleton {
     }
   }
 
-  // Handle potential case where Quickshell doesnt't properly update adapter after system wakeup
+  // Handle potential case where Quickshell doesnt't properly update adapter after boot or system wakeup
   Connections {
     target: Time
     function onResumed() {
+      ctlPollTimer.restart();
+    }
+  }
+
+  onAdapterChanged: {
+    if (adapter) {
       ctlPollTimer.restart();
     }
   }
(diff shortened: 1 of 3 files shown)


Services/Networking/BluetoothService.qml
```diff
@@ -71,14 +71,20 @@ Singleton {
...
+  // Handle potential case where Quickshell doesnt't properly update adapter after boot or system wakeup
   Connections {
     target: Time
     function onResumed() {
       ctlPollTimer.restart();
     }
   }
 
+  onAdapterChanged: {
+    if (adapter) {
```
turannul, line 83: What if we also add? `} else { ctlPollTimer.stop(); }`
adapter became null we stop requesting, would improve efficiency aspect?

Services/Networking/BluetoothService.qml
```diff
@@ -71,14 +71,20 @@ Singleton {
     }
   }
 
-  // Handle potential case where Quickshell doesnt't properly update adapter after system wakeup
+  // Handle potential case where Quickshell doesnt't properly update adapter after boot or system wakeup
```
turannul, line 74: potential? at this point we know this happens. 😄

notiant: Well, "potential" because it doesn't happen on all systems. I've never encountered this problem.

Services/Networking/BluetoothService.qml
notiant: I really don't think that's an issue, especially given the temporary nature of the Quickshell code. I don't know if any PRs will even get merged from now on.

ItsLemmy:

Cheers!

PR #2557, opened 2026-04-25
Implement media volume handling

Scrumplex:
# Pull Request

## Motivation

I used to change my media volume by scrolling over my MPD component when I was
using waybar.

This adds that behavior to MediaMini and hooks the media volume up to OSD.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring

## Related Issue

## Testing
I was running my code and used mpdris2 to change the volume of my music. It
works in both directions, so changes through mpc get shown in the OSD and vice
versa.

Note that Quickshell exposes `volume` and `volumeSupported`. I have not tested how this behaves with `volumeSupported = false`

See https://github.com/noctalia-dev/noctalia-qs/blob/75d180c28a9ab4470e980f3d6f706ad6c5213add/src/services/mpris/player.hpp#L147-L149

- [x] Tested on niri
- [ ] Tested on Hyprland
- [ ] Tested on sway

## Screenshots / Videos

## Checklist
- [x] Self-reviewed my code

## Additional Notes

~~I think this should have other icons in the OSD, but I wasn't sure how to add
new icons to the ttf. At the same time, Tabler does not have any other volume
icons, though we could use `music` and `music-off` kinda like how the input
volume OSD works.~~ I chose to use `music` and `music-off`

One oddity I noticed is that `onWheel` of `MouseArea` gives different values
than `onWheel` of `Item` (as used in the Volume widget). The magic number 120
probably needs to be tweaked due to this. Alternatively, we could switch all
other widgets to use `MouseArea` to make this more consistent.

### DIFF
--- a/Assets/Translations/en-GB.json
+++ b/Assets/Translations/en-GB.json
@@ -1574,6 +1574,8 @@
       "types-lockkey-label": "Lock keys",
       "types-media-description": "Show OSD when media playback state changes (play, pause, skip).",
       "types-media-label": "Media playback",
+      "types-media-volume-description": "Show OSD when media volume changes.",
+      "types-media-volume-label": "Media volume",
       "types-title": "OSD trigger events",
       "types-volume-description": "Show OSD when audio output volume changes.",
       "types-volume-label": "Output volume"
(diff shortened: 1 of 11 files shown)


Scrumplex:
I guess I could also add a volume bar to the popup of MediaMini

ItsLemmy:

Cheers!

PR #2560, opened 2026-04-25
Fix stuck modifiers when auto-paste is enabled

shokerplz:
# Pull Request

## Motivation
For some reason when auto-paste is enabled wtype can sometimes not release modifiers, which causes all sorts of problems.
Even tho in the man wtype says `the modifiers get  released  automatically once the program terminates.` that does not seem to be the case on all systems.

This patch fixes this specific problem by releasing modifiers explicitly, such fix should not be disruptive for current users as it does not change current behavior it just adds explicit release of modifiers that should work implicitly.

I also totally understand if such change will be rejected as the problem seems to be on the wtype side, but I am not sure if it will be fixed there and as I said before it does not change current behavior, neither it adds any complexity.
For reference https://github.com/atx/wtype/issues/66

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring

## Testing

- [ ] Tested on niri
- [x] Tested on Hyprland
- [ ] Tested on sway

## Screenshots / Videos

## Checklist
- [x] Self-reviewed my code

## Additional Notes

### DIFF
--- a/Services/Keyboard/ClipboardService.qml
+++ b/Services/Keyboard/ClipboardService.qml
@@ -471,7 +471,7 @@ Singleton {
     }
     const isImage = mime && mime.startsWith("image/");
     const typeArg = isImage ? ` --type ${mime}` : "";
-    const pasteKeys = isImage ? "wtype -M ctrl -k v" : "wtype -M ctrl -M shift v";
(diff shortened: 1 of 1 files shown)


ItsLemmy:

Cheers!

PR #2629, opened 2026-05-06
Fix/lockscreen autofocus on resume

john-okeefe:
# Pull Request

## Motivation

After waking from suspend (e.g. opening a laptop lid), the lock screen's password input field loses focus. The user must move the mouse to trigger the existing `MouseArea.onEntered` workaround before they can type their password. The `TextInput` only receives initial focus via `Component.onCompleted`, which fires once when the component is created — it does not re-fire after a suspend/resume cycle.

This PR hooks into the existing `Time.resumed` singleton signal (which detects system resume via clock jumps) to automatically restore focus to the password input when the system wakes, eliminating the need to move the mouse first.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring

## Related Issue

## Testing

- [ ] Tested on niri
- [x] Tested on Hyprland
- [ ] Tested on sway

## Screenshots / Videos

N/A — no visual changes. The behavior change is that the password input is immediately ready for typing after resume, without requiring mouse movement.

## Checklist

- [x] Self-reviewed my code

## Additional Notes

The fix follows an established pattern in the codebase — `NightLightService`, `BluetoothService`, `DarkModeService`, and `SystemStatService` all connect to `Time.resumed` to handle post-resume behavior. No new dependencies or imports were introduced; `qs.Commons` (where the `Time` singleton lives) was already imported.

Changes are isolated to a single file (`Modules/LockScreen/LockScreen.qml`):

- Added a `Connections` block in the full lock screen component that calls `passwordInput.forceActiveFocus()` on `Time.resumed` (with an `activeFocus` guard to avoid redundant calls).
- Added an equivalent `Connections` block in the black-screen fallback component for the `blackScreenPasswordInput`.

### DIFF
--- a/Helpers/TrayIconResolver.js
+++ b/Helpers/TrayIconResolver.js
@@ -1,0 +1,29 @@
+function resolveCandidates(icon) {
+    if (!icon)
+        return [""];
+    if (!icon.includes("?path="))
+        return [icon];
+
+    const chunks = icon.split("?path=");
+    const name = chunks[0];
+    const path = chunks[1];
+    const fileName = name.substring(name.lastIndexOf("/") + 1);
+
+    return [
+        `file://${path}/${fileName}`,
+        `file://${path}/${fileName}.svg`,
+        `file://${path}/${fileName}.png`,
+        `file://${path}/hicolor/scalable/status/${fileName}.svg`,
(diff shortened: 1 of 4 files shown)


ItsLemmy:

Cheers!

PR #2636, opened 2026-05-06
Feat/lockscreen grace period

john-okeefe:
# Pull Request

## Motivation

Adds a configurable grace period to the lock screen that allows unlocking without a password shortly after the screen locks. This matches hyprlock's `--grace` behavior — any keypress or mouse movement (>5px dead zone) during the grace window triggers an immediate unlock, bypassing PAM authentication entirely.

This is useful when the screen locks due to idle timeout and the user returns within a few seconds, avoiding the friction of typing a password for what was clearly not a security event.

Based on a study of hyprlock's implementation — uses wall-clock timestamp comparison so the grace period naturally expires during suspend/resume, with an explicit `Time.resumed` handler as a safety net.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring

## Related Issue

## Testing

- [] Tested on niri
- [x] Tested on Hyprland
- [ ] Tested on sway

**Test scenarios verified:**
- Grace period unlock via keypress within the grace window
- Grace period unlock via mouse movement (>5px) within the grace window
- Grace period correctly expires after the configured duration — password required
- Grace period does NOT survive suspend/resume — password required after wake
- Black-screen lock surfaces (disabled monitors) also respect grace period
- Settings UI: toggle enables/disables, text input adjusts duration, inline layout on the right
- Default of 5 seconds, configurable 1–3600 seconds
- Grace period set to 0 (disabled) behaves identically to pre-feature behavior

## Screenshots / Videos

## Checklist
- [x] Self-reviewed my code

### Key implementation details

**Grace state lives on the Loader level** (`LockScreen.qml`). The `lockedAt` timestamp and `graceAllowed` flag are set in the Loader's `onActiveChanged` handler rather than inside the sourceComponent. This ensures they persist across `WlSessionLockSurface` recreations (which can happen on monitor hotplug or resume) and are set exactly once per lock/unlock cycle.

**Grace check uses a function, not a property binding.** The `isInGracePeriod()` function in `LockContext.qml` calls `Date.now()` on every invocation. This was critical — an initial implementation used a `readonly property bool` binding containing `Date.now()`, but QML's binding engine caches property values and only re-evaluates when tracked dependencies change. Since `Date.now()` is a JavaScript function call (not a trackable QML property), the cached value was never invalidated, causing the grace period to remain active indefinitely.

**Suspend safety** uses a two-layer defense:
1. Wall-clock comparison (`Date.now() < lockedAt + grace * 1000`) naturally expires the grace during suspend since the system clock advances past the deadline.
2. `Time.resumed` signal explicitly sets `graceAllowed = false` and `lockedAt = 0` on wake, ensuring the grace is dead before any user input can be processed.

### Files changed
| File | Change |
|------|--------|
| `Commons/Settings.qml` | `lockScreenGracePeriod` setting (int, default 5) |
| `Modules/LockScreen/LockContext.qml` | `isInGracePeriod()` function, `tryGraceUnlock()` method |
| `Modules/LockScreen/LockScreen.qml` | Grace state on Loader, keyboard/mouse interception with 5px dead zone |
| `Modules/Panels/Settings/Tabs/LockScreen/BehaviorSubTab.qml` | Inline toggle + text input UI |
| `Assets/Translations/en.json` | 4 new i18n keys |


ItsLemmy:

Cheers!

PR #2647, opened 2026-05-09
fix(MangoService): keyboard layout indicator delay for MangoWC

SUDOER1337:
I'm unsure if i have setup MangoWC keyboard layout switching the right way  

The keyboard layout indicator in the bar is delayed or never updates when switching layouts via XKB combo (e.g. `grp:win_space_toggle`) on MangoWC. The compositor doesn't emit DWL IPC `kbLayoutChanged` signals for XKB-driven switches — only when `switch_keyboard_layout` dispatch is explicitly called. This makes `DwlIpcOutput.kbLayout` go stale, so the bar can't detect layout changes in real-time.

Also enables cycleKeyboardLayout() using the switch_keyboard_layout dispatch command built into MangoWC.

The changes here come from help of OpenCode Big Pickle

- [x] Bug fix
- [ ] Breaking change
- [ ] New feature
- [ ] Refactoring

## Related Issue
-  N/A

- [x] Tested on MangoWC

## Checklist
- [x] Self-reviewed my code

### DIFF
--- a/Services/Compositor/MangoService.qml
+++ b/Services/Compositor/MangoService.qml
@@ -345,14 +345,47 @@ Item {
       }
 
       function onKbLayoutChanged() {
-        if (KeyboardLayoutService) {
-          KeyboardLayoutService.setCurrentLayout(modelData.kbLayout);
-        }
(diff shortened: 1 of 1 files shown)


ItsLemmy:

Cheers!

PR #2650, opened 2026-05-09
Add optional dock running dots

DFSko:
## Summary
- Add an optional `show_dots` dock setting, disabled by default.
- Render up to three palette-aware running window dots beside dock icons.
- Keep instance count badges unchanged for `2+` windows.

## Screenshot
Dock running dots

## Test Plan
- `git diff --check upstream/v5..HEAD`
- Manually verified the dock with `show_dots = true` on a right-positioned dock.

### DIFF
--- a/assets/translations/en.json
+++ b/assets/translations/en.json
@@ -1289,7 +1289,11 @@
         },
         "show-running": {
           "label": "Show Running",
-          "description": "Show indicators for running applications"
+          "description": "Show running applications that are not pinned"
+        },
+        "show-dots": {
+          "label": "Show Dots",
+          "description": "Show running window dots beside dock icons"
         },
         "show-instance-count": {
           "label": "Show Instance Count",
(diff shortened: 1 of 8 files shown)


Ly-sec:
Looks good overall, but I’d ask for one small fix before merging.

The new `show_dots` option is wired up cleanly, but it currently depends on `show_running` in a slightly wrong way. If someone disables `show_running` so the dock only shows pinned apps, the dots can disappear too, even though those pinned apps are still running. That feels backwards: `show_running` should mean “also show unpinned running apps”, not "track running state at all".

I’d suggest changing the dot visibility to use the actual window count that’s already fetched in `updateVisuals()`, instead of `item.running`. Something like:

```cpp
const std::size_t dotCount = std::min<std::size_t>(count, 3);
```

So: nice PR, good scope, just needs that behavior fixed before merge.

DFSko:
Fixed in c015d5211. Dots now use the actual window count, so pinned running apps still show dots when `show_running` is off, and apps with no windows do not show a dot.

Ly-sec:
Thank you very much @DFSko ! I'll merge it :)

PR #2652, opened 2026-05-10
Add dock launcher button options

DFSko:
## Summary
- Add optional dock launcher button placement via dock.launcher_position: none, start, or end.
- Add dock.launcher_icon with dots and search variants.
- Expose both options in settings and example config.

## Test Plan
- jq empty assets/translations/en.json
- python3 -c 'import tomllib; tomllib.load(open("example.toml", "rb")); print("ok")'
- git diff --check
- meson test -C build-debug (No tests defined.)

### DIFF
--- a/assets/translations/en.json
+++ b/assets/translations/en.json
@@ -841,10 +841,6 @@
         "start": "Start",
         "end": "End"
       },
-      "dock-launcher-icon": {
-        "dots": "Dots",
-        "search": "Search"
-      },
       "layer": {
         "top": "Top",
         "overlay": "Overlay"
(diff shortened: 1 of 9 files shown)


Ly-sec:
Hey @DFSko , I just tested the PR and it does work fine but I do think it might be better if we'd use `glyph_picker` instead of just having two options (while still providing a sane default). Offers more customizability and doesn't add lots of code.

DFSko:
Updated: launcher_icon now uses the existing glyph picker with grid-dots as the default.

Ly-sec:
Hey, @DFSko do you mind resolving the conflicts? Then I'll merge it.

DFSko:
Rebased on the latest v5 and resolved the dock config conflict. CI is green now.

Ly-sec:
> Rebased on the latest v5 and resolved the dock config conflict. CI is green now.

Thank you very much, I'll merge it now!

PR #2654, opened 2026-05-10
Add launcher category tabs

DFSko:
## Summary
- Add icon-only category tabs to the launcher.

## Screenshot
Launcher category tabs

## Test Plan
- `meson compile -C build-debug noctalia`
- Manually checked category switching, hover tooltip behavior, active tab state, and launcher reopen reset behavior.


ItsLemmy:
- Categories must be optional per provider. I personally wont use them on the app provider and like the minimal look
- Launcher_panel must not include provider specific code:  no appProvider()  same for launcher_panel_categories.cpp
- Button's glyphOnlySquare seems  unnecessary
- We now have a proper tooltip manager so no need  to implement custom ones in the controls.

DFSko:
Pushed an update for this.

I changed the category work so the launcher panel no longer knows about AppProvider directly. Category support now lives on the launcher provider interface, and providers opt into it only when they actually have categories.

I also added launcher.show_categories with the default set to false, so the launcher keeps the minimal look unless someone explicitly enables the tabs.

The custom hover label code is gone too. Category buttons now go through the existing InputArea/TooltipManager tooltip path, and I removed the glyph-only square mode from Button.

While rebasing this onto current v5, I also fixed the stale category state cases: if category tabs are disabled or hidden, the selected category is reset so the app provider cannot keep filtering behind the scenes.

ItsLemmy:
- docs/superpowers should not be versionned in the repo
- ```row->setRadius(Style::radiusMd * scale);```  should be ```row->setRadius(Style::scaledRadiusMd(scale));```
- tests/ui_control_test_stubs.cpp is orphaned, not wired into meson,  probably just remove it?

PR #2658, opened 2026-05-10
tray: harden dbusmenu loading and keep menu interactions non-blocking

Mathew-D:
tray: harden dbusmenu loading and keep menu interactions non-blocking


ItsLemmy:
tysm!

PR #2662, opened 2026-05-11
Fix stale template catalog entries

DFSko:
## Summary
- Remove stale Yazi entries from the built-in template catalog after the template moved to community templates.
- Make the Niri apply hook check for the exact generated root include instead of any noctalia.kdl include.

## Test Plan
- build-debug/noctalia theme --list-builtins
- temporary XDG_CONFIG_HOME checks for Niri root include insertion and idempotency
- git diff --check


ItsLemmy:
thanks!

Issue #2614, opened 2026-05-04
Incompatible with Hyprland 0.55's Lua-based dispatcher syntax


Gipphe:
### Description
In the upcoming version `0.55` of Hyprland, [the default syntax for configuration will be in Lua instead of Hyprconf][lua-ification]. This is already the default for the `git` version (which tracks `master`).

What is not described in the news post announcing these changes is that the `hyprctl dispatch` syntax also changes.

The following dispatcher
```
hyprctl dispatch 'exec noctalia-shell'
```
is now invoked as follows:
```
hyprctl dispatch 'hl.dsp.exec_cmd("noctalia-shell")'
```

This impacts the following:
- Launcher is unable to start applications.
- Unable to turn off or on monitors with `dpms`.
- Unable to kill windows.
- Unable to focus windows and bring them to the front.
- Unable to focus workspaces.

And I'm probably missing some that are less obvious.

### Steps to Reproduce
1. Use Hyprland `v0.55` (once it's out) or commit `5ba33f846129f3e6c0505b19f94634b24505e828` or later.
2. Start `noctalia-shell`.
3. Try to launch an application with the launcher.
4. Click a workspace in the bar.
5. Wait for the idle service to turn off the screen(s).

### Expected Behavior
The application launches, the workspace is switched to and the screen(s) should turn off.

### Environment
- Distro:  NixOS
- Compositor: Hyprland
- Noctalia-shell Version: v4.7.7-git (f0469d2)
- Noctalia QS Version: 0.0.12
- Monitor resolution: 1920x1080

### Additional Context

Related issue pertaining to the Lua config's impact on the template: #2609

[lua-ification]: https://hypr.land/news/26_lua/

Gipphe:
I have a proof-of-concept commit of changes to make `noctalia-shell` work for the current Hyprland `git` version as of commit `21fa9b2ee27227c964fdeb8090d98255edb89bec` [here](https://github.com/Gipphe/noctalia-shell/commit/74fd90d253991c337337a7affdbe57645234d822).

DarkXero-dev:
Yeah Hyprland 0.55 with lua just got released and noctalia dispatcher n color schemes broke

Puk4400:
I also updated to Hyprland 0.55 last night and today got greeted by a bunch of errors, after working to switch to the new lua syntax and fixing everything - noctalia is now broken.

What doesn't work on top of what's already mentioned here:
- Session menu

[...]

PR #2667, opened 2026-05-12
compositor: Updated Hyprland functions to support new Lua syntax

SirQuestion:
## Description
Since updating to Hyprland v0.55, all of the hyprctl syntax in Services/Compositor/HyprlandService.qml broke silently. None of the error logging worked aside from the few that I was able to catch, such as having to change workspaces with the bar, and it logging the syntax errors. However, when I first discovered this issue when I tried to launch my applications, it just failed silently without logging any errors. 

I've also taken into account that there's going to be people that will keep their old hyprlang config until last minute, so I've created a helper function to determine what environment Noctalia is running in. Since Lua is optional as of v0.55, detectDispatchMode() probes hyprctl to determine usage of either hyprlang or Lua. If "hl.dsp.no_op()" returns ok, HyprlandService will switch to the new Lua Syntax. If not, fall back to old hyprlang syntax. This is to support both those that haven't converted yet to Lua, and those that have.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Refactoring

## Related Issue
- Issue #2614 

## Testing

- [ ] Tested on niri
- [x] Tested on Hyprland
- [ ] Tested on sway

## Checklist
- [x] Self-reviewed my code
- [] Documentation or comments updated (if relevant)

## Additional Notes
No Further notes

### DIFF
--- a/Services/Compositor/HyprlandService.qml
+++ b/Services/Compositor/HyprlandService.qml
@@ -579,7 +579,7 @@ Item {
 
             const monitorsEvents = ["configreloaded", "monitoradded", "monitorremoved", "monitoraddedv2", "monitorremovedv2"];
 
-            if (monitorsEvents.indexOf(event.name) !== -1) {
+            if (monitorsEvents.includes(event.name)) {
                 Qt.callLater(queryDisplayScales);
             }
 
--- a/Services/Compositor/HyprlandService.qml
+++ b/Services/Compositor/HyprlandService.qml
@@ -6,724 +6,724 @@ import qs.Commons
 import qs.Services.Keyboard
 
 Item {
-    id: root
+  id: root
 
-    // Properties that match the facade interface
-    property ListModel workspaces: ListModel {}
-    property var windows: []
-    property int focusedWindowIndex: -1
+  // Properties that match the facade interface
+  property ListModel workspaces: ListModel {}
+  property var windows: []
+  property int focusedWindowIndex: -1
 
-    // Signals that match the facade interface
-    signal workspaceChanged
-    signal activeWindowChanged
-    signal windowListChanged
-    signal displayScalesChanged
+  // Signals that match the facade interface
+  signal workspaceChanged
+  signal activeWindowChanged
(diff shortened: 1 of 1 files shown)


SirQuestion:
Managed to fix the whitespace issue. Diff is now clean

Thomas-Philippot:
Could also be a good idea to change generated colors for hyprland in `~/.config/hypr/noctalia`

`noctalia-colors.conf` should be `noctalia-colors.lua`

don't know how to handle retro-compatibility though

TheLastZombie:
> don't know how to handle retro-compatibility though

I mean, I don't see any issues with just generating both `.conf` and `.lua`.

SirQuestion:
@Thomas-Philippot I'll see what I can do later today. Hyprland will not load anything related to hyprlang if it detects a .lua file in the config folder, so if I wanted to make the logic simple, I'll just generate both .lua and .conf. A better solution though would be to determine which syntax to generate. 

For now though, general compatibility is my priority, as workspaces, launching apps, and logging out simply don't work without this patch.

Ly-sec:
> @Thomas-Philippot I'll see what I can do later today. Hyprland will not load anything related to hyprlang if it detects a .lua file in the config folder, so if I wanted to make the logic simple, I'll just generate both .lua and .conf. A better solution though would be to determine which syntax to generate.
> 
> For now though, general compatibility is my priority, as workspaces, launching apps, and logging out simply don't work without this patch.

Shall we wait with merging it then? We're not planning to merge more stuff to v4 after this one.

Thomas-Philippot:
I'll rather see this merged quickly as it completely breaks noctalia on v0.55 

Maybe theme could be done later on ?

Ly-sec:
> I'll rather see this merged quickly as it completely breaks noctalia on v0.55
> 
> Maybe theme could be done later on ?

Well the issue is mostly that we're planning one final release with the lua stuff fixed (even though we really didn't plan to do that either).

PR #2672, opened 2026-05-12
Refactor hyprland sockets and update commands for lua config

ElJeffe:
- deduplicate code, all socket interaction and json handling for hyprland is moved to hyprland_runtime.
- use the corect command for hyprland, depending on whether the config is lua or conf.

Tested with hyprland lua:
- changing workspaces
- keyboard layout changing
- dpms
- logout

# Pull Request

## Motivation

## Type of Change
- [x ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ x] Refactoring

## Related Issue

## Testing

- [ ] Tested on niri
- [x ] Tested on Hyprland
- [ ] Tested on sway
- [ x] Tested with multiple monitors (if applicable)

## Screenshots / Videos

## Checklist
- [ x] Code follows project style guidelines
- [ x] Self-reviewed my code

## Additional Notes

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -590,6 +590,7 @@ _noctalia_sources = files(
   'src/compositors/ext_workspace/ext_workspace_output_backend.cpp',
   'src/compositors/ext_workspace/ext_workspace_backend.cpp',
   'src/compositors/hyprland/hyprland_runtime.cpp',
+  'src/compositors/hyprland/hyprland_event_handler.cpp',
   'src/compositors/hyprland/hyprland_output_backend.cpp',
   'src/compositors/hyprland/hyprland_workspace_backend.cpp',
   'src/compositors/hyprland/hyprland_keyboard_backend.cpp',
--- a/src/compositors/compositor_platform.cpp
+++ b/src/compositors/compositor_platform.cpp
@@ -172,10 +172,6 @@ namespace {
     NiriWorkspaceBackend m_backend;
   };
 
-  [[nodiscard]] bool setHyprlandOutputPower(WaylandConnection& /*wayland*/, bool on) {
-    return compositors::hyprland::setOutputPower(on);
-  }
-
   [[nodiscard]] bool setMangoOutputPower(WaylandConnection& wayland, bool on) {
     return compositors::mango::setOutputPower(wayland, on);
   }
@@ -188,7 +184,10 @@ namespace {
   createOutputPowerBackend(compositors::CompositorRuntimeRegistry& runtimeRegistry) {
     switch (compositors::detect()) {
     case compositors::CompositorKind::Hyprland:
-      return std::make_unique<LambdaOutputPowerBackend>(&setHyprlandOutputPower);
+      return std::make_unique<LambdaOutputPowerBackend>(
+          [&runtime = runtimeRegistry.hyprland()](WaylandConnection& /*wayland*/, bool on) {
+            return compositors::hyprland::setOutputPower(runtime, on);
+          });
     case compositors::CompositorKind::Niri:
       return std::make_unique<LambdaOutputPowerBackend>(
           [&runtime = runtimeRegistry.niri()](WaylandConnection& /*wayland*/, bool on) {
--- a/src/compositors/hyprland/hyprland_event_handler.cpp
+++ b/src/compositors/hyprland/hyprland_event_handler.cpp
@@ -1,0 +1,13 @@
+#include "compositors/hyprland/hyprland_event_handler.h"
+
+#include "compositors/hyprland/hyprland_runtime.h"
+
+namespace compositors::hyprland {
+
+  HyprlandEventHandler::HyprlandEventHandler(HyprlandRuntime& runtime) : m_runtime(runtime) {
+    m_runtime.registerEventHandler(this);
+  }
+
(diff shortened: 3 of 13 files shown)


LionHeartP:
Can't test atm, but thanks for this. I can finally drop my patch.

ItsLemmy:
### Critical Issue

  **Missing files: `hyprland_event_handler.h` and `hyprland_event_handler.cpp` are not included in the PR.** These files are:
  - Added to `meson.build` (line 589)
  - `#include`d by both `hyprland_keyboard_backend.h` and `hyprland_workspace_backend.h`
  - Used as a base class (`compositors::hyprland::HyprlandEventHandler`) by both backends
  - Forward-declared in `hyprland_runtime.h`
  - Included by `hyprland_runtime.cpp`

  This PR **will not compile** as submitted. The author likely forgot to `git add` these new files.

  ### Design Concerns

  1. **`HyprlandRuntime` is no longer stateless — but used as a temporary in `setOutputPower()` and `performLogout()`.**

     In `hyprland_output_backend.cpp:63` and `session_panel.cpp:51`, a local `HyprlandRuntime runtime;` is constructed on the stack. The new constructor
  calls `ensureResolved()` + `updateConfigProvider()` (which does a synchronous IPC round-trip to `j/status`). This means every DPMS toggle and every
  logout creates a fresh socket connection and queries the config provider — wasteful. Worse, these stack-local runtimes have their own `m_eventHandlers`
  vector and `m_eventSocketFd = -1`, completely disconnected from the main runtime instance the rest of the shell uses. If `connectSocket()` were called on
   them, it would create a parallel event socket with no handlers.

     **Recommendation:** These call sites should receive a reference to the existing `HyprlandRuntime` instead of constructing throwaway instances.

  2. **Synthetic events via string formatting feel fragile.**

     `handleEvent("cleanup>>");` and `handleEvent("changecallback>>");` in `HyprlandRuntime::cleanup()`/`dispatchPoll()` synthesize fake IPC event strings
  that get parsed through `line.find(">>")`. These are internal control signals masquerading as Hyprland wire protocol messages. If Hyprland ever adds real
   events with those names, there's a collision.

     **Recommendation:** Use a separate method or enum for lifecycle signals rather than routing them through the IPC event parser.

  3. **`available()` semantics changed silently.**

     Before: `available()` returned true if the request socket path was resolved (i.e., Hyprland is running).
     After: `available()` returns `m_eventSocketFd >= 0` — true only if the *event* socket is connected.

     This is a behavioral change. Code that calls `available()` to check "can I send a request?" will now get `false` before `connectSocket()` is called,
  even though `request()` still works independently via the request socket. The temporary `HyprlandRuntime` instances in `setOutputPower`/`performLogout`
  never call `connectSocket()`, so they'd report `!available()` while successfully sending requests.

  4. **`m_configIsLua` is `mutable` but written once in the constructor.**

     `updateConfigProvider()` is `const` and writes to `mutable m_configIsLua`. This is fine for lazy init patterns, but here it's called eagerly in the
  constructor. If the config changes at runtime (Hyprland supports `configreloaded`), this value goes stale. Consider refreshing it on `configreloaded`
  events.

  5. **Multiple inheritance diamond risk.**

     `HyprlandWorkspaceBackend` now inherits from `WorkspaceBackend`, `WorkspaceOutputNameResolver`, `WorkspaceSocketConnector`, AND
  `HyprlandEventHandler`. That's four bases. The `handleEvent` virtual override is clean, but the class is getting complex. Not blocking, but worth noting.

  ### Code Style

  - `configProvider.compare("lua") == 0` — use `configProvider == "lua"` per idiomatic C++.
  - `const std::string_view command` — the top-level `const` on `string_view` parameters is unconventional (it's already cheap to copy). Not wrong, just
  inconsistent with the rest of the codebase which uses `std::string_view` without `const`.
  - The `bool res` variable in `HyprlandWorkspaceBackend::connectSocket()` could be inlined: `if (m_runtime.connectSocket()) { refreshSnapshot(); return
  true; } return false;`.

  ### Positive Aspects

  - The deduplication goal is sound — socket management, JSON parsing, and event dispatch were copy-pasted across backends.
  - Lua config detection is a real need for Hyprland's evolving config system.
  - The event handler registration pattern (`registerEventHandler`/`unregisterEventHandler`) is a clean observer approach.

  ### Summary

  The refactoring direction is good, but this PR needs work before merge:

  1. **Blocker:** Add the missing `hyprland_event_handler.h/cpp` files
  2. **High:** Fix the temporary `HyprlandRuntime` construction in `setOutputPower()` and `performLogout()` — pass the existing runtime by reference
  3. **Medium:** Don't route internal lifecycle signals (`cleanup`, `changecallback`) through the IPC event parser
  4. **Medium:** Reconcile the `available()` semantics change
  5. **Low:** Style nits (`compare` vs `==`, unnecessary `const` on value params)

ElJeffe:
about the comments:
- The session_panel still uses a stateless HyprlandRuntime. More code needs to be refactored to use the existing runtime instance. I noticed that Niri does this too. I can look into this in another commit.
-  the isAvailable() call is only used to decide wether to call openSocket() or not. The OpenSocket opens the event socket, so I think ckecking if the event socket exists is better here.

ItsLemmy:
Thanks!  Looking forward to  more cleanup!

PR #2675, opened 2026-05-13
fix: prevent notification history crash on invalid UTF-8

DFSko:
## Problem

Long notification strings were truncated with a raw byte `substr(0, 1024)`. When the cut landed in the middle of a multi-byte UTF-8 code point, the stored notification body contained invalid UTF-8.

Notification history persistence later serialized that body with the default strict `nlohmann::json::dump(2)` path. That throws `json.exception.type_error.316` for incomplete UTF-8. In the observed crash, the uncaught exception triggered `std::terminate` and aborted the process.

## Summary

- truncate notification strings at UTF-8 code point boundaries
- serialize notification history with replacement handling for malformed persisted strings
- add a regression test for split UTF-8 notification bodies

## Tests

- `find src \( -name '*.cpp' -o -name '*.h' \) -print0 | xargs -0 clang-format --dry-run -Werror`
- `meson test -C build-pr notification_history_utf8 --print-errorlogs`
- `meson compile -C build-pr`
- `meson test -C build-pr --print-errorlogs`

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -715,3 +715,18 @@ executable('noctalia',
 install_subdir('assets',
   install_dir: get_option('datadir') / 'noctalia',
 )
+
+notification_history_utf8_test = executable('notification_history_utf8_test',
+  sources: files(
+    'tests/notification_history_utf8_test.cpp',
+    'src/core/log.cpp',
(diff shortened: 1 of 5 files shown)


ItsLemmy:
Thanks

PR #2678, opened 2026-05-13
feat(ncspot): add theme template for ncspot

aberrios85:


aberrios85:
this writes to `.config/ncspot/colors.toml` instead of `.config/ncspot/config.toml` to avoid clobbering a user's config already sitting there.

ItsLemmy:

Cheers!


We now have a community-templates repo if you wanna prepare it for v5

PR #2684, opened 2026-05-14
feat(nix): add Home Manager module

linusammon:


pengolord:
Why are widgets and palletes separate from the other configurations?

pengolord:
actually, why are we modifying state files at all? `~/.local/state/noctalia` is for runtime overrides, I don't think it's supposed to be set at buildtime

linusammon:
> actually, why are we modifying state files at all? `~/.local/state/noctalia` is for runtime overrides, I don't think it's supposed to be set at buildtime

Because that's where widgets are stored, there is no `XDG_CONFIG_HOME` equivalent.  
https://docs.noctalia.dev/v5/desktop/widgets/#desktop-widgets

As for palettes, noctalia expects them in `XDG_CONFIG_HOME/palettes`.  
https://docs.noctalia.dev/v5/theming/#custom_palette

pengolord:
alright, palletes make sense, but I still think we should not be doing anything in the `~/.local/state/noctalia` directory, users can use the `settings.desktop_widgets` option

linusammon:
`desktop_widgets` is only a toggle whether or not to enable desktop widgets, it does not contain the widgets themselves.

> Desktop widgets are enabled from config.toml. Widget instances and edit-mode grid settings are stored in a separate state file.
> 
> - Config toggle: [desktop_widgets] in ~/.config/noctalia/config.toml
> - State file: $NOCTALIA_STATE_HOME/noctalia/desktop_widgets.toml, $XDG_STATE_HOME/noctalia/desktop_widgets.toml, or ~/.local/state/noctalia/desktop_widgets.toml  

> Widget definitions are not read from config.toml — edit mode writes them to the state file so positions and transforms can be changed interactively.

PR #2693, opened 2026-05-15
fix(mpris): Stops hammering unresponsive players after a few attempts

Mathew-D:
Stops hammering unresponsive players after a few attempts
 Uses exponential backoff to reduce CPU usage while trying 
 Cleans up resources by canceling timers for unresponsive players
 Allows recovery if a player becomes responsive again 
 Maintains existing behavior for functional MPRIS players


ItsLemmy:
Thanks!

PR #2694, opened 2026-05-15
Theming: Added support for hyprland lua

SirQuestion:
## Description
In the current version of Hyprland v0.55, Lua is optional. I've applied the same theming fix as v4, where apply.sh probes hyprctl to determine whether or not hyprland is running in Lua mode or hyprlang mode. 

With this probing, Noctalia will be the one to determine which template color to generate, either the .conf or .lua. I've also made some slight modifications towards the template_engine, where I've added support for dynamic input paths. I've mirrored the current implementation with the dyanmic output path, since this was required if Noctalia were to determine which config file to generate.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Refactoring

## Related Issue
- N/A

## Testing

- [ ] Tested on niri
- [x] Tested on Hyprland
- [ ] Tested on sway

## Checklist
- [x] Self-reviewed my code


ItsLemmy:
Thanks!

ItsLemmy:
```  PR #2694 Review — Hyprland Lua Theming Support

  Overview

  Adds support for Hyprland's optional Lua config mode (v0.55+). A shell script probes whether Hyprland is running in Lua or hyprlang mode, then the
  template engine dynamically selects the correct input template and output path. Introduces a new input_path_dynamic field mirroring the existing
  output_path_dynamic pattern.
  template engine dynamically selects the correct input template and output path. Introduces a new input_path_dynamic field mirroring the existing

  What's Good

  - Clean architecture: The three-mode script (input/output/apply) is well-structured and keeps compositor-specific logic in the shell script, not in C++.
  - Mirrors existing pattern: input_path_dynamic follows the same render→runSync→parse flow as output_path_dynamic, keeping the engine consistent.
  - Graceful fallback: Detection falls back from hyprctl probe → file existence → default conf mode.
  - The Lua template is well-formed and covers the same color slots as the existing .conf template.

  Issues & Suggestions

  1. Missing hyprland.lua template file in the diff tree
  The script references ./hyprland/hyprland.lua but the current repo only has hyprland.conf. The new hyprland.lua file is in the diff — make sure the
  build/install step copies it (check meson.build or the assets install list).

  2. effectiveInput can be empty — no error handling
  In template_engine.cpp around the new block (~line 1290 in the PR), if inputPathDynamic is set but the command fails or returns no paths, effectiveInput
  stays at entry.inputPath which could be empty (since builtin.toml hyprland entry no longer has input_path). This would pass an empty string to
  renderFile().

  // If input_path_dynamic is set but input_path is absent, effectiveInput == ""
  // and renderFile("", outputPath) will fail silently or error.

  Suggestion: Add a guard after resolving effectiveInput:
  if (effectiveInput.empty()) {
    ok = false;
    continue; // or break
  }

  3. detect_mode is called three times in a single apply run
  The apply.sh script runs detect_mode separately for input, output, and apply invocations. Each call probes hyprctl dispatch which is an IPC round-trip.
  Not a correctness bug, but:
  - There's a theoretical race if the user switches config modes between calls (unlikely but possible).
  - Three IPC calls where one would suffice.

  Consider caching the result in an env var or accepting it as a second argument.

  4. detect_mode probe command could be fragile
  hyprctl dispatch 'hl.dsp.no_op()' 2>/dev/null | grep -qx 'ok'
  This relies on hl.dsp.no_op() existing and returning exactly ok. If Hyprland changes this behavior, detection silently falls back to file-based
  heuristic, which is fine — but worth a comment noting the specific Hyprland version this targets.

  5. dofile() uses absolute path — correct but worth noting
  dofile("/home/user/.config/hypr/noctalia.lua")
  This is fine for a single-user config but won't survive if $XDG_CONFIG_HOME differs between template-apply time and Hyprland load time. The existing
  .conf template has the same characteristic (source = /absolute/path), so this is consistent.

  6. Shell script set -euo pipefail with grep -q
  In apply_conf(), the grep -qE will cause the script to exit if the pattern is NOT found (exit code 1 from grep + set -e). This is guarded by the if !
  negation so it should be fine, but double-check — set -e and negated commands can be tricky across shells.

  7. The input_path_dynamic field isn't parsed for input_path_modes (dark/light)
  The existing code handles input_path_modes with dark/light variants. If someone wants dynamic + mode-aware input, that's not supported. Probably fine for
   now but worth documenting.

  Minor Nits

  - The new hyprland.lua template uses tabs for indentation while .conf uses spaces — not a functional issue but inconsistent.
  - The detect subcommand in the script is a nice debug aid.

  Verdict

  The approach is solid and follows existing patterns well. The main actionable concern is issue #2 (empty effectiveInput fallback) — that's a real bug
  path since the hyprland entry in builtin.toml removes input_path entirely, so if the dynamic command fails, there's no static fallback and renderFile
  gets an empty string. The rest are minor improvements.
  ```

SirQuestion:
I've added some error handling for effectiveInput, as well as caching to avoid multiple calls to detect_mode for probing. This should fix the concerns that you might have.

ItsLemmy:
Excellent, thank you!

PR #2697, opened 2026-05-16
fix: match app identity separator variants

DFSko:
## Problem
Mattermost Desktop changed its Wayland app id from the desktop-entry-style `mattermost-desktop` to `Mattermost.Desktop`. The dock could resolve the pinned desktop entry for launching and icons, but running-window detection and activation still compared app ids literally. As a result, Mattermost appeared unpinned/not running and clicking the dock item launched another instance instead of focusing the existing window.

## Summary
- Normalize app identity matching across desktop ids and StartupWMClass by ignoring separator differences (`.`, `-`, `_`, spaces) while keeping display-name matching exact.
- Reuse the shared identity matcher for dock pinned lookup and Wayland toplevel window lookup.
- Add regression coverage for desktop entry resolution, running app dedupe, hidden/noDisplay entries, and toplevel window matching.

## Test Plan
- `meson compile -C build`
- `meson test -C build --print-errorlogs`


Ly-sec:
Hey @DFSko , looks good overall, and the fix makes sense to me.

One thing I would tweak before merging: `resolveRunningApps()` now dedupes by the separator-stripped identity key. That means unrelated apps like `foo.bar` and `foobar` could collapse into one running app, so one of them might disappear from the dock/taskbar.

I think we should keep the separator-insensitive matching for lookup, but dedupe a bit more carefully:

- if a running app resolves to a real desktop entry, dedupe by that resolved `entry.id`
- if it stays as an unknown/fallback app, dedupe by the original lowercased app id, not the stripped key

That should keep the wanted behavior while avoiding accidental app collisions.

DFSko:
Adjusted the dedupe path: resolved desktop entries now dedupe by the resolved entry id, while fallback/unknown apps dedupe by the original lowercased running app id. Added a regression case for unknown separator variants staying distinct.

Verification:
- `meson compile -C build`
- `meson test -C build --print-errorlogs`
- `git diff --check`

Ly-sec:
@DFSko if you fix the conflict I'll merge it asap :) thank you very much.

DFSko:
@Ly-sec done

PR #2709, opened 2026-05-17
fix(launcher): invalidate cache when refreshing desktop entries

jasdeep-dhillon:
# Pull Request

## Motivation
Currently whenever opening the launcher there is a very big delay before it renders and is able to accept input. This is happening because the icon cache gets invalidated whenever it opens and has to rebuild before it is able to display the launcher panel. This PR aims to fix the slow launch of the launcher panel.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring

## Testing

- [x] Tested on niri
- [ ] Tested on Hyprland
- [ ] Tested on sway

## Screenshots / Videos

Before:

https://github.com/user-attachments/assets/c2e5f15c-38f3-49cc-86eb-b0cfc228906e

After:

https://github.com/user-attachments/assets/d4ef4afb-a045-4788-b804-f7b8dc34c04e

## Checklist
- [x] Self-reviewed my code

### DIFF
--- a/src/shell/launcher/launcher_panel.cpp
+++ b/src/shell/launcher/launcher_panel.cpp
@@ -380,8 +380,6 @@ void LauncherPanel::onOpen(std::string_view context) {
   if (m_grid != nullptr) {
     m_grid->scrollView().setScrollOffset(0.0f);
   }
-  // Clear cached icon misses before each open so newly installed app icons appear.
-  m_iconResolver.invalidateCache();
   onInputChanged(initialValue);
 }
 
--- a/src/system/icon_resolver.cpp
+++ b/src/system/icon_resolver.cpp
@@ -466,7 +466,11 @@ const std::string& IconResolver::resolve(const std::string& iconName, int target
   if (it != m_cache.end()) {
     return it->second;
   }
-  auto [ins, _] = m_cache.emplace(key, findIcon(iconName, targetSize));
+  auto icon = findIcon(iconName, targetSize);
+  if (icon.empty()) {
+    return m_empty;
+  }
+  auto [ins, _] = m_cache.emplace(key, icon);
   return ins->second;
 }
 


ItsLemmy:
Thanks for the PR!

Unfortunately can't merge as is,  the new invalidation does nothing:

``` 
IconResolver resolver;
resolver.invalidateCache();
```

IconResolver::m_cache is a per-instance unordered_map. The only shared state is the static theme plan (iconThemeState()), not the resolved-icon cache.

So this code constructs a throwaway local IconResolver, clears its own (already-empty) cache, and immediately destroys it. The launcher's actual m_iconResolver member is never touched.

So this PR indeed boost up performances on launcher open but also bring in a major regression as new icons wont appear anymore.

The codebase already has the right pattern to fix it properly via the generation counter,  you could add something like: 
```
void IconResolver::invalidateAll() {
    ensureThemeState();
    ++iconThemeState().generation;
}
```
And call it from DesktopEntryCache::refreshIfNeeded()

Also kLog.info("Refreshing desktop entries")  should be kLog.debug. 

Thanks again.

jasdeep-dhillon:
Does the icon resolver handle theme changes or not? Because even without using invalidating cache at all new desktop entries do end up showing in the launcher and even changing the icon seems to work. I tried it by making a new desktop entry and deleting/editing it to check if entries are getting updated. Icon theme changes do not seem to reflect but they do not seem to be working when using with the original implementation either. Let me know if I'm missing something. Sorry, for taking up for time with this.


https://github.com/user-attachments/assets/09bd2cfb-bafd-41ae-bce8-73b8bdadcf13

ItsLemmy:
Yeah, your observations are right. New entries show up via the desktop entry watcher (inotify + version counter), which is independent of the icon cache. And a new app has a new icon name that was never cached, so it resolves fresh anyway.
  
Theme changes are handled, just elsewhere: IconThemePollSource polls every 60s, calls checkThemeChanged(), bumps a shared generation counter, and resolvers rebuild on next resolve(). So the old per-open invalidateCache() didn't really help theme changes either, you're right about that

The only thing it actually protected: an icon name gets looked up, misses (cached empty), then the file gets installed later under the same name while the shell keeps running.   So yep it's a really tiny edge case.

Best fix is to reuse the generation counter instead of poking a per-instance cache.

jasdeep-dhillon:
> The only thing it actually protected: an icon name gets looked up, misses (cached empty), then the file gets installed later under the same name while the shell keeps running. So yep it's a really tiny edge case.
> 
I tried using the cache generation but since the file does not get changed it doesn't trigger desktop entries refresh. Instead I made a change to icon resolver so it doesn't cache the `iconName` if it cannot find the icon. Though it leaves the icon blank if icon is moved again but I do not think that will be solvable there's no way to find the same icon without scanning the whole filesystem.


https://github.com/user-attachments/assets/42fede73-d987-4e34-a5e7-fb6a628c9a04

ItsLemmy:
Thanks!

PR #2717, opened 2026-05-17
wpa_supplicant support for noctalia-v5

rXelelo:
# Pull Request

## Motivation

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring

## Related Issue

## Testing

- [x] Tested on niri
- [ ] Tested on Hyprland
- [ ] Tested on sway

## Screenshots / Videos

## Checklist
- [x] Self-reviewed my code

## Additional Notes


ItsLemmy:
Thanks for this.

1. Synchronous blocking D-Bus I/O in signal handlers (performance)  rebuildState() is invoked from every  propertiesChanged and ScanDone signal. For each call it iterates all interfaces and all BSSes, and for every BSS it does sdbus::createProxy() plus several synchronous getProperty() round-trips (SSID, Signal, RSN, WPA). With 20–40 visible APs that is 100+ blocking D-Bus  calls per signal, and PropertiesChanged fires frequently (e.g. on signal-strength updates). This runs on the poll-based main loop and will stall the UI. 
The NetworkManager backend is deliberately async. This one should be too.

2. dynamic_cast to gate the secret agent.
if (dynamic_cast<NetworkService*>(m_networkService.get()) != nullptr) {
Type-switching on the polymorphic interface defeats the point of the abstraction and it will silently do the wrong thing if a third backend is ever added.  Prefer a virtual capability query on INetworkService (e.g. virtual bool supportsSecretAgent() const).

3. Missing #include <algorithm>
wpa_supplicant_service.cpp uses std::sort and std::find_if but includes only <cstdint>/<map>/<string>/<vector> (+ sdbus/log). It compiles only via  transitive includes, fragile and a likely build break under a different libstdc++/header layout. Add the include explicitly.

rXelelo:
@ItsLemmy i fixed 3rd was fixed in progress

src/dbus/network/wpa_supplicant_service.cpp
```diff
@@ -0,0 +1,482 @@
...
+
+      return info;
+    } catch (const sdbus::Error&) {
+      return std::nullopt;
+    }
+  }
+
+} // namespace
+
+WpaSupplicantService::WpaSupplicantService(SystemBus& bus) : m_bus(bus) {
```
ItsLemmy, line 106: WpaSupplicantService::WpaSupplicantService(SystemBus& bus) : m_bus(bus) {
    if (!bus.nameHasOwner("fi.w1.wpa_supplicant1")) {
      throw sdbus::Error(sdbus::Error::Name{"org.freedesktop.DBus.Error.ServiceUnknown"},
                         "The name fi.w1.wpa_supplicant1 was not provided by any .service files");
    }

This should help people that have neither NM nor wpa_supplicant, so that the "network service disabled" path is properly reached.

src/dbus/network/wpa_supplicant_service.cpp
```diff
@@ -0,0 +1,482 @@
...
+  NetworkState next;
+  next.wirelessEnabled = m_wirelessEnabledOverride.value_or(!m_interfaces.empty());
+
+  std::vector<AccessPointInfo> aps;
+  std::string activeBssPath;
+
+  for (const auto& [ifacePath, proxy] : m_interfaces) {
+    const std::string state = getPropertyOr<std::string>(*proxy, k_wpaIfaceInterface, "State", "inactive");
+    const std::string ifname = getPropertyOr<std::string>(*proxy, k_wpaIfaceInterface, "Ifname", "");
+
```
ItsLemmy, line 394: we miss the next.scanning for parity with NM.
I think wpa_supplicant has the property, so something like 

next.scanning = next.scanning ||
                  getPropertyOr<bool>(*proxy, k_wpaIfaceInterface, "Scanning", false);

ItsLemmy:
Could be cool to also document the differences/limitation of wpa_supplicant vs NM in the header file, ex:
- No VPN support
- No IP address detection
etc...

The rest is really looking very good, ty!

rXelelo:
this is for two 

image

ItsLemmy:
Thanks again!

PR #2727, opened 2026-05-19
dock: add per-monitor output selector & hide when empty

CLEMENTINATOR:
# Pull Request

## Motivation
Dock: Settings should have an output/monitor selector

## Type of Change
- [ ] Bug fix
- [X] New feature
- [ ] Breaking change
- [ ] Refactoring

## Testing

- [X] Tested on niri
- [ ] Tested on Hyprland
- [ ] Tested on sway

## Screenshots / Videos
image

## Checklist
- [X] Self-reviewed my code

## Additional Notes

### DIFF
--- a/assets/translations/en.json
+++ b/assets/translations/en.json
@@ -1613,6 +1613,10 @@
           "label": "Active Monitor Only",
           "description": "Only show the dock on the active monitor"
         },
+        "monitors": {
+          "label": "Displays",
+          "description": "Choose which displays show the dock; leave empty to show on all displays"
+        },
         "icon-size": {
           "label": "Icon Size",
           "description": "Size of dock icons in pixels"
--- a/assets/translations/en.json
+++ b/assets/translations/en.json
@@ -1611,7 +1611,7 @@
       "dock": {
         "active-monitor-only": {
           "label": "Active Monitor Only",
-          "description": "Only show the dock on the active monitor"
+          "description": "Only show applications running on the active monitor"
         },
         "monitors": {
           "label": "Displays",
--- a/assets/translations/en.json
+++ b/assets/translations/en.json
@@ -1615,7 +1615,7 @@
         },
         "monitors": {
           "label": "Displays",
-          "description": "Choose which displays show the dock; leave empty to show on all displays"
+          "description": "Choose which displays show the dock; leave empty for all. Active-monitor filtering applies within these displays."
         },
         "icon-size": {
           "label": "Icon Size",
--- a/src/config/config_export.cpp
+++ b/src/config/config_export.cpp
@@ -507,6 +507,7 @@ namespace config_export {
       table.insert_or_assign("launcher_position", dock.launcherPosition);
       table.insert_or_assign("launcher_icon", dock.launcherIcon);
       table.insert_or_assign("pinned", stringArray(dock.pinned));
+      table.insert_or_assign("monitors", stringArray(dock.monitors));
       return table;
     }
 
--- a/src/config/config_overrides.cpp
+++ b/src/config/config_overrides.cpp
@@ -319,7 +319,8 @@ namespace {
            nearlyEqual(a.activeScale, b.activeScale) && nearlyEqual(a.inactiveScale, b.inactiveScale) &&
            nearlyEqual(a.activeOpacity, b.activeOpacity) && nearlyEqual(a.inactiveOpacity, b.inactiveOpacity) &&
            a.showDots == b.showDots && a.showInstanceCount == b.showInstanceCount &&
-           a.launcherPosition == b.launcherPosition && a.launcherIcon == b.launcherIcon && a.pinned == b.pinned;
+           a.launcherPosition == b.launcherPosition && a.launcherIcon == b.launcherIcon && a.pinned == b.pinned &&
+           a.monitors == b.monitors;
   }
 
   bool shellConfigEqual(const ShellConfig& a, const ShellConfig& b) {
--- a/src/config/config_service.cpp
+++ b/src/config/config_service.cpp
@@ -1837,6 +1837,8 @@ void ConfigService::parseTableInto(const toml::table& tbl, Config& config, bool 
       dock.launcherIcon = *v;
     if (auto* arr = (*dockTbl)["pinned"].as_array())
       dock.pinned = readStringArray(*arr);
+    if (const auto* v = (*dockTbl).get("monitors"))
+      dock.monitors = readStringArray(*v);
   }
 
   // Parse [desktop_widgets]
--- a/src/config/config_service.cpp
+++ b/src/config/config_service.cpp
@@ -1837,8 +1837,8 @@ void ConfigService::parseTableInto(const toml::table& tbl, Config& config, bool 
       dock.launcherIcon = *v;
     if (auto* arr = (*dockTbl)["pinned"].as_array())
       dock.pinned = readStringArray(*arr);
-    if (const auto* v = (*dockTbl).get("monitors"))
-      dock.monitors = readStringArray(*v);
+    if (auto* arr = (*dockTbl)["monitors"].as_array())
+      dock.monitors = readStringArray(*arr);
   }
 
   // Parse [desktop_widgets]
--- a/src/config/config_types.h
+++ b/src/config/config_types.h
@@ -347,6 +347,7 @@ struct DockConfig {
   std::string launcherPosition = "none";  // none | start | end
   std::string launcherIcon = "grid-dots"; // Tabler glyph name
   std::vector<std::string> pinned;        // desktop entry IDs to always show
+  std::vector<std::string> monitors;      // connector names to show on; empty = all outputs
 
   bool operator==(const DockConfig&) const = default;
 };
--- a/src/shell/dock/dock.cpp
+++ b/src/shell/dock/dock.cpp
@@ -154,6 +154,22 @@ namespace {
     return (cfg.launcherPosition == "start" || cfg.launcherPosition == "end") ? 1U : 0U;
   }
 
+  bool outputHasAnyDockItems(const DockConfig& cfg, const CompositorPlatform& platform, const WaylandOutput& output) {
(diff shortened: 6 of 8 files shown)


Ly-sec:
Hey @CLEMENTINATOR , I just checked the code and found two things that do need some attention.

1: Use `outputMatchesSelector(m, output)` instead of `m == output.connectorName`, same as notifications and bar overrides, so manual TOML values stay consistent.

2: When `erase_if` drops an instance (especially with `hide_when_empty`), clear it from `m_surfaceMap` and null `m_hoveredInstance` / `m_popupOwnerInstance` if they point at it, otherwise you can keep stale map entries or use-after-free on hover while the output still exists.

ItsLemmy:
Thanks a lot, looking pretty good.   I've not worked on the dock at all so here is an AI review for you. Take it with a large grain of salt as usual...

---

  ### Overview

  Adds a `monitors` config key to `DockConfig` (string array of connector names) that limits which displays show the dock. When empty, the dock appears on all outputs (current behavior).
  Also hides the dock on outputs that have no items to show. Includes a settings GUI entry using the existing `ListSetting` + `availableOutputs` pattern.

  ### What's Good

  - **Follows existing patterns well.** The `ListSetting` with `suggestedOptions = env.availableOutputs` is the exact same pattern used for notification monitors at
  `settings_registry.cpp:1263`. The `outputMatchesSelector` reuse is correct.
  - **`detachInstanceState` is a solid addition.** It properly cleans up dangling pointers (`m_hoveredInstance`, `m_popupOwnerInstance`, surface map entries) before instance destruction,
  and it's called from both `closeAllInstances` and the erase-if in `syncInstances`. This fixes a potential use-after-free that existed before this PR.
  - **Translation strings are clean** and the description for `active-monitor-only` is improved ("running on the active monitor" is clearer than "show the dock on the active monitor").

  ### Issues

  **1. Parsing inconsistency — `monitors` vs `pinned` (minor)**

  ```cpp
  // pinned (existing):
  if (auto* arr = (*dockTbl)["pinned"].as_array())
      dock.pinned = readStringArray(*arr);

  // monitors (new):
  if (const auto* v = (*dockTbl).get("monitors"))
      dock.monitors = readStringArray(*v);
```
  pinned guards on .as_array() then dereferences, while monitors calls .get() and passes the toml::node* to readStringArray. Both work — readStringArray internally calls node.as_array()
  and returns empty if it's not an array — but the monitors style silently accepts a non-array value and returns [], whereas pinned would skip the assignment entirely leaving the default.
   For a new field where the default is also [], this is functionally equivalent, but using the .as_array() guard would be more consistent with surrounding code.

  2. outputHasAnyDockItems calls runningAppIds() on every syncInstances tick (medium)

  refresh() now calls syncInstances(), which calls outputAllowed() for every output, which calls outputHasAnyDockItems(), which may call platform.runningAppIds(filter). If refresh() is
  called frequently (e.g., on every compositor running-app signal), this constructs and scans the running app list multiple times per frame just to decide whether to show the dock
  surface. The cost is likely small in practice (few outputs, few apps), but worth noting.

  3. refresh() behavior change — early return semantics shifted (worth verifying)

  Before:
```cpp
  void Dock::refresh() {
      if (m_instances.empty()) return;  // bail if no instances
      refreshPinnedAppsIfNeeded();
      // ... request updates
  }
```
  After:
```cpp
  void Dock::refresh() {
      if (m_config == nullptr || m_platform == nullptr || !m_config->config().dock.enabled) return;
      refreshPinnedAppsIfNeeded();
      syncInstances();                  // may create or destroy instances
      if (m_instances.empty()) return;
      // ... request updates
  }
 ```
  The old code bailed early if no instances existed (dock not shown). The new code always runs refreshPinnedAppsIfNeeded() + syncInstances() even when the dock has zero instances, which
  is needed so the dock can self-start when a running app appears on a monitor. The null-pointer guards are a good defensive addition. This looks correct but is a meaningful behavior
  change worth testing: the dock should now appear on a previously-empty monitor when an app launches there.

  4. m_popupOwnerInstance reset added to closeAllInstances — not in the original diff context

  The PR adds m_popupOwnerInstance = nullptr; to closeAllInstances(). This looks like a pre-existing bug fix bundled in — previously closeAllInstances would leave m_popupOwnerInstance
  dangling after clearing m_instances.

  Suggestions

  - Settings placement: The monitors entry is placed right after active_monitor_only. These two settings interact — active_monitor_only filters apps shown while monitors filters which 
  displays get a dock surface. The proximity is good, but consider whether the description should mention their interaction (e.g., "Which displays show the dock. Active-monitor filtering
  applies within these displays.").
  - Missing config_export round-trip for empty: config_export.cpp writes monitors via stringArray(). Verify that an empty array round-trips correctly (written as monitors = [] in TOML,
  read back as empty vector).

  Verdict

  Clean, well-scoped PR that reuses established patterns. The detachInstanceState extraction and m_popupOwnerInstance fix are genuine improvements. The parsing style inconsistency is
  minor. The syncInstances-in-refresh change is the main thing to test carefully — verify the dock appears/disappears dynamically as apps launch/close on filtered monitors, and that the
  outputHasAnyDockItems check doesn't cause stutter with many outputs.
  ```

CLEMENTINATOR:
> Thanks a lot, looking pretty good. I've not worked on the dock at all so here is an AI review for you. Take it with a large grain of salt as usual...
> ### Overview
> 
> Adds a `monitors` config key to `DockConfig` (string array of connector names) that limits which displays show the dock. When empty, the dock appears on all outputs (current behavior). Also hides the dock on outputs that have no items to show. Includes a settings GUI entry using the existing `ListSetting` + `availableOutputs` pattern.
> ### What's Good
> 
>     * **Follows existing patterns well.** The `ListSetting` with `suggestedOptions = env.availableOutputs` is the exact same pattern used for notification monitors at
>       `settings_registry.cpp:1263`. The `outputMatchesSelector` reuse is correct.
> 
>     * **`detachInstanceState` is a solid addition.** It properly cleans up dangling pointers (`m_hoveredInstance`, `m_popupOwnerInstance`, surface map entries) before instance destruction,
>       and it's called from both `closeAllInstances` and the erase-if in `syncInstances`. This fixes a potential use-after-free that existed before this PR.
> 
>     * **Translation strings are clean** and the description for `active-monitor-only` is improved ("running on the active monitor" is clearer than "show the dock on the active monitor").
> 
> 
> ### Issues
> 
> **1. Parsing inconsistency — `monitors` vs `pinned` (minor)**
> 
> ```c++
> // pinned (existing):
> if (auto* arr = (*dockTbl)["pinned"].as_array())
>     dock.pinned = readStringArray(*arr);
> 
> // monitors (new):
> if (const auto* v = (*dockTbl).get("monitors"))
>     dock.monitors = readStringArray(*v);
> ```
> 
> pinned guards on .as_array() then dereferences, while monitors calls .get() and passes the toml::node* to readStringArray. Both work — readStringArray internally calls node.as_array() and returns empty if it's not an array — but the monitors style silently accepts a non-array value and returns [], whereas pinned would skip the assignment entirely leaving the default. For a new field where the default is also [], this is functionally equivalent, but using the .as_array() guard would be more consistent with surrounding code.
> 
>     2. outputHasAnyDockItems calls runningAppIds() on every syncInstances tick (medium)
> 
> 
> refresh() now calls syncInstances(), which calls outputAllowed() for every output, which calls outputHasAnyDockItems(), which may call platform.runningAppIds(filter). If refresh() is called frequently (e.g., on every compositor running-app signal), this constructs and scans the running app list multiple times per frame just to decide whether to show the dock surface. The cost is likely small in practice (few outputs, few apps), but worth noting.
> 
>     3. refresh() behavior change — early return semantics shifted (worth verifying)
> 
> 
> Before:
> 
> ```c++
>   void Dock::refresh() {
>       if (m_instances.empty()) return;  // bail if no instances
>       refreshPinnedAppsIfNeeded();
>       // ... request updates
>   }
> ```
> 
> After:
> 
> ```c++
>   void Dock::refresh() {
>       if (m_config == nullptr || m_platform == nullptr || !m_config->config().dock.enabled) return;
>       refreshPinnedAppsIfNeeded();
>       syncInstances();                  // may create or destroy instances
>       if (m_instances.empty()) return;
>       // ... request updates
>   }
> ```
> 
> The old code bailed early if no instances existed (dock not shown). The new code always runs refreshPinnedAppsIfNeeded() + syncInstances() even when the dock has zero instances, which is needed so the dock can self-start when a running app appears on a monitor. The null-pointer guards are a good defensive addition. This looks correct but is a meaningful behavior change worth testing: the dock should now appear on a previously-empty monitor when an app launches there.
> 
>     4. m_popupOwnerInstance reset added to closeAllInstances — not in the original diff context
> 
> 
> The PR adds m_popupOwnerInstance = nullptr; to closeAllInstances(). This looks like a pre-existing bug fix bundled in — previously closeAllInstances would leave m_popupOwnerInstance dangling after clearing m_instances.
> 
> Suggestions
> 
>     * Settings placement: The monitors entry is placed right after active_monitor_only. These two settings interact — active_monitor_only filters apps shown while monitors filters which
>       displays get a dock surface. The proximity is good, but consider whether the description should mention their interaction (e.g., "Which displays show the dock. Active-monitor filtering
>       applies within these displays.").
> 
>     * Missing config_export round-trip for empty: config_export.cpp writes monitors via stringArray(). Verify that an empty array round-trips correctly (written as monitors = [] in TOML,
>       read back as empty vector).
> 
> 
> Verdict
> 
> Clean, well-scoped PR that reuses established patterns. The detachInstanceState extraction and m_popupOwnerInstance fix are genuine improvements. The parsing style inconsistency is minor. The syncInstances-in-refresh change is the main thing to test carefully — verify the dock appears/disappears dynamically as apps launch/close on filtered monitors, and that the outputHasAnyDockItems check doesn't cause stutter with many outputs.

Looks like everything is good then ? :grin:

Ly-sec:
Hey @CLEMENTINATOR I'll merge this PR now, looks good to me.
Thank you very much :heart: !

PR #2731, opened 2026-05-19
refactor(tray_menu): Remove custom scrollbar and use scoll_views

Mathew-D:
Removed the gutter for the scrollbar and use scroll_views new bar


ItsLemmy:
Thanks

PR #2734, opened 2026-05-19
feat(notification_manager): don't track dismissed or low urgency notifications

Ape:
## Motivation
I guess this is mostly personal preference, but I wanted to share the commit since to me it's quite obviously better default behavior.

This could be further enhanced with adjustable settings, but I think this is already just better than the previous version.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring

## Testing
- [ ] Tested on niri
- [x] Tested on Hyprland
- [ ] Tested on sway

## Checklist
- [x] Self-reviewed my code

### DIFF
--- a/src/notification/notification_manager.cpp
+++ b/src/notification/notification_manager.cpp
@@ -70,7 +70,14 @@ namespace {
            notification.body == body;
   }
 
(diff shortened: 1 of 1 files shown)


ItsLemmy:
Thanks, I like it.  Makes sense to me to skip the low prio stuff.

PR #2735, opened 2026-05-19
net: add IWD backend with passphrase agent support

rXelelo:
# Pull Request

## Motivation

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring

## Related Issue

## Testing

- [x] Tested on niri
- [ ] Tested on Hyprland
- [ ] Tested on sway

## Screenshots / Videos

## Checklist
- [x] Self-reviewed my code

## Additional Notes

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -354,6 +354,8 @@ _noctalia_sources = files(
   'src/dbus/network/network_manager_service.cpp',
   'src/dbus/network/network_secret_agent.cpp',
   'src/dbus/network/wpa_supplicant_service.cpp',
+  'src/dbus/network/iwd_service.cpp',
+  'src/dbus/network/iwd_agent.cpp',
   'src/dbus/polkit/polkit_agent.cpp',
   'src/dbus/notification/notification_poll_source.cpp',
   'src/dbus/notification/notification_service.cpp',
--- a/src/app/application.cpp
+++ b/src/app/application.cpp
@@ -8,6 +8,8 @@
 #include "core/log.h"
 #include "core/process.h"
 #include "core/resource_paths.h"
+#include "dbus/network/iwd_agent.h"
+#include "dbus/network/iwd_service.h"
 #include "dbus/network/network_manager_service.h"
 #include "dbus/network/wpa_supplicant_service.h"
 #include "i18n/i18n.h"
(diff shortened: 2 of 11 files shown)


ItsLemmy:
## wirelessEnabled is wrong when the adapter is powered off

  next.wirelessEnabled = !m_stations.empty();

  This returns true as long as any station object exists, regardless of whether the device is actually powered on. A station object persists even when Device.Powered is false. This should check the Powered property on the device interface:

```
  bool anyPowered = false;
  for (const auto& [path, proxy] : m_stations) {
      anyPowered = anyPowered || getPropertyOr(*proxy, k_deviceInterface, "Powered", false);
  }
  next.wirelessEnabled = anyPowered;
```

Without this, setWirelessEnabled(false) will toggle the device off, but wirelessEnabled will still report true on the next rebuildState().

##  submitPasswordPrompt has incorrect dispatch when both agents exist

  The diff in network_tab.cpp adds:
```
    } else if (m_iwdAgent != nullptr) {
      m_iwdAgent->submitSecret(value);
    } 
```

  But this branch is only reached when m_pendingAccessPoint has no value AND m_secrets is null. In practice, if iwd is active, m_secrets should be null  (NetworkSecretAgent is only created when supportsSecretAgent() returns true, which IwdService doesn't). So the logic happens to work — but it's fragile.      If someone adds supportsSecretAgent() to IwdService later, the m_secrets branch would swallow iwd agent requests.

  A clearer approach: check which agent actually has a pending request:
```
    } else if (m_secrets != nullptr && m_secrets->hasPendingRequest()) {
      m_secrets->submitSecret(value);
    } else if (m_iwdAgent != nullptr && m_iwdAgent->hasPendingRequest()) {
      m_iwdAgent->submitSecret(value);
    } 
```

 ## cancelPasswordPrompt has the same fragility

  Same else if chain — m_iwdAgent->cancelSecret() is only reached if m_secrets is null.

##  Synchronous GetOrderedNetworks in rebuildState()

  rebuildState() is called on every PropertiesChanged signal from every station. Each call does a synchronous GetOrderedNetworks() D-Bus method call for each station. On systems with many visible networks or multiple adapters, this could cause noticeable latency on the main loop. The other backends (NM,
  wpa_supplicant) use cached/signal-driven data. Consider debouncing or caching.

## Synchronous Connect call in activateAccessPoint

```
  it->second->callMethod("Connect").onInterface(k_networkInterface);
```

Network.Connect() in iwd can block for several seconds (especially when the agent prompt is involved). This is a synchronous call on the main thread. The other backends may have the same issue, but it's worth noting — this will freeze the UI during connection attempts.

## rebuildState() clears m_knownNetworks on every call

```
  m_knownNetworks.clear();
```

This rebuilds known networks from network objects only. But network objects for out-of-range saved networks won't be present, so hasSavedConnection() will return false for any saved network that isn't currently visible. forgetSsid() won't work for those either. To properly enumerate saved networks, you'd need to query net.connman.iwd.KnownNetwork objects separately.

## Logging SSID in onRequestPassphrase
```
  kLog.info("RequestPassphrase for ssid=\"{}\" path={}", ssid, std::string(netPath));
```  
 We try to avoid login any sensible/private data in the log.

 ## Missing InterfacesRemoved cleanup for networks used by known networks

When InterfacesRemoved fires, both m_stations and m_networks are cleaned, but m_knownNetworks entries referencing removed network paths are left dangling until the next rebuildState(). This is benign since rebuildState() clears + rebuilds, but the ordering in the signal handler (erase then rebuildState) means it's fine.

##  Minor Style Notes
  - k_iwdBusName, k_agentManagerPath etc..., should be kIwdBusName per project's convention
  - ifaces.count() → ifaces.contains() (C++20 is available).

PR #2736, opened 2026-05-19
Small fixes for icon resolver, network tab and translation

notiant:
1. Fixing an issue with the icon resolver I experience with my icon pack: an app that has the icon "printer" will show the symbolic icon in actions/32 instead of the normal icon in apps/scalable. To solve this, I have added a comparision for the `Size` entry if `MaxSize` is equal.
2. If there is no active connection in the network tab, the "Current Connection" box is empty instead of showing the intended "Not connected" row.
3. The launcher icon in the dock is now centered instead of aligned to the top (necessary because it's scaled down to 0.8)
4. I have changed some translation keys for more consistency.

### DIFF
--- a/assets/translations/en.json
+++ b/assets/translations/en.json
@@ -29,7 +29,7 @@
     },
     "empty": {
       "history-title": "Clipboard history is empty",
-      "history-message": "Clipboard history is empty.",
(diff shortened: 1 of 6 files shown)


ItsLemmy:
Thanks

PR #2748, opened 2026-05-20
Feat/custom hex widget colors

mattestanka:
# Pull Request

  ## Motivation

  Users currently can't theme an individual widget with a one-off color without redefining a palette slot in their colorscheme — which then bleeds into every widget bound to that slot. This PR adds first-class per-widget custom hex colors, selectable
  directly from the existing widget settings color picker.

  Two small, composable changes plus an i18n key:

  1. **`Commons/Color.qml`** — `resolveColorKey`, `resolveOnColorKey`, and `resolveColorKeyOptional` short-circuit and return the input unchanged when it starts with `#`. Any setting already routed through these helpers now accepts `"#rrggbb"` /
  `"#rrggbbaa"` in addition to the existing palette keys (`primary` / `secondary` / `tertiary` / `error` / `none`).

  2. **`Widgets/NColorChoice.qml`** — appends a rainbow-gradient "custom color" circle after the palette circles. Clicking it opens the existing `NColorPickerDialog`; the picked color is converted to `#rrggbb` and emitted via the standard `selected`
  signal. Every widget settings panel that already uses `NColorChoice` (Volume, Network, Clock, Workspace, Battery, NotificationHistory, SystemMonitor, …) gains this with zero per-widget changes.

  3. **`Assets/Translations/en.json`** — adds `common.custom-color` for the new tooltip; other locales fall back to en via the existing `I18n.tr` fallback chain until translations are contributed.

  Existing palette keys keep working identically. Settings files written before this PR are untouched — the new circle is purely additive. No new dependencies, no settings-schema changes, no migrations.

  ## Type of Change
  - [ ] Bug fix
  - [x] New feature
  - [ ] Breaking change
  - [ ] Refactoring

  ## Related Issue
  - N/A

  ## Testing
  - [ ] Tested on niri
  - [ ] Tested on Hyprland
  - [ ] Tested on sway

  Tested on **mangowc** (not listed in the template). Verified:
  - New rainbow circle appears at the end of every widget's color row in settings
  - Picking a color via `NColorPickerDialog` saves as `"#rrggbb"` in `settings.json` and renders correctly on the widget
  - Switching back to a palette circle (primary/secondary/…) restores the palette behavior
  - Manually editing `settings.json` to set e.g. `"iconColor": "#91d7e3"` renders correctly without touching the UI
  - Existing settings load with no behavior change
  - Confirmed across two monitors with different bar positions (top / bottom) and density settings (spacious / comfortable)

  Not yet tested on niri / Hyprland / sway or at non-default UI scale — happy to validate on those if a maintainer can point me at a quick way to spin them up, or if a reviewer with that setup can spot-check.

  ## Screenshots / Videos
image

image

  ## Checklist
  - [x] Self-reviewed my code

  (No user-facing docs found in the repo for the color picker / widget settings; happy to add a note somewhere if maintainers point me at the right place.)

  ## Additional Notes

  None.

### DIFF
--- a/Assets/Translations/en.json
+++ b/Assets/Translations/en.json
@@ -424,6 +424,7 @@
     "contributors": "Contributors",
     "copied-to-clipboard": "Copied to clipboard",
     "countdown": "Countdown",
+    "custom-color": "Custom color",
     "customize": "Customize",
     "date": "Date",
     "default": "Default",
--- a/Commons/Color.qml
+++ b/Commons/Color.qml
@@ -303,6 +303,8 @@ Singleton {
   }
 
   function resolveColorKey(key) {
+    if (key && key.length > 0 && key.charAt(0) === "#")
+      return key;
     switch (key) {
     case "primary":
       return root.mPrimary;
@@ -318,6 +320,8 @@ Singleton {
   }
 
   function resolveOnColorKey(key) {
+    if (key && key.length > 0 && key.charAt(0) === "#")
+      return key;
     switch (key) {
     case "primary":
       return root.mOnPrimary;
@@ -333,6 +337,8 @@ Singleton {
   }
 
   function resolveColorKeyOptional(key) {
+    if (key && key.length > 0 && key.charAt(0) === "#")
+      return key;
     switch (key) {
     case "primary":
       return root.mPrimary;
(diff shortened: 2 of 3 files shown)


ItsLemmy:

We are currently pouring all of our focus into v5. Because v4 is now frozen, we won’t be merging further changes into that codebase.

While we are closing this PR, we really value your effort and would love to see your work in the v5 ecosystem. If you’d like to port this over to the new version, we’d be happy to review it there!

Cheers!


PS: This is fixed in v5 every color field can accept a custom color

PR #2755, opened 2026-05-21
bar: add panel_background_opacity to detach attached panels from bar alpha

mattestanka:
## Motivation

Attached panels (the clock dropdown, control center, wallpaper picker, session menu, etc.) currently inherit their host bar's `background_opacity`. That means setting the bar to 0% alpha — a popular "floating capsules only" look — also collapses every attached panel's background to invisible, leaving panel content floating with no surface behind it. There was no way to keep the bar transparent while preserving legible panel chrome.

Detached/centered panels already escape this via `shell.panel.transparency_mode` returning fixed alpha per mode; only attached panels were stuck inheriting the bar.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring

Adds an optional per-bar `panel_background_opacity` (with matching per-monitor override) under `[bar.*]`. When unset, behavior is unchanged — attached panels keep inheriting `background_opacity`. When set, it replaces the inherited value as the source for the panel's background fill and the seed passed into `panelCardOpacityForTransparencyMode`, so panel cards and the contact-shadow gradient stay consistent.

Touched files:
- `src/config/config_types.h` — new optional on `BarConfig` + `BarMonitorOverride`
- `src/config/config_service.cpp` — TOML parser for both scopes and resolution in `resolveForOutput`
- `src/config/config_overrides.cpp` — equality + override-apply
- `src/config/config_export.cpp` — write the field back out (only when set)
- `src/shell/panel/panel_manager.cpp` — swap the inheritance source at both `m_attachedBackgroundOpacity` assignment sites (initial open + live config reload)
- `src/shell/settings/settings_registry.cpp` — slider entry in the bar Shape section (single-bar + per-monitor)
- `assets/translations/en.json` — label + description

## Testing
- [ ] Tested on niri
- [ ] Tested on Hyprland
- [ ] Tested on sway

Verified manually: with `bar.main.background_opacity = 0.0` and `bar.main.panel_background_opacity = 0.9`, the bar is fully invisible while the clock dropdown, control center, and other attached panels render with a normal solid surface. Toggling the new slider live updates open panels via the existing `onConfigReloaded` path. With the new field unset, behavior matches `main` exactly.

`just build release` clean, `meson test` 9/9 green, `just format` applied.

### DIFF
--- a/assets/translations/en.json
+++ b/assets/translations/en.json
@@ -1429,6 +1429,9 @@
         "background-opacity": {
           "label": "Background Opacity"
         },
+        "panel-background-opacity": {
+          "label": "Panel Background Opacity"
+        },
         "shadow": {
           "label": "Shadow"
         },
@@ -2233,6 +2236,9 @@
         },
         "background-opacity": {
           "description": "Background opacity of the bar"
+        },
+        "panel-background-opacity": {
+          "description": "Opacity for panels attached to this bar"
         },
         "border": {
           "label": "Border",
--- a/src/config/config_export.cpp
+++ b/src/config/config_export.cpp
@@ -152,6 +152,9 @@ namespace config_export {
       table.insert_or_assign("reserve_space", bar.reserveSpace);
       table.insert_or_assign("thickness", static_cast<std::int64_t>(bar.thickness));
       table.insert_or_assign("background_opacity", static_cast<double>(bar.backgroundOpacity));
+      if (bar.panelBackgroundOpacity.has_value()) {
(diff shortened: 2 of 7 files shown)


ItsLemmy:
This is the kind of thing I want to avoid (settings creep) - when the bar is fully transparent there is no point attaching any panel to it as it will look ugly.   Or am I missing something ?

mattestanka:
Okay, I see what you mean. It's not an issue with floating windows.

PR #2758, opened 2026-05-22
Better names in battery tooltip

notiant:
1. My laptop battery is currently displayed as "0x4C 0x32 0x32 0x4D 0x34 0x50 0x41 0x35", it's better to just use "Battery" in this case. This PR also adds naming support for multiple batteries.
2. Hiding some text in the clipboard panel for no results/history since this message is currently shown 3 times.

### DIFF
--- a/assets/translations/en.json
+++ b/assets/translations/en.json
@@ -16,7 +16,7 @@
     "title": "Clipboard",
     "filter-placeholder": "Filter clipboard...",
     "entry": {
-      "title": "Clipboard entry",
+      "title": "Clipboard Entry",
       "image": "Image"
     },
     "preview": {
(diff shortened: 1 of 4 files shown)


ItsLemmy:
merging it, but I have a few comments: 
- "session menu" is not a thing.  In v5 it's "session panel"
- I dont see why Control Center should stay capitalized if all other panels are not.

notiant:
> merging it, but I have a few comments:
> 
>     * "session menu" is not a thing.  In v5 it's "session panel"
> 
>     * I dont see why Control Center should stay capitalized if all other panels are not.

Since you were capitalizing Control Center in most cases I thought that this is a stylistic choice to refer to it as a proper noun like it's done in iOS for example (so "in Control Center" as opposed to "in the Control Center"). Not sure if this works with the other panels, though. There's also the issue that if you translate with an LLM you'll most likely get random and inconsistent capitalizations.

PR #2767, opened 2026-05-22
Add Triad compositor backend

greenm01:
Adds a Triad backend.

Triad is a programmable window manager built on River. It has its own IPC, but also exposes a Niri-compatible facade so shells like Noctalia can already run on it without changes. This backend uses the native IPC instead, which is cleaner and avoids treating Triad as fake Niri inside Noctalia.

I think it is worth adding because the shared Noctalia changes are small, and Triad users get a better integration path than the compatibility shim. Workspace clicks, overview state, monitor power, keyboard layout, spawn, and window updates can all go through the compositor backend directly.

Triad is a little different from most compositors Noctalia supports because each workspace can use a different layout model. The backend reads workspace, window, output, overview, keyboard layout, and capability state directly from Triad, then maps that into the normal compositor service shape.

I know v5 is where most work is happening. Happy to move this there if that makes more sense.

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring

Tested on Triad with three monitors. Workspace switching, window focus updates, overview bar hiding, keyboard layout, spawn, monitor power, and workspace content scrolling all work here.

- [ ] Tested on niri
- [ ] Tested on Hyprland
- [ ] Tested on sway

No screenshots. It should look the same.

- [x] Self-reviewed my code

Triad: https://github.com/greenm01/triad

### DIFF
--- a/Modules/Background/Overview.qml
+++ b/Modules/Background/Overview.qml
@@ -8,7 +8,7 @@ import qs.Services.Power
 import qs.Services.UI
 
 Loader {
-  active: CompositorService.isNiri && Settings.data.wallpaper.enabled && Settings.data.wallpaper.overviewEnabled && (!PowerProfileService.noctaliaPerformanceMode || !Settings.data.noctaliaPerformance.disableWallpaper)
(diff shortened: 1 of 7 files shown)


ItsLemmy:
Cool project, but yeah we don't accept any PR on v4 anymore.
Feel free to open one for v5 if you wanna give it a try.

greenm01:
> Cool project, but yeah we don't accept any PR on v4 anymore. Feel free to open one for v5 if you wanna give it a try.

Ah, thanks! Will do.

PR #2769, opened 2026-05-23
Improve dock window switching

DFSko:
## Summary
- Left-clicking a running dock app now cycles through its open windows.
- The right-click dock menu now shows open windows first, followed by the existing app actions.
- Removed the old separate window picker path.

## Test plan
- meson test -C build --print-errorlogs


ItsLemmy:
# Overview

  Replaces the left-click multi-window popup picker with inline window cycling: clicking a dock item with multiple windows activates the next one in round-robin order, remembering the last-active window per app. The right-click context menu now lists open windows at the top (above desktop actions and Close/Close All), so users still have a way to target a specific window.

  The old openWindowPicker / closeWindowPicker / m_windowMenu infrastructure is fully removed (~220 lines deleted).

  # Positive aspects

  - Significant simplification: Removing the entire separate popup surface for the window picker is a big maintenance win — one popup type instead of two,
  no dual routing in onPointerEvent.
  - Named constants for menu IDs: kMenuCloseId, kMenuCloseAllId, kMenuSeparatorId, kMenuWindowBaseId replace magic -1, -2, -3 literals. Nice cleanup.
  - Clean cycling logic: nextActivatableWindowHandle is straightforward and handles the edge cases (no active match, preferred handle, first-available
  fallback).

  # Issues

  ## 1. Dangling pointer in m_lastActiveHandleByAppIdLower (Bug)

  This is the main concern. The map stores raw zwlr_foreign_toplevel_handle_v1* pointers. When a toplevel is closed, WaylandToplevels::onHandleClosed calls   zwlr_foreign_toplevel_handle_v1_destroy(handle) and erases it from its own map — but nobody purges m_lastActiveHandleByAppIdLower. The stale pointer  will sit there until the same app is focused again.

  The cycling code does guard against this somewhat: nextActivatableWindowHandle compares its preferredHandle argument against live windows[i].handle values, so it won't activate a dangling pointer directly. But the comparison itself is technically undefined behavior (comparing a pointer to a destroyed object), and if the allocator reuses that address for a new toplevel, the wrong window could be activated.

  Suggestion: Either subscribe to toplevel-close events and erase the entry, or clear the map in closeAllInstances / on config reload / whenever the dock rebuilds. Alternatively, store something with a stable identity (like the ToplevelInfo itself) rather than a raw protocol handle.

 ## 3. Single-window left-click changed from direct-activate to cycling

  The old code had an explicit fast path: if (windows.size() == 1) { activateToplevel(windows[0].handle); return; }. The new code goes through
  nextActivatableWindowHandle which will still activate the single window, but it's doing more work (checking active handle, preferred handle, iterating).
  Functionally correct, but the explicit single-window case was arguably clearer. Minor.

  ## 5. Missing cleanup in closeAllInstances

  closeAllInstances resets m_itemMenu (and in the old code, m_windowMenu). It doesn't clear m_lastActiveHandleByAppIdLower. Since all dock instances are  being destroyed (output change, config reload), the handle cache should arguably be cleared too — those handles may become invalid.


  Summary

  The refactor is a good simplification overall. The dangling-pointer risk in m_lastActiveHandleByAppIdLower is the one issue I'd want addressed before
  merge — at minimum, clear it in closeAllInstances, and ideally add a mechanism to erase entries when toplevels are destroyed. The rest is clean.

DFSko:
Addressed the review items:

- Dangling cached handle: the dock now prunes cached wlr toplevel handles during the existing toplevel-change refresh path instead of adding a separate close callback. The cache checks handles through `containsWlrToplevelHandle()` and drops entries that are no longer live.
- `closeAllInstances`: now clears `m_lastActiveHandleByAppIdLower` when dock instances are torn down.
- Single-window click: restored the explicit fast path, so a single matching window is activated directly before the cycling logic runs.

Verification: `ninja -C build` and `meson test -C build` both pass.

Ly-sec:
Hey @DFSko this looks good to me after the changes. Sorry for the delay on my end!
I'll be merging this one, thank you very much.

PR #2770, opened 2026-05-23
Add Triad compositor backend

greenm01:
Adds a Triad backend for v5.

Triad is a programmable window manager built on River. It has its own IPC, but also exposes a Niri-compatible facade so shells like Noctalia can already run on it without changes. This backend uses the native IPC instead, which is cleaner and avoids treating Triad as fake Niri inside Noctalia.

I think it is worth adding because the shared Noctalia changes are small, and Triad users get a better integration path than the compatibility shim. Workspace clicks, overview state, monitor power, keyboard layout, focused window state, and window updates can all go through the compositor backend directly.

Triad is a little different from most compositors Noctalia supports because each workspace can use a different layout model. The backend reads workspace, window, output, overview, and keyboard layout state directly from Triad, then maps that into the normal compositor service shape.

Tested on Triad with three monitors. Workspace switching, window focus updates, overview state, keyboard layout, monitor power, taskbar state, and workspace content updates all work here.

Triad: https://github.com/greenm01/triad


ItsLemmy:
Could not get Triad to run on any of my machines to test the PR.
I just get the mouse cursor, and none of the keybinds are working.
I'm on Fedora 44, using the default config.

image

Tried to autostart noctalia and that did work either, as if the configuration was just not loaded.
Am I missing something?

greenm01:
What does the session log say?

tail -50 ~/.local/state/triad/river-triad-session-latest.log

Do you have Nim 2.2.4+?

Fedora packages might be older than that.

Did you compile River 0.4+?

https://triadwm.org/getting-started/installation/

That log should give good info. 

Thanks for giving it a go.

greenm01:
I stripped the default config of all assumptions and added a Fedora section to the install wiki. Hope this makes it easier to get going.

[Install Wiki](https://triadwm.org/getting-started/installation/)

[New default config](https://github.com/greenm01/triad/blob/master/config.default.kdl)

My own config example is here, with all my local apps:
[My personal config](https://github.com/greenm01/triad/blob/master/examples/config/niltempus_config.kdl)

All my shells in the example have:

`niri-compat #true`

This allows Triad to communicate with shells via the Niri IPC protocol until native Triad IPC is implemented (the aim of this PR).

Session log and behavior logs:

```
ls -la ~/.local/state/triad/
tail -n 200 ~/.local/state/triad/river-triad-session-latest.log
tail -n 200 ~/.local/state/triad/triad-latest.log
```
Cheers!

ItsLemmy:
The issue was my river binary was 3-4 weeks old.
- I had  River 0.5.0-dev.15+7c9f32b and it did not work
- It works with 0.5.0-dev.68+da8cf20 

FWIW Fedora 44 packages are recent enough.

Package "nim-2.2.10-1.fc44.x86_64" is already installed.
Package "nim-2.2.10-1.fc44.x86_64" is already installed.
Package "pkgconf-2.5.1-1.fc44.x86_64" is already installed.
Package "wayland-devel-1.24.0-3.fc44.x86_64" is already installed.
Package "libxkbcommon-devel-1.13.1-2.fc44.x86_64" is already installed.
Package "pixman-devel-0.46.2-3.fc44.x86_64" is already installed.

greenm01:
Nice! 

I spent the last few hours on the install instructions to make it (hopefully) more clear:

https://triadwm.org/getting-started/installation/

Note: to try my PR with native Triad IPC, in your config set:

niri-compat #false

e.g.

```
shells {
  enabled #true
  active "noctalia"
  cycle "noctalia"

  watchdog {       // this is a fallback shell, e.g. waybar or whatever
    enabled #true
    fallback "noctalia"
  }

  profile "noctalia" {
    launch "noctalia-shell"  // or whatever the binary is
    stop "pkill" "-f" "noctalia-shell"
    niri-compat #false
  }
}
```

You can dynamically switch shells in the middle of a session. See my config example in the repo for full setup.

ItsLemmy:
Thanks

PR #2772, opened 2026-05-23
feat(taskbar): add icon opacity and active indicator toggle

linusammon:


ItsLemmy:
inactive_opacity = 0.0 makes icons invisible but still clickable

I think we should enforce a minimum of 0.1

PR #2773, opened 2026-05-23
feat(power): add logind PrepareForSleep monitoring to lock screen on external suspend

john-okeefe:
# Pull Request

## Motivation

Noctalia's lockscreen only activated on suspend when the suspend was initiated from **within** the shell (idle timeout, session menu, launcher). External suspend triggers — laptop lid close, `systemctl suspend`, power button, or any other process calling logind — bypassed the lock entirely, leaving the desktop unauthenticated after resume.

This PR adds a new `LoginService` that monitors logind's `PrepareForSleep` D-Bus signal to lock the screen before the system suspends, regardless of what triggered it. A secondary fallback via the existing `Time.resumed` heuristic locks on resume if the D-Bus signal was missed.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring

## Testing
- [x] Tested on niri
- [ ] Tested on Hyprland
- [ ] Tested on sway

### Testing performed

**Void Linux (elogind) with niri:**
- Verified `org.freedesktop.login1.Manager.PrepareForSleep` signal is emitted by elogind via `busctl`
- Verified `dbus-monitor --system` captures the signal with eavesdrop fallback
- Verified `LoginService` probes succeed: logind on system bus detected, `dbus-monitor` binary found
- Verified graceful degradation when dependencies are missing (clean log messages, no crashes)

## Screenshots / Videos

N/A — no visual changes. The lockscreen activates as expected on suspend.

## Checklist
- [x] Self-reviewed my code

### Architecture

**Layer 1 — LoginService (primary):** Runs `dbus-monitor --system` as a persistent `Process`, parsing stdout for `org.freedesktop.login1.Manager.PrepareForSleep`. On `PrepareForSleep(true)`, activates the lockscreen. On `PrepareForSleep(false)`, emits the signal for future use (e.g., waking services).

**Layer 2 — Time.resumed (fallback):** The existing time-jump heuristic in `Time.qml` detects system resume. The `LockScreen.qml` handler now also activates the lockscreen if `lockOnSuspend` is enabled and the screen isn't already locked — catching cases where the D-Bus signal was missed or logind is unavailable.

### Compatibility

- **elogind** (Void, Alpine, Artix) — tested, works
- **systemd-logind** (Fedora, Arch, NixOS, Ubuntu) — same D-Bus interface, works
- **Systems without logind** — LoginService probes gracefully, logs a clear message, and defers to the `Time.resumed` fallback only

### Robustness

- Two-phase init probe checks for logind on bus AND `dbus-monitor` binary before starting
- Bounded restart: if monitor exits without ever receiving a signal, retries up to 3 times then stops (no infinite loop on systems without logind)
- Successful signal reception resets the restart counter
- `Component.onDestruction` stops the monitor timer and cleans up the process

### New files
- `Services/Power/LoginService.qml` — new singleton (~220 lines)

### Modified files
- `shell.qml` — added `LoginService.init()` to deferred startup
- `Modules/LockScreen/LockScreen.qml` — added lock-on-resume fallback in `Time.resumed` handler
h


ItsLemmy:

Cheers!

PR #2774, opened 2026-05-23
shell: add ipc command for setting bar auto-hide state

Ape:
Adds:
```
bar-auto-hide-set <on|off|true|false|1|0> [bar-name] [monitor-selector]  Set auto-hide state for a bar
```

## Motivation
I need this IPC for my personal setup.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring

## Testing
- [x] Tested on Hyprland

## Checklist
- [x] Self-reviewed my code

### DIFF
--- a/src/shell/bar/bar.cpp
+++ b/src/shell/bar/bar.cpp
@@ -315,6 +315,70 @@ namespace {
     float height = 0.0f;
   };
 
+  struct BarSurfaceSpec {
+    std::int32_t marginTop = 0;
+    std::int32_t marginRight = 0;
+    std::int32_t marginBottom = 0;
+    std::int32_t marginLeft = 0;
+    std::uint32_t surfaceWidth = 0;
+    std::uint32_t surfaceHeight = 0;
+    std::int32_t exclusiveZone = 0;
+  };
+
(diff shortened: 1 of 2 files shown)


Ape:
Fun example for hyprland

```lua
hl.bind("SUPER_L", hl.dsp.exec_cmd("noctalia msg bar-auto-hide-set off default HDMI-A-1"))
hl.bind("SUPER + SUPER_L",
	hl.dsp.exec_cmd("noctalia msg bar-auto-hide-set on default HDMI-A-1"),
	{ release = true, transparent = true }
)
```

ItsLemmy:
It's cool, but it's the first IPC command that would actually alter your settings via setOverride.
All other IPC commands just affect runtime and are transient, I believe we should continue to enforce this.

Ape:
> All other IPC commands just affect runtime and are transient

That's actually a good point, and for my use case a transient change to the autohide mode might be even more suitable than actual persisted setting change.

Ape:
Updated.

ItsLemmy:
Nice, thanks

PR #2776, opened 2026-05-23
Battery: make hideWhenFull work with PendingCharge

notiant:
The `hideWhenFull` option should also work if a battery threshold is reached, not just when the battery is at 100%.


Ape:
Just FYI, you can update your existing PR with a force push without closing it and opening a new one.

PR #2777, opened 2026-05-23
config: add notification popup scale setting

Ape:
## Motivation
This is an important setting for me personally as I need bigger notification popups for my TV without scaling up any other shell UI.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring

## Testing
- [x] Tested on Hyprland

## Checklist
- [x] Self-reviewed my code

### DIFF
--- a/assets/translations/en.json
+++ b/assets/translations/en.json
@@ -2205,6 +2205,10 @@
           "label": "Screen Position",
           "description": "Screen position for notification toasts"
         },
+        "scale": {
+          "label": "Size",
+          "description": "Scale notification toasts relative to the shell UI scale"
+        },
         "layer": {
           "label": "Wayland Layer",
           "description": "Choose whether toasts use the top or overlay layer"
--- a/example.toml
+++ b/example.toml
@@ -112,6 +112,7 @@ tint_intensity         = 0.3        # 0.0 = no tint, 1.0 = opaque
 [notification]
 enable_daemon      = true
 layer              = "top"          # top | overlay
+scale              = 1.0            # notification size multiplier applied on top of shell.ui_scale
 background_opacity = 0.97
 offset_x           = 20             # absolute horizontal margin from the screen edge
 offset_y           = 8              # absolute vertical margin from the screen edge
--- a/src/config/config_export.cpp
+++ b/src/config/config_export.cpp
@@ -574,6 +574,7 @@ namespace config_export {
       table.insert_or_assign("enable_daemon", notification.enableDaemon);
       table.insert_or_assign("position", notification.position);
       table.insert_or_assign("layer", notification.layer);
+      table.insert_or_assign("scale", static_cast<double>(notification.scale));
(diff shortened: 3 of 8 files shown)


ItsLemmy:
## 1. Missed kGap scaling in collision checks (bug)

  Lines 1272 and 1307 use bare kGap in displaceOverlapping() and canKeepPlacement(). The PR scales kGap everywhere else (the 4 uses in findPlacementY, the  onConfigReload block, fallbackSurfaceHeight), but misses these two overlap/separation checks. At scale > 1.0, toasts could be placed with gaps smaller  than the visual intent; at scale < 1.0, toasts that should fit could be spuriously evicted.

  ## 2. Repeated notificationUiScale(m_config) calls in hot paths

  findPlacementY calls notificationUiScale(m_config) 4 times in a single invocation (lines ~1502–1522 in the diff). The function reads two config fields   and does a multiply+clamp each time. It's cheap but noisy — a single const float scale = notificationUiScale(m_config); at the top of findPlacementY  would be cleaner and consistent with how buildCard, onConfigReload, and refreshEntryGeometry already do it.

  Same applies to entryHeight(), maxPlacementBottom(), layoutBottomForSurfaceHeight(), surfaceHeightForOutput(), cardReveal(), applyCardReveal(),   resolveNotificationIconPath(), and updateInputRegion() — each calls notificationUiScale(m_config) inline. Consider caching the effective scale as a  member (updated in onConfigReload/ensureSurfaces) instead of recomputing it ~30+ times per frame.

  ## 3. onConfigReload relayout duplicates logic from existing placement code

  The new 50-line block in onConfigReload() reimplements the top-stacking and bottom-stacking placement logic that already exists in findPlacementY() and   the reveal flow. If the placement algorithm changes, this copy must be updated in lockstep. Consider whether onConfigReload could instead invalidate all   placements and call revealQueuedEntries() to re-place them through the existing path — or extract a shared relayout helper.

  ## 4. kPaddingTop is 0.0f — intentionally unscaled?

  kPaddingTop is 0 so scaling is a no-op currently, but if it ever becomes nonzero it'll be missed since there's no scale-aware helper for it. A //   intentionally unscaled: always 0 comment or a paddingTop(scale) helper would prevent future bugs.

Ape:
Made some changes in reaction to all of those review points.

ItsLemmy:
Almost there: 

## Inconsistent scale capture in pointer callbacks (minor bug risk)

setOnEnter reads scale live via notificationUiScale(m_config) (through this), while setOnMotion captures scale by value at card build time. If scale changes via hot reload, onEnter will use the new scale for hit-testing the close button, but onMotion will use the old scale. The card's onClick lambda  also captures scale by value. These should be consistent — either all live or all captured. Since the card geometry (close button position/size) is baked at build time, capturing by value is correct for all three.

it's a one-line change to capture scale by value in setOnEnter instead of calling notificationUiScale live

Ape:
> Almost there

Fixed.

PR #2779, opened 2026-05-23
shell/settings: move max width inside content scroll

Ape:
## Motivation
The max width made the settings window look worse in tiling wm. See screenshots for comparison.

## Testing
- [x] Tested on Hyprland

### Before
image

### After
image

## Checklist
- [x] Self-reviewed my code

### DIFF
--- a/src/shell/settings/settings_window_scene.cpp
+++ b/src/shell/settings/settings_window_scene.cpp
@@ -74,18 +74,6 @@ namespace {
     });
   }
 
-  std::unique_ptr<Flex> centeredRow(std::unique_ptr<Flex> child, float scale) {
-    child->setFlexGrow(1.0f);
-    child->setMaxWidth(kBodyMaxWidth * scale);
-    return ui::row(
-        {
-            .align = FlexAlign::Stretch,
-            .justify = FlexJustify::Center,
-        },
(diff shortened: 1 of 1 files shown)


ItsLemmy:
This is by design / on purpose so the controls are not totally disconnected from their label.
Was like this in v4 too. 

I'm fine tweaking the maximum a bit, but probably not going back to full stretch.

Ape:
Could we make the scrolling container 100% wide so that the scroll bar is on the right edge of the window and you can scroll wheel anywhere on the window? We can then limit the max width of the inner container with the actual content.

ItsLemmy:
sure

Ape:
Like this.

image

ItsLemmy:
Then I would make the header (filter, advanced, modified, etc..) match the main content width

Ape:
I have these version for you

### v3 (currently on the PR)
image

### v4 (alternative idea)
image

ItsLemmy:
None of them really clicked with me, so I went back to the original idea of removing the clamping.
Up to the user to size his window accordingly.  It's probably the best  long term solution.

Ape:
This will also be a non-issue if the settings window is a panel attached to the bar and sized by us.

PR #2783, opened 2026-05-24
osd: add settings for margins

Ape:
Adds a config setting for OSD margins similar to what notifications have.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring

## Testing
- [x] Tested on Hyprland
- [x] Tested with different bar positions

## Checklist
- [x] Self-reviewed my code
- [x] Documentation or comments updated

### DIFF
--- a/assets/translations/en.json
+++ b/assets/translations/en.json
@@ -1677,6 +1677,14 @@
           "label": "Size",
           "description": "Scale OSD popups relative to the shell UI scale (1.0 keeps the default size)"
         },
+        "osd-offset-x": {
+          "label": "Horizontal Margin",
+          "description": "Absolute horizontal margin from the screen edge"
+        },
+        "osd-offset-y": {
+          "label": "Vertical Margin",
+          "description": "Absolute vertical margin from the screen edge"
+        },
         "osd-lock-keys": {
           "label": "Lock Keys",
           "description": "Show OSD popups for Caps Lock, Num Lock, and Scroll Lock changes"
--- a/example.toml
+++ b/example.toml
@@ -126,6 +126,8 @@ offset_y           = 8              # absolute vertical margin from the screen e
 position = "top_right"              # top_right | top_left | top_center | bottom_right | bottom_left | bottom_center | center_right | center_left
 orientation = "horizontal"          # horizontal | vertical
 scale = 1.0                         # OSD size multiplier applied on top of shell.ui_scale
+offset_x = 20                       # absolute horizontal margin from the screen edge
+offset_y = 8                        # absolute vertical margin from the screen edge
 lock_keys = true                    # show Caps/Num/Scroll Lock OSD popups; disable to stop polling when no widget uses them
 keyboard_layout = true            # show an OSD popup when the input keyboard layout changes
 
--- a/src/config/config_export.cpp
+++ b/src/config/config_export.cpp
@@ -677,6 +677,8 @@ namespace config_export {
     osd.insert_or_assign("position", config.osd.position);
     osd.insert_or_assign("orientation", config.osd.orientation);
     osd.insert_or_assign("scale", static_cast<double>(config.osd.scale));
+    osd.insert_or_assign("offset_x", static_cast<std::int64_t>(config.osd.offsetX));
+    osd.insert_or_assign("offset_y", static_cast<std::int64_t>(config.osd.offsetY));
     osd.insert_or_assign("lock_keys", config.osd.lockKeys);
     osd.insert_or_assign("keyboard_layout", config.osd.keyboardLayout);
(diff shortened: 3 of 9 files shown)


ItsLemmy:
##  Double state callback in close() dismiss path

In close(), the PR captures hadUnreadBefore at the top (line ~359 in the PR), then on the Dismissed path calls removeHistoryEntry(id, reason). But  removeHistoryEntry now also captures its own hadUnreadBefore and calls notifyUnreadStateChangedIfNeeded internally. When close() then falls through to its own notifyUnreadStateChangedIfNeeded(hadUnreadBefore) at the bottom, it re-evaluates against the original stale hadUnreadBefore:

  close():
    hadUnreadBefore = true  (captured)
    removeHistoryEntry(id)  → fires callback (true→false transition)
    notifyUnreadStateChangedIfNeeded(true)  → current=false, fires callback AGAIN

This fires m_stateCallback twice for a single dismiss. The fix would be to skip the bottom notifyUnreadStateChangedIfNeeded when the dismiss path was  taken (since removeHistoryEntry already handles it), or to remove the notification from close() and rely solely on removeHistoryEntry's internal check for that branch.

Ape:
> Double state callback in close() dismiss path

Wrong PR :)

This is clearly about #2785.

ItsLemmy:
Ahaha, damn need more coffee I guess xD

  Issues & Suggestions

##  1. osdSurfaceWidth change looks like an unrelated fix bundled in

  - const float w = cardWidth(s, orientation) + Style::spaceMd * s;
  + const float w = cardWidth(s, orientation) + Style::spaceMd * s * 2.0f;

  This doubles the horizontal padding in the surface buffer. The commit message doesn't mention this, and it's separate from the margin feature. It changes
   the surface allocation for everyone, not just users who configure margins. Was this needed to prevent clipping when the card is centered in the wider
  surface? If so, it should be called out — currently it reads as an accidental regression that makes all OSD surfaces wider.

##  2. horizontalLayerMarginFromScreenMargin math could produce negative values

  std::int32_t horizontalLayerMarginFromScreenMargin(int offsetX, float scale) {
    return static_cast<std::int32_t>(offsetX) - static_cast<std::int32_t>(std::lround(Style::spaceMd * scale));
  }

  If the user sets offset_x = 0 (allowed by the parser: std::max(0, ...)), the result is negative. Layer-shell margins can be negative on some compositors
  but behavior is undefined/compositor-specific. At minimum, clamp to 0 here, or document that small values are intentionally inset.

##  3. cardBaseXForPosition now ignores position entirely

  float cardBaseXForPosition(const std::string& position, float surfaceWidth, float cardW) {
    (void)position;
    return (surfaceWidth - cardW) * 0.5f;
  }

  The parameter and function name are misleading now that position is unused. Consider either:
  - Removing the position parameter (callers pass it for nothing), or
  - Renaming to cardBaseX(float surfaceWidth, float cardW).

  The (void)position suppresses the warning but leaves dead code.

##  4. Comments in config_types.h

  int offsetX = 20; // absolute horizontal margin from the screen edge
  int offsetY = 8;  // absolute vertical margin from the screen edge

  Per project convention, these comments just restate what the name says. The translation strings are the user-facing documentation; code comments here add noise.

##  5. Config equality check in config_overrides.cpp — consider grouping

  The diff reformats the line-wrapping of the equality chain. While the new fields are correctly added, the reflow makes the diff noisier than necessary.  Minor, but a note for future PRs.

  Risks

  - The surface width doubling (point 1) changes visual appearance for all users, even those not using the new settings. This deserves testing at the
  default config to confirm no visual regression.
  - Negative layer-shell margins (point 2) may behave differently across Hyprland/Sway/Niri — worth clamping for safety.
  
  
 ----- 
 
 Is #1 intentionnal ?
 
 This PR would also require a documentation PR update.

Ape:
> 1. osdSurfaceWidth change looks like an unrelated fix bundled in

That change matches the margin with the corresponding setting for notifications. With that, notifications and osd toasts line up exactly when using the same config value.

Ape:
> 2. horizontalLayerMarginFromScreenMargin math could produce negative values

I don't think this is valid. Anyway, it's similar to the margin calculation for notifications.

Ape:
> 5. Config equality check in config_overrides.cpp — consider grouping

Not under my control. I would set these clang-format options to reduce the diff churn on these in the future:
```
AlignOperands: DontAlign
BreakBeforeBinaryOperators: All
BreakBinaryOperations: RespectPrecedence
```

Ape:
Rebased and pushed changes for points 3 and 4.

ItsLemmy:
Thanks

PR #2785, opened 2026-05-24
notification: track unread history with per-entry seen state

Ape:
## Motivation
Previously, the notification widget’s "Hide When No New" behavior was driven by `m_unreadSinceHistoryVisit`. That broke down in two ways:
- The widget could stay visible even after the only new notification had been dismissed or otherwise removed from history.
- If an active notification was marked read by opening the notifications tab, it could become unread again when the popup later expired/closed and its history entry was rewritten.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring

## Testing
- [x] Tested on Hyprland

## Checklist
- [x] Self-reviewed my code

### DIFF
--- a/src/notification/notification_history_store.cpp
+++ b/src/notification/notification_history_store.cpp
@@ -650,6 +650,7 @@ bool loadNotificationHistoryFromFile(const std::filesystem::path& path, std::deq
     NotificationHistoryEntry he;
     he.notification = notificationFromJson(item.at("notification"), path);
     he.active = item.value("active", false);
+    he.seen = item.value("seen", true);
     if (item.contains("close_reason") && !item["close_reason"].is_null()) {
       const auto crs = item["close_reason"].get<std::string>();
       he.closeReason = closeReasonFrom(crs);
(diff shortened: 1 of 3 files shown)


Ape:
> https://github.com/noctalia-dev/noctalia-shell/pull/2783#issuecomment-4528923149
> > ## Double state callback in close() dismiss path
> In close(), the PR captures hadUnreadBefore at the top (line ~359 in the PR), then on the Dismissed path calls removeHistoryEntry(id, reason). But removeHistoryEntry now also captures its own hadUnreadBefore and calls notifyUnreadStateChangedIfNeeded internally. When close() then falls through to its own notifyUnreadStateChangedIfNeeded(hadUnreadBefore) at the bottom, it re-evaluates against the original stale hadUnreadBefore:
> 
> close(): hadUnreadBefore = true (captured) removeHistoryEntry(id) → fires callback (true→false transition) notifyUnreadStateChangedIfNeeded(true) → current=false, fires callback AGAIN
> 
> This fires m_stateCallback twice for a single dismiss. The fix would be to skip the bottom notifyUnreadStateChangedIfNeeded when the dismiss path was taken (since removeHistoryEntry already handles it), or to remove the notification from close() and rely solely on removeHistoryEntry's internal check for that branch.

Fixed.

ItsLemmy:
Great, tysm

PR #2798, opened 2026-05-25
refactor: share desktop entry launch handling

DFSko:
## Summary
- Add a shared desktop entry launch helper for Exec parsing, terminal wrapping, systemd-run mode, working directory, and activation tokens.
- Migrate launcher, dock, and taskbar desktop-entry action launches to the shared helper.
- Remove duplicated dock/taskbar launch wrappers while preserving live config behavior for taskbar launches.

## Test Plan
- [x] meson test -C build --print-errorlogs
- [x] meson compile -C build noctalia
- [x] duplicate launch scan confirms only the shared helper owns desktop-entry launch handling

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -579,6 +579,7 @@ _noctalia_sources = files(
   'src/shell/bar/widgets/workspaces_widget.cpp',
   'src/system/dependency_service.cpp',
   'src/system/desktop_entry.cpp',
+  'src/system/desktop_entry_launch.cpp',
   'src/system/brightness_service.cpp',
   'src/system/lock_keys_service.cpp',
   'src/system/distro_info.cpp',
@@ -828,6 +829,18 @@ process_test = executable('process_test',
 )
 
 test('process', process_test)
+
+desktop_entry_launch_test = executable('desktop_entry_launch_test',
+  sources: files(
+    'tests/desktop_entry_launch_test.cpp',
+    'src/system/desktop_entry_launch.cpp',
+    'src/core/process.cpp',
+    'src/core/log.cpp',
+  ),
+  include_directories: _noctalia_inc,
+)
+
+test('desktop_entry_launch', desktop_entry_launch_test)
 
 string_utils_test = executable('string_utils_test',
   sources: files(
--- a/src/app/application.cpp
+++ b/src/app/application.cpp
@@ -1110,7 +1110,7 @@ void Application::initUi() {
   );
   {
     auto launcherPanel = std::make_unique<LauncherPanel>(&m_configService, &m_asyncTextureCache);
-    launcherPanel->addProvider(std::make_unique<AppProvider>(&m_configService, &m_wayland));
+    launcherPanel->addProvider(std::make_unique<AppProvider>(&m_configService, &m_compositorPlatform));
     launcherPanel->addProvider(std::make_unique<WallpaperProvider>(&m_configService, &m_wayland));
     launcherPanel->addProvider(std::make_unique<SessionProvider>(&m_configService, &m_sessionActionRunner));
     launcherPanel->addProvider(std::make_unique<MathProvider>(&m_clipboardService));
--- a/src/compositors/compositor_platform.cpp
+++ b/src/compositors/compositor_platform.cpp
@@ -435,6 +435,12 @@ const compositors::niri::NiriRuntime& CompositorPlatform::niriRuntime() const no
 
 bool CompositorPlatform::hasXdgShell() const noexcept { return m_wayland.hasXdgShell(); }
 
+bool CompositorPlatform::hasXdgActivation() const noexcept { return m_wayland.hasXdgActivation(); }
+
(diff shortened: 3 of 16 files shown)


ItsLemmy:
Very nice, I like the idea! Looking forward to merge this.


  Overview

  Extracts the duplicated desktop entry launch logic (field-code stripping, tokenization, terminal discovery, executable path expansion, systemd-run,  activation tokens) from three separate locations (launcher/AppProvider, dock, taskbar) into a shared desktop_entry_launch module. Adds unit tests for the   prepareCommand function.

  What improved

  - Dock and taskbar action launches now get proper treatment. Previously they had a crude inline field-code stripper (just skipping the char after %) and launched with raw process::runAsync(cmd) — no terminal wrapping, no working directory, no systemd-run, no activation tokens. Now they go through the same  full pipeline as the launcher.
  - Terminal candidate list expanded. Added kgx and ptyxis to the fallback terminal list.
  - CompositorPlatform now exposes hasXdgActivation() and requestActivationToken(). The taskbar previously had no way to get activation tokens; now it goes  through CompositorPlatform rather than reaching into WaylandConnection directly.
  - Good test coverage for prepareCommand — field codes, quoting, terminal wrapping, empty commands.

  Issues

  1. const Config& m_config in TaskbarWidget is a dangling reference risk. The widget factory passes m_config (which is const Config& from  ConfigService::config()) to the taskbar constructor. If config hot-reload replaces the Config object, the reference in TaskbarWidget points to the old   one. The dock avoids this by going through m_config->config() (pointer to ConfigService). The taskbar should store const ConfigService& or const  ConfigService* instead, and read config().shell.launchAppsAsSystemdServices at launch time. This is the most significant issue.

  2. prepareCommand ignores workingDir. The parameter is accepted but explicitly cast away with (void)workingDir. The working directory is only used later   in launchEntry/launchAction when calling process::runAsync. This is fine architecturally (working dir is a process concern, not a command-preparation   concern), but the parameter on prepareCommand is misleading. Either remove it from the signature or document why it's there (future use?).

  3. Launcher still uses m_wayland->hasXdgActivation() directly (line 289) instead of going through CompositorPlatform like the dock and taskbar now do.   Since the PR added CompositorPlatform::hasXdgActivation() specifically for this purpose, the launcher should be migrated too for consistency. The   launcher holds m_wayland (a WaylandConnection*) while the other callers use CompositorPlatform& — this is a pre-existing inconsistency but this PR was   the right time to fix it.
  
  4. discoverTerminal returns options.terminalCandidates without validating they're executable. If a caller passes terminalCandidates =   {"nonexistent-terminal"}, it's used as-is, while the system discovery path validates with isExecutableOnPath. The test passes {"test-terminal"} which   won't exist on the system, but since tests only call prepareCommand (not launchEntry), the actual exec will fail silently at runtime.
  
  5. kTerminalCandidates grew from 9 to 11 entries (added kgx and ptyxis). These were already in the usesCommandSeparator check in the old code's  terminalLaunchArgs, but weren't in the candidate list — so this is a bugfix rolled into the refactor.

  Style

  - Clean namespace (desktop_entry_launch::) with clear separation of concerns.
  - PrepareOptions / LaunchOptions / PreparedCommand structs are well-designed.
  - Tests use a simple expect pattern consistent with the project's other test files.
  - tokenize improved to use range-for instead of index iteration.

  Verdict

  Good refactor that eliminates real duplication and upgrades dock/taskbar launches to feature parity with the launcher. The dangling const Config&  reference (1) needs to be fixed before merge — it's a latent crash on hot-reload.
  The workingDir parameter on prepareCommand (2) is odd but harmless.
  The launcher's direct WaylandConnection access (3) is a missed consistency opportunity.

DFSko:
Addressed the review items in follow-up commits:

1. Taskbar launch config now goes through `ConfigService&` instead of storing a `const Config&` in `TaskbarWidget` (`a351a1bfb`). The action callback reads `m_configService.config().shell.launchAppsAsSystemdServices` at launch time, matching the dock-style live config access.

2. Removed the unused `workingDir` parameter from `desktop_entry_launch::prepareCommand(...)` (`a351a1bfb`). `workingDir` now stays only at the process launch boundary in `launchEntry(...)` / `launchAction(...)`.

3. Migrated `AppProvider` to use `CompositorPlatform*` for XDG activation token access instead of reaching into `WaylandConnection` directly (`53bf49889`).

4. `PrepareOptions::terminalCandidates` are now validated before use (`b9efec5a4`). Each candidate is tokenized, and the first candidate whose executable is resolvable/executable is selected. Added a test with a missing candidate followed by a temporary executable fixture.

5. Left `kgx` / `ptyxis` in the fallback list intentionally. They were already handled by the terminal separator logic, so adding them to discovery makes the terminal handling internally consistent.

Current PR head is `b9efec5a4`; CI is green (`clang-format` and `Build (debug)`).

ItsLemmy:
Thanks

PR #2802, opened 2026-05-25
sysmon: add threshold settings and widget color highlighting

Ape:
This allows colorizing the sysmon widget values based on the value. The color will be interpolated between the regular widget color and the chosen highlight color as the value increases.

## Type of Change
- [x] New feature

## Testing
- [x] Tested on Hyprland

## Screenshots
image
image
image

## Checklist
- [x] Self-reviewed my code


ItsLemmy:
It's cool but I don't feel like having the thresholds per widget is the good call here.  This means we cannot show warning or critical in the two other places where we have system stats:  control center system tab and desktop widgets. 

I'd rather go with v4 approach where we had the poll interval and the thresholds all in one place.

I would take this part and move it to a new "System" tab and add all the threshold there, so we can reuse them globally. 
image

Let me know.

Ape:
> I'd rather go with v4 approach where we had the poll interval and the thresholds all in one place.

Yeah, agreed.

Ape:
@ItsLemmy, I reimplemented this feature with that approach.

The threshold settings take quite a bit of room in the Services tab. Perhaps we should move System to its own tab. Might be also nice to have a slider widget that can adjust two values in one slider.

ItsLemmy:
The System tab is now available.

Ape:
> The System tab is now available.

Rebased and now using the new System tab.

ItsLemmy:
Thanks

  - Dead code in sysmon_threshold_profile.h. parseStat() and the ThresholdProfile::minValue/maxValue/step fields are never used anywhere in the PR. The  settings registry hardcodes its own min/max/step per addThresholdPair(...) call instead of pulling them from the profile. Worse, they disagree: the  profile gives temps maxValue = 100 (the struct default) while the registry passes 120. 
  Either:
    - wire the registry to read thresholdProfile(stat).{minValue,maxValue,step} (preferred — single source of truth, matches the spirit of the central  profile), or
    - drop the unused fields and parseStat() to avoid a second, drifting source of these constants.

  - Threshold defaults vs. settings ranges duplicated. Each addThresholdPair repeats min/max/step already conceptually owned by the profile. This is the  same drift risk as above and is the most impactful cleanup here.

  - namespace sysmon = noctalia::sysmon; alias in config_types.h (global scope). This is a very widely-included header, so the alias leaks into every  translation unit that includes it. Prefer using the fully-qualified noctalia::sysmon:: in the member initializers, or scope the alias more narrowly, to  avoid polluting the global namespace.

  - currentThresholds() fallback {50.0, 100.0} when m_configService == nullptr doesn't match the profile-derived defaults. In practice the factory now  always passes &m_configService, so the null path is unreachable — consider asserting non-null instead of returning magic numbers, or deriving the  fallback from thresholdProfile(m_stat) so it stays consistent if the null path ever becomes live.
  i18n::tr(...activity-threshold.label...). Translated locales will render translated template text with English stat names interpolated. If full  localization matters, the stat label should itself come from a translation key.

  - valueSuffix rendering inconsistency. StepperSetting already renders its suffix inside the factory, but this PR renders SliderSetting's suffix by  manually wrapping the control in a row in settings_content.cpp. Functionally fine, but two different mechanisms for the same concept — consider pushing  slider suffix handling into factory.makeSlider for parity.

Ape:
@ItsLemmy, Rebased and updated with fixes including the things you mentioned.

ItsLemmy:
Thanks

PR #2805, opened 2026-05-26
feat(nix): add default.nix for compatibility with non-flake setups

pengolord:
## Motivation
The package is currently very cumbersome to install without flakes, unless you package it manually.
A lot of nix/nixos users don't use flakes, so I thought it'd be a nice idea to have a simpler frontend for grabbing the package and home-manager/hjem modules.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring

## Testing
Swapped out my pin for noctalia-shell to my fork, imported it, and grabbed the package out. Worked fine.

## Additional Notes
I wasn't able to replicate the `shortRev` or `lastModifiedDate` functionality from the flake, so I marked this as a draft in case someone knows how to do it. However, the package works fine, and nix's hashing system prevent conflicts, so it's fully functional

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -1,5 +1,10 @@
 let
-  inherit (builtins) fromJSON readFile;
+  inherit (builtins)
+    fromJSON
+    readFile
+    substring
+    pathExists
+    ;
 
   lock = fromJSON (readFile ./flake.lock);
   namedNode = lock.nodes.${lock.nodes.root.inputs.nixpkgs}.locked;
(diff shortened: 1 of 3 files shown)


ItsLemmy:
@linusammon

linusammon:
Two things, 

1. I looked into the `shortRev`/`lastModifiedDate` problem some more and found `builtins.fetchGit`, it exposes the same shortRev and lastModifiedDate that the flake gives us.

```nix
nix eval --impure --expr 'let gi = builtins.fetchGit { url = ./.; }; in { shortRev = gi.shortRev; lastModifiedDate = gi.lastModifiedDate; }'
```

2. In the flake we set `programs.noctalia.package = lib.mkDefault <package>`, but default.nix just returns the raw modules, so package is null. We should add that default here too so it works out of the box.

pengolord:
I think using `fetchGit` would duplicate everything, not sure if that's a good solution

linusammon:
> I think using `fetchGit` would duplicate everything, not sure if that's a good solution

I haven't tested this but maybe we could avoid that by using it as the src?

pengolord:
This works, but I think it would be preferable to remove the `shortRev` and `version` attributes. When you don't have them built-in from flakes, there's some weird edge cases where you can't really get them because the git directory is removed

pengolord:
for example, if you pull in a github tarball of a specific commit without flakes, there isn't a git directory, so the `fetchGit ./.` doesn't work

linusammon (approved):
@ItsLemmy

PR #2815, opened 2026-05-26
refactor(dock): split dock responsibilities

DFSko:
## Summary
- Split dock internals by responsibility: geometry helpers, context menu handling, instance lifecycle, and item model/scene code.
- Kept `Dock` as the public facade while reducing private implementation detail exposure from `dock.h`.
- Preserved existing dock behavior and fixed lifecycle edge cases found during review.

## Test Plan
- [x] `meson compile -C build`
- [x] `meson test -C build desktop_entry_launch app_identity wayland_toplevels_identity`
- [x] `git diff --check`

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -458,6 +458,7 @@ _noctalia_sources = files(
   'src/shell/dock/dock.cpp',
   'src/shell/dock/dock_context_menu.cpp',
   'src/shell/dock/dock_geometry.cpp',
+  'src/shell/dock/dock_instance.cpp',
   'src/shell/desktop/desktop_widget.cpp',
   'src/shell/desktop/desktop_widget_factory.cpp',
   'src/shell/desktop/desktop_widget_settings_registry.cpp',
--- a/meson.build
+++ b/meson.build
@@ -459,6 +459,7 @@ _noctalia_sources = files(
   'src/shell/dock/dock_context_menu.cpp',
   'src/shell/dock/dock_geometry.cpp',
   'src/shell/dock/dock_instance.cpp',
+  'src/shell/dock/dock_items.cpp',
   'src/shell/desktop/desktop_widget.cpp',
   'src/shell/desktop/desktop_widget_factory.cpp',
   'src/shell/desktop/desktop_widget_settings_registry.cpp',
--- a/meson.build
+++ b/meson.build
@@ -456,6 +456,7 @@ _noctalia_sources = files(
   'src/shell/bar/bar.cpp',
   'src/shell/clipboard/clipboard_paste.cpp',
   'src/shell/dock/dock.cpp',
+  'src/shell/dock/dock_geometry.cpp',
   'src/shell/desktop/desktop_widget.cpp',
   'src/shell/desktop/desktop_widget_factory.cpp',
   'src/shell/desktop/desktop_widget_settings_registry.cpp',
--- a/meson.build
+++ b/meson.build
@@ -456,6 +456,7 @@ _noctalia_sources = files(
   'src/shell/bar/bar.cpp',
   'src/shell/clipboard/clipboard_paste.cpp',
   'src/shell/dock/dock.cpp',
+  'src/shell/dock/dock_context_menu.cpp',
   'src/shell/dock/dock_geometry.cpp',
   'src/shell/desktop/desktop_widget.cpp',
   'src/shell/desktop/desktop_widget_factory.cpp',
--- a/src/shell/dock/dock.cpp
+++ b/src/shell/dock/dock.cpp
@@ -44,6 +44,11 @@ namespace {
     assert(renderContext != nullptr);
   }
 
+  void assertDockCoreInitialized(const CompositorPlatform* platform, const ConfigService* config) {
+    assert(platform != nullptr);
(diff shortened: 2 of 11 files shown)


ItsLemmy:
Excellent initiative, but very large diff to review manually.

  ### Overview

  Splits the monolithic `Dock` class into four focused modules under `src/shell/dock/`:

  | File | Responsibility |
  |------|---------------|
  | `dock_geometry` | Surface/panel geometry, anchors, input regions, auto-hide slide deltas |
  | `dock_context_menu` | Right-click popup creation, event routing, menu callbacks |
  | `dock_instance` | `DockInstance` struct, scene building, palette, blur, hide fade-out |
  | `dock_items` | `DockItemView` struct, model sync, item rebuild, visual update, click handling |

  `Dock` remains the public facade, delegating to free functions in `shell::dock::`. The header shrinks dramatically — internal types and methods are now
  hidden.

  ### Issues

  #### 1. Behavioral change in dismissed callback

  `dock_context_menu.cpp` wraps the dismissed callback in `DeferredCall::callLater()`:

  ```cpp
  menu->surface->setDismissedCallback([callbacks]() {
      DeferredCall::callLater([callbacks]() {
          if (callbacks.closeMenu) {
              callbacks.closeMenu();
          }
      });
  });
```

  The original called closeItemMenu() directly:

  menu->surface->setDismissedCallback([this]() { closeItemMenu(); });

  The PR description mentions "fixed lifecycle edge cases found during review" — this may be intentional. Worth confirming: was there a specific crash or double-free this fixes? The onActivate path already uses DeferredCall, so this makes the two paths consistent, but it changes
   the timing of popup teardown.

  2. Dependency struct duplication

  DockItemSceneDependencies is a strict superset of DockItemModelDependencies — they share 5 of 7 fields. Consider composition:

  ```cpp
  struct DockItemSceneDependencies {
      DockItemModelDependencies model;
      RenderContext& renderContext;
      IconResolver& iconResolver;
  };
```

  This avoids maintaining two identical field lists.

  3. Inconsistent null-guarding

  Several call sites gained new null checks that weren't present before, e.g.:

  - dock.cpp line adding m_config != nullptr && before the existing m_config->config().dock.autoHide check
  - prepareFrame callback now guards m_platform == nullptr || m_config == nullptr || m_renderContext == nullptr where the original only checked m_renderContext

  These are defensive, but the original code didn't need them — these pointers are set in initialize() and never cleared. Adding guards in some paths but not others creates inconsistency: some callers silently no-op on null, others still dereference unconditionally. Either add a class invariant (assert non-null after init) or guard consistently everywhere.

  4. Per-item callback overhead

  Each dock item's click handler now captures deps and callbacks by value:

  areaNode->setOnClick([itemPtr, instPtr, deps, callbacks](const InputArea::PointerData& d) { ... });

  callbacks contains 4 std::function objects. For a dock with N items, that's 4N std::function copies (each potentially heap-allocating). The original code captured this and called member functions directly. Not a correctness issue, but a regression in allocation overhead. Consider storing the callbacks once on the DockInstance or passing a pointer/reference to a single shared
  copy.

  Minor nits

  - std::vector<shell::dock::DockItemView> in dock_instance.h — already inside shell::dock namespace, can just be std::vector<DockItemView>.
  - Three separate constexpr Logger kLog("dock") definitions across the new TUs. Not a bug (anonymous namespace), just duplication.

  Positives

  - Clean namespace boundaries (shell::dock::)
  - dock.h is dramatically simpler — internal types and methods are hidden behind forward declarations
  - Geometry helpers are now independently testable pure functions
  - Correct Dock::~Dock() = default in the .cpp for forward-declared unique_ptr members
  - DockPopup::~DockPopup() correctly defaulted in .cpp where complete types are visible
  - Mechanical code motion with no intended behavioral changes (aside from the lifecycle fixes)

  Summary

  Well-structured split along natural responsibility boundaries. The main items to address are: confirming the dismissed-callback deferral is intentional (1), reducing the per-item callback copy overhead (4), and either committing to null-guarding everywhere or nowhere (3). The dependency struct duplication (2) is a nice-to-have cleanup.

DFSko:
Addressed the review feedback in follow-up commits.

On the dismissed callback: the deferral is intentional. The dismissed callback can be emitted while the popup surface is still inside its own event/callback handling path, and `closeMenu` resets `m_itemMenu`, which destroys the owning `DockPopup`/popup surface. Deferring moves that teardown to the next event-loop turn, so we do not destroy the popup from inside one of its own callbacks. This also matches the existing activate path, which already deferred menu teardown.

Other follow-ups:
- `c4e14503f` composes `DockItemSceneDependencies` from `DockItemModelDependencies` to avoid maintaining duplicate dependency field lists.
- `1521653ca` preserves the existing model/rebuild guard semantics while keeping the newly split lifecycle paths from silently no-oping on violated invariants.
- `d747fe516` replaces per-item copies of the callback/dependency bundle with one shared click context per rebuild.

Local verification passed:
- `clang-format --dry-run -Werror`
- `meson compile -C build`
- `meson test -C build desktop_entry_launch app_identity wayland_toplevels_identity`

ItsLemmy:
Thanks a lot!

PR #2821, opened 2026-05-27
refactor(dock): introduce snapshot model boundary

DFSko:
## Summary
- Introduce a DockSnapshot model boundary for dock item identity, active/running state, output filtering, and instance counts.
- Move app/window discovery out of dock_items so rendering consumes prepared snapshot data.
- Make DockItemView UI-only and use copied DockItemAction data for launch/menu callbacks.

## Problem solved
The dock rendering layer previously mixed UI construction with domain-model discovery: it resolved desktop entries, queried running apps/windows, tracked active state, and owned click/menu identity directly in DockItemView. That made the dock hard to reason about, tightly coupled rendering to compositor/app services, and increased the risk of stale state or lifetime issues when item views were rebuilt.

This refactor separates those responsibilities: dock_model builds the snapshot, Dock orchestrates sync/actions/menus, and dock_items renders from the prepared data.

## Verification
- git diff --check cc1b0a8fa071ab4f6bba30ad5db3b618810630d2..HEAD
- rg checks confirmed dock_items no longer performs app/window discovery or stores domain state in DockItemView
- just build: not available in this environment (just: command not found)
- meson test -C build-debug --print-errorlogs (12/12 passed)

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -460,6 +460,7 @@ _noctalia_sources = files(
   'src/shell/dock/dock_geometry.cpp',
   'src/shell/dock/dock_instance.cpp',
   'src/shell/dock/dock_items.cpp',
+  'src/shell/dock/dock_model.cpp',
   'src/shell/desktop/desktop_widget.cpp',
   'src/shell/desktop/desktop_widget_factory.cpp',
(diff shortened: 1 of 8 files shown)


ItsLemmy:
Thank you!

PR #2822, opened 2026-05-27
mpris: ignore placeholder players without usable metadata

Ape:
Currently the media player widget can show useless media players that don't have any other information besides the player name. One example would be if you had media on Chromium, but navigated away from it. We should ignore these empty, invalid players as if they don't exist. The player will be shown again when it provides media metadata.

## Type of Change
- [x] Bug fix

## Testing
- [x] Tested on Hyprland

## Checklist
- [x] Self-reviewed my code

### DIFF
--- a/src/dbus/mpris/mpris_service.cpp
+++ b/src/dbus/mpris/mpris_service.cpp
@@ -214,6 +214,18 @@ namespace {
     return !info.title.empty() || !info.artists.empty() || !info.album.empty();
   }
 
+  std::vector<std::string> normalizeArtists(std::vector<std::string> artists) {
+    std::erase_if(artists, [](const std::string& artist) { return StringUtils::trim(artist).empty(); });
+    return artists;
+  }
+
+  std::string normalizeTrackId(std::string trackId) {
+    if (trackId.ends_with("/NoTrack")) {
+      return {};
+    }
+    return trackId;
+  }
+
   std::string canonicalTrackSourceUrl(std::string_view rawUrl) {
     if (rawUrl.empty()) {
       return {};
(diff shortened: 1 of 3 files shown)


Ly-sec:
Nice fix overall - keeping the proxy subscribed but dropping empty entries from the cache is the right model for Chromium-style placeholders, and stripping `/NoTrack` makes sense.

One thing I'd fix before merge: `addOrRefreshPlayer` already tolerates one of the two `GetAll` calls failing. If the player interface fails, `readPlayerInfoFromProperties` can come back with empty metadata even though something is still playing. With this PR we'd call `removePlayerCacheEntry` and drop a player that was fine a moment ago. Previously `applyPlayerSnapshot` would merge and keep the last good snapshot on weak updates.

I'd only evict when the player `GetAll` actually succeeded and metadata is still empty after normalization, something like:

```cpp
if (!playerFailed && !hasAnyPlayerMetadata(info)) {
  removePlayerCacheEntry(busName);
  return;
}
```

Worth sanity-checking with a brief D-Bus hiccup while Spotify/Firefox is playing, the bar shouldn't flash empty.

Smaller nits:

- `hasAnyPlayerMetadata` is looser than the existing `hasStrongNowPlayingMetadata` helper (which already calls out track IDs / URLs as loading placeholders). A player with only `artUrl` (app icon) still passes the gate and can show up as identity-only in the widget - might be worth using the strong check for the gate, or calling out if art-only is intentional.

- `removePlayerCacheEntry` is basically a copy of `removePlayer` but doesn't clear `m_lastPropertiesUpdate` / the properties failure backoff maps. Would be nice to share teardown between the two paths.

- Tests for `normalizeTrackId`, the metadata gate, and "don't evict when `playerFailed`" would make this easier to regress later.

The `audio_tab` switch to `listPlayers()` is a good catch - I'd keep that.

Ape:
@Ly-sec, I pushed some changes after rethinking this with your comments in mind.

Ly-sec:
Looks good to me now @Ape  the `!playerFailed` guard and switching to `hasStrongNowPlayingMetadata` address the main concern, and `clearPlayerState` cleans up the duplication nicely. I’d only sanity-check skip-next on Spotify/Firefox in case partial metadata during track changes causes a flicker, but I’m fine with merging this.
Thank you very much!

PR #2839, opened 2026-05-29
feat(weather): add independent temperature and wind speed units with kelvin, ms, kn and bft

sic-quod-tsavorite:
# Pull Request

## Motivation
Where I'm from wind speed is measured in meters per second, so instead of unit selection being just metric/imperial, I split it into temperature and wind speed. Then added the wind speed option for meters per second, and in case others would want it, the temperature option for kelvin and the wind speed options for knots and beaufort scale.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring

## Testing

- [x] Tested on niri
- [ ] Tested on Hyprland
- [ ] Tested on sway

## Screenshots / Videos
Screenshot from 2026-05-29 20-48-12
Screenshot from 2026-05-29 20-50-40

## Checklist
- [x] Self-reviewed my code

### DIFF
--- a/assets/translations/en.json
+++ b/assets/translations/en.json
@@ -1080,9 +1080,17 @@
         "overlay": "Overlay"
       },
       "weather": {
-        "unit": {
-          "metric": "Metric",
-          "imperial": "Imperial"
+        "temperature_unit": {
+          "celsius": "Celsius",
+          "fahrenheit": "Fahrenheit",
+          "kelvin": "Kelvin"
+        },
+        "wind_speed_unit": {
+          "kmh": "km/h",
+          "mph": "mph",
+          "ms": "m/s",
+          "kn": "knots",
+          "bft": "Beaufort"
         }
       },
       "wallpaper": {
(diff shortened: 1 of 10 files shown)


ItsLemmy:
Hey, thanks for the PR but we won't be merging that.

We wanna keep it simple and this PR adds unnecessary bloat and weird assumptions like checking the windspeed unit to determine if the elevation should be shown in meters. 

Noctalia is not a scientific weather station or even a weather app, if people need to know the wind speed in knots they should not use Noctalia for that.

sic-quod-tsavorite:
@ItsLemmy Fair enough, I'd still personally like to see m/s since that's what we use at least in scandinavia. 

I mainly added the other options while I was at it since they were options in kde. 
I'm happy to make a simpler PR that only contains m/s otherwise just consider this a feature request. 

I use v4 of noctalia right now and on the lock screen it saying km/h, is to me more of a bother than it is nice. 
Anyway if you read this far, no matter your response, thanks for taking the time, noctalia is awesome so thanks a bunch to the entire team for making it. :)

Ape:
I'd also like to see wind speed in m/s instead of km/h.

PR #2840, opened 2026-05-30
refactor(dock): share pinned app matching

DFSko:
## Summary
- Extract shared dock pinned app matching and resolution into `shell::dock::pinned_apps`.
- Replace duplicate launcher pin/unpin helpers with direct shared dock pinned app calls.
- Add focused pinned apps coverage and Meson test target.

## Test Plan
- [x] `meson test -C build-debug dock_pinned_apps app_identity --print-errorlogs`
- [x] `meson compile -C build-debug noctalia`
- [x] `git diff --check`

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -466,6 +466,7 @@ _noctalia_sources = files(
   'src/shell/dock/dock_instance.cpp',
   'src/shell/dock/dock_items.cpp',
   'src/shell/dock/dock_model.cpp',
+  'src/shell/dock/pinned_apps.cpp',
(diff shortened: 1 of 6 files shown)


ItsLemmy:
Logic's good. Just needs a rebase onto v5, this predates your snapshot-model refactor, so the loop it removes is now in dock_model.cpp::refreshPinnedAppsIfNeeded, not dock_items.cpp. Move resolveEntries() there and it's good to go.

ItsLemmy:
Thanks!

PR #2847, opened 2026-05-30
bar: animate transitions when toggling auto-hide via ipc

Ape:
Continuation on #2774. The current version lacks show / hide animation, but this commit adds it-

## Type of Change
- [x] New feature

## Testing
- [x] Tested on Hyprland

## Checklist
- [x] Self-reviewed my code

### DIFF
--- a/src/shell/bar/bar.cpp
+++ b/src/shell/bar/bar.cpp
@@ -1878,6 +1878,9 @@ void Bar::syncBarAutoHideInputRegion(BarInstance& instance) const {
 }
 
 void Bar::revealAutoHideBar(BarInstance& instance) {
+  if (instance.autoHideDisablePending) {
+    return;
+  }
   if (!instance.barConfig.autoHide || instance.surface == nullptr || instance.slideRoot == nullptr) {
     return;
   }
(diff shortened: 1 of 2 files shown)


ItsLemmy:
Issues / risks

### 1. Medium — rapid-toggle race on hideOpacity (owner mismatch).
The new disable animation registers with owner = instance.slideRoot, but the reused  revealAutoHideBar/startHideFadeOut helpers register with owner = nullptr (confirmed: AnimationManager::animate defaults owner to nullptr, and these helpers pass no owner). The guard at the top of applyTransientAutoHide is cancelForOwner(instance.slideRoot), which cannot cancel a nullptr-owned animation, and animate() does not auto-dedup by setter or owner (it just  push_backs).

So this sequence stacks two live animations writing hideOpacity:
- bar-auto-hide-set on (no pointer) → startHideFadeOut, animating 1.0 → 0.0, owner=nullptr.
- a frame later, bar-auto-hide-set off → enters the animate branch (autoHide still true, hideOpacity < 0.999) → reveal 0.x → 1.0, owner=slideRoot. The top-level cancelForOwner(slideRoot) does not kill the still-running fade-out.
  
Result: two animations fight over hideOpacity for one animNormal → brief jitter. It largely self-heals (the slide transform reads 0,0 once autoHide flips  false), but it's visible. IPC handlers run sequentially on the main loop, but the animations span frames, so a double-fired keybind/toggle script reaches it.
  
Cleanest fix is to tag revealAutoHideBar/startHideFadeOut's animate() calls with owner = slideRoot too, so the existing   cancelForOwner(slideRoot) covers every auto-hide opacity animation (this also tightens the pre-existing hover enter/leave flow, which has the same latent stacking).

Ape:
@ItsLemmy, updated with fixes.

ItsLemmy:
Thanks

PR #2851, opened 2026-05-30
notification_toast: clean up both visually and in code

Ape:
The notification toast widget needed some polish.

## Testing
- [x] Tested on Hyprland

## Video
https://github.com/user-attachments/assets/42fc4c53-c5a6-4c71-bb5c-1d6cf207015c

## Checklist
- [x] Self-reviewed my code

### DIFF
--- a/assets/translations/en.json
+++ b/assets/translations/en.json
@@ -2366,6 +2366,10 @@
           "label": "Notification Daemon",
           "description": "Enable the built-in notification daemon"
         },
+        "show-app-name": {
+          "label": "Show App Name",
+          "description": "Show the sending application's name in notification toasts"
+        },
         "position": {
           "label": "Screen Position",
           "description": "Screen position for notification toasts"
--- a/example.toml
+++ b/example.toml
@@ -111,6 +111,7 @@ tint_intensity         = 0.3        # 0.0 = no tint, 1.0 = opaque
 
 [notification]
 enable_daemon      = true
+show_app_name      = true
 layer              = "top"          # top | overlay
 scale              = 1.0            # notification size multiplier applied on top of shell.ui_scale
 background_opacity = 0.97
--- a/src/config/config_export.cpp
+++ b/src/config/config_export.cpp
@@ -600,6 +600,7 @@ namespace config_export {
     toml::table notificationTable(const NotificationConfig& notification) {
       toml::table table;
       table.insert_or_assign("enable_daemon", notification.enableDaemon);
+      table.insert_or_assign("show_app_name", notification.showAppName);
       table.insert_or_assign("position", notification.position);
       table.insert_or_assign("layer", notification.layer);
       table.insert_or_assign("scale", static_cast<double>(notification.scale));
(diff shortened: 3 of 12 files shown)


ItsLemmy:
Issues / risks

### 1. Medium — card height cap (maxToastCardHeight) no longer bounds the rendered card.
refreshEntryGeometry clamps the placement height:
entry.height = std::min(maxToastCardHeight(scale), measureToastCardHeight(...));

but buildCard sizes the actual card to the unclamped content height:  foreground->layout(*m_renderContext);
const float cardHeight = foreground->height();   // not clamped to maxToastCardHeight
viewport->setSize(cardW, cardHeight);

Since measureToastCardHeight and buildCard build the same tree, for any notification taller than maxToastCardHeight (long body at 3 lines + actions row + app-name footer + icon, especially at high scale) entry.height is clamped but the rendered card is taller — so the card overruns its reserved stacking slot and can overlap the neighbour. The old planToastLayout/clampBodyLabelHeight path guaranteed card ≤ maxCard. Either drop the std::min clamp (accept tall cards) or clamp+clip the card in buildCard to match placement. This is the one I'd most want resolved.
  
###  2. Low–Medium — inline-reply mode changes card height after build, but the slot doesn't follow.

enterInlineReplyMode now toggles participatesInLayout on the actions row (off) and reply row (on) and calls requestLayout(). The viewport/cardRoot are  fixed at build time to foreground->height() with clipChildren = true, and entry.height (placement) was measured without the reply row (measureToastCardHeight never adds it). If the reply row is taller than the actions row it replaces, the extra height is clipped by the viewport and the  placement slot is stale. In practice the reply row (input + send, controlHeightSm) is probably ~ the actions row height so it likely fits — but it's  worth confirming in the demo, since the video may not have exercised a reply card whose reply row exceeds its action row.

###  3. Low — two sources of truth for the card tree.

measureToastCardHeight and buildCard independently construct what must be the same Flex tree; the header comment even warns "buildCard must match these for placement vs paint." They currently agree (same maxLines, same children, body added-vs-participatesInLayout=false is layout-equivalent), but any  future edit to one must be mirrored. Consider extracting a single buildToastContent(...) that both measurement and the real card call, to make drift impossible.

Ape:
> 1. Medium — card height cap (maxToastCardHeight) no longer bounds the rendered card.

I'm unable to produce any issues like this. I tried extremely long messages with action buttons and extreme ui / notification scales, but it all works. I think the code is good around this.

Ape:
> 2. Low–Medium — inline-reply mode changes card height after build, but the slot doesn't follow.

Same story here. I'm unable to produce any issues in practice. I don't see any reason to change the code.

Ape:
> 3. Low — two sources of truth for the card tree.

I agree that this is an issue about code maintainability. We might want to do further refactoring, but I think this PR greatly improves this exact situation already.

Ape:
@ItsLemmy, any other comments? I think this PR is ready to merge.

PR #2862, opened 2026-06-01
Hide trash button in notification panel instead of disabling it

notiant:
Since the clear history button is hidden in the clipboard panel if there is no history, the same behavior should also apply in the notification panel.
Also some i18n corrections & removing an unused property in the bar widget editor.


ItsLemmy:
The famous notiant's multi purpose PR

PR #2867, opened 2026-06-01
fix(ThemeIcons): prevent short-id fuzzy/clean matches from outscoring real entries

### LINKED ISSUE #2866: ThemeIcons.findAppEntry misresolves "opencode" to code.desktop (VS Code) via checkCleanMatch/checkFuzzySearch
## Noctalia version
v4.7.7

## Reproduction
1. Install `opencode-desktop` (sets `StartupWMClass=OpenCode`; ships `opencode-desktop.desktop` with `Icon=opencode-desktop`).
2. Launch the app — its Wayland `app_id` resolves to `OpenCode` (Electron `app.setName("OpenCode")`).
3. Look at the dock / taskbar / active-window indicator.

**Observed:** VS Code's blue icon (`code.desktop`).
**Expected:** the dark-grey "0" icon shipped at `/usr/share/icons/hicolor/*/apps/opencode-desktop.png`.

## Root cause

In `Commons/ThemeIcons.qml`, `findAppEntry(appId)` falls through the 6-stage lookup chain (heuristic → substitutions → regex → `checkSimpleTransforms` → `checkFuzzySearch` → `checkCleanMatch`).

For `appId = "OpenCode"`:

1. **`checkHeuristic`** — `DesktopEntries.heuristicLookup("OpenCode")` should hit `opencode-desktop.desktop` via `StartupWMClass=OpenCode`, but in practice does not return it for this case. Falls through.
2. **`checkSubstitutions`** — no entry for `OpenCode` / `opencode`. Falls through.
3. **`checkRegex`** — none of the patterns (`steam_app_*`, `Minecraft.*`, `.*polkit.*`, `gcr.prompter`) match. Falls through.
4. **`checkSimpleTransforms`** (line 181) — queries `DesktopEntries.byId(...)` for `OpenCode`, `opencode`, etc. No `.desktop` has `id="opencode"`; the actual id is `opencode-desktop`. Falls through.
5. **`checkFuzzySearch`** (line 200) — fuzzy-matches against `preppedIds`. The 4-char id `"code"` is a tight substring of `"opencode"` and scores higher than `"opencode-desktop"` (a 16-char partial match). Returns `code.desktop`.
6. **`checkCleanMatch`** (line 233) — same hazard, more deterministically: `entryList` is sorted alphabetically by `name` (`"Code" < "Codex" < "OpenCode" < "Visual Studio Code"`), and the first iteration hits `code.desktop` whose `cleanId` is `"code"`. The condition `cleanStr.includes(cleanId)` (`"opencode".includes("code") === true`) returns `code.desktop`.

The same bug will surface for any app whose `appId` / `id` is a superset of another app's id (`logseq` vs `log`, `kile` vs `k`, `codeberg` vs `code`, etc.).

## Suggested fix

Three changes in `Commons/ThemeIcons.qml`:

\`\`\`qml
// 1. Raise the score threshold so substring matches don't dominate real matches.
property real scoreThreshold: 0.5   // was 0.2

// 2. In checkFuzzySearch, score per-entity by a density function so a 4-char
//    substring can't outscore a longer near-prefix match.
function checkFuzzySearch(str) {
  ...
  const density = (match.length) / (str.length);
  if (density < 0.6) continue;   // require >=60% of query chars matched
  ...
}

// 3. In checkCleanMatch, drop the \`cleanStr.includes(cleanId)\` direction and
//    only accept entries whose id is at least as long as the query minus 2 chars.
function checkCleanMatch(str) {
  ...
  // drop: if (cleanId.includes(cleanStr) || cleanStr.includes(cleanId))
  if (cleanId.includes(cleanStr) && cleanId.length >= cleanStr.length - 2)
    return entry;
  ...
}
\`\`\`

Plus a one-line substitution entry as an immediate unblocker:

\`\`\`qml
property var substitutions: ({
  ...
  "opencode": "opencode-desktop",
  "@opencode-aidesktop": "opencode-desktop",
  ...
})
\`\`\`

## Workaround (no patch required)

- Drop a `~/.local/share/applications/opencode-desktop.desktop` with `Icon=/usr/share/icons/hicolor/128x128/apps/opencode-desktop.png` (absolute path bypasses the lookup), or
- Add the substitution above to `ThemeIcons.qml` (gets overwritten on update).

TitoTFP:
Fixes #2866

## Problem

`ThemeIcons.findAppEntry` misresolves apps whose `appId` is a superset of another app's `id`. Concrete case: `opencode-desktop` (wayland `app_id = "OpenCode"`, `id = "opencode-desktop"`) was being matched to `code.desktop` (VS Code) in the dock/taskbar/active-window indicator.

Root cause is a length-normalization issue in `checkFuzzySearch` and a broken match direction in `checkCleanMatch`:

1. `checkFuzzySearch` (line 200) fuzzy-matches against `.desktop` ids. FuzzySort scores matches normalized by target length, so a tight 4-char substring like `"code"` in `"opencode"` outscores the 16-char partial match of `"opencode-desktop"`.

2. `checkCleanMatch` (line 233) uses `cleanId.includes(cleanStr) || cleanStr.includes(cleanId)`. The reverse direction (`"opencode".includes("code")`) returns the first alphabetically-sorted entry whose cleaned id is a substring of the query — `code` (VS Code) wins before `opencode-desktop` is even checked.

The same bug class affects any app whose name/id is a superset of another (e.g. `logseq` vs `log`, `kile` vs `k`, `codeberg` vs `code`).

## Changes

**`Commons/ThemeIcons.qml`:**

- `checkFuzzySearch`: add a length-coherence filter — `entry.id.length >= max(3, str.length - 2)`. A 4-char id cannot match an 8-char query.
- `checkCleanMatch`: drop the broken `cleanStr.includes(cleanId)` direction. Only accept entries whose cleaned id is a superset of the cleaned query.
- `substitutions`: add explicit entries for `opencode` and the Electron binary-name variant `@opencode-aidesktop`, mapping both to `opencode-desktop`. This is an immediate unblocker that bypasses the lookup entirely.

## Test plan

Manual repro from the linked issue:
1. Install `opencode-desktop` and launch it.
2. Open the dock / taskbar.
3. **Before:** VS Code's blue icon shows for the opencode window.
4. **After:** the dark-grey "0" icon from `/usr/share/icons/hicolor/*/apps/opencode-desktop.png` shows.

Regression checks (apps that should still resolve correctly):
- VS Code (`code.desktop`, `app_id = "code"`): still resolves to VS Code (`code.desktop.id = "code"`, length 4 ≥ `max(3, 4-2)=4`).
- Zed (`code-oss` etc., if installed): resolves as before via the `cleanId.includes(cleanStr)` direction.
- `gnome-tweaks` → `org.gnome.tweaks` substitution: still works (handled before the fuzzy stages).
- `pavucontrol-qt` → `pavucontrol` substitution: still works.

### DIFF
--- a/Commons/ThemeIcons.qml
+++ b/Commons/ThemeIcons.qml
@@ -14,6 +14,8 @@ Singleton {
                                  "code-url-handler": "visual-studio-code",
                                  "Code": "visual-studio-code",
                                  "gnome-tweaks": "org.gnome.tweaks",
+                                 "opencode": "opencode-desktop",
+                                 "@opencode-aidesktop": "opencode-desktop",
                                  "pavucontrol-qt": "pavucontrol",
                                  "wps": "wps-office2019-kprometheus",
                                  "wpsoffice": "wps-office2019-kprometheus",
@@ -201,13 +203,21 @@ Singleton {
     if (typeof FuzzySort === 'undefined')
       return null;
 
+    // Require the matched entry id to be length-coherent with the query.
+    // Prevents a short id (e.g. "code", 4 chars) from outscoring a longer
+    // partial match (e.g. "opencode-desktop", 16 chars) when searching for
+    // "opencode" — FuzzySort normalizes score by target length, so tight
+    // short substrings otherwise dominate.
+    const minIdLen = Math.max(3, str.length - 2);
+    const lengthOk = (entry) => entry && entry.id && entry.id.length >= minIdLen;
+
     // Check filenames (IDs) first
     if (preppedIds.length > 0) {
-      let results = fuzzyQuery(str, preppedIds);
+      let results = fuzzyQuery(str, preppedIds).filter(lengthOk);
       if (results.length === 0) {
         const underscored = str.replace(/-/g, '_').toLowerCase();
         if (underscored !== str)
-          results = fuzzyQuery(underscored, preppedIds);
+          results = fuzzyQuery(underscored, preppedIds).filter(lengthOk);
       }
       if (results.length > 0)
         return results[0];
@@ -215,14 +225,14 @@ Singleton {
 
     // Then icons
     if (preppedIcons.length > 0) {
-      const results = fuzzyQuery(str, preppedIcons);
+      const results = fuzzyQuery(str, preppedIcons).filter(lengthOk);
       if (results.length > 0)
         return results[0];
     }
 
     // Then names
     if (preppedNames.length > 0) {
-      const results = fuzzyQuery(str, preppedNames);
+      const results = fuzzyQuery(str, preppedNames).filter(lengthOk);
       if (results.length > 0)
         return results[0];
     }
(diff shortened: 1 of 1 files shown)


ItsLemmy:

We are currently pouring all of our focus into v5. Because v4 is now frozen, we won’t be merging further changes into that codebase.

While we are closing this PR, we really value your effort and would love to see your work in the v5 ecosystem. If you’d like to port this over to the new version, we’d be happy to review it there!

Cheers!

PR #2906, opened 2026-06-06
fix(script): Shared scope getting lost on mult-monitors

Mathew-D:
# Pull Request

## Motivation
Fix the shared scope from disappearing on mult-monitors when screens are turned off.

## Type of Change
- [X] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring

## Testing

- [X] Tested on niri
- [ ] Tested on Hyprland
- [ ] Tested on sway

## Checklist
- [X ] Code follows project style guidelines
- [X] Self-reviewed my code


ItsLemmy:
subscribers[id] = callback; should be std::move(callback) (script_runtime.cpp:133 area). callback is unused after the insert — the replay path  re-fetches it->second from the map,  so this is a needless copy of a std::function (a small regression vs. the original std::move).

ItsLemmy:
Thank you good sir

PR #2915, opened 2026-06-06
audio: add easyeffects profiles support

Ape:
This commit adds support for switching between easyeffects profiles (output and input audio) if available.

## Type of Change
- [x] New feature

## Testing
- [x] Tested on Hyprland

## Screenshots
image
image

## Checklist
- [x] Self-reviewed my code

### DIFF
--- a/assets/translations/en.json
+++ b/assets/translations/en.json
@@ -113,6 +113,8 @@
   "control-center": {
     "audio": {
       "application-volumes": "Application Volumes",
+      "choose-effects-profile": "Choose a profile",
+      "effects-profile": "Effects:",
       "input-volume": "Input Volume",
       "no-application-audio": "No application audio",
       "no-application-audio-body": "No application playback streams are currently available.",
@@ -596,6 +598,10 @@
     "dnd": {
       "off": "DND Off",
       "on": "DND On"
+    },
+    "effects": {
+      "input": "Input Effects",
+      "output": "Output Effects"
     },
     "lock-keys": {
       "caps-off": "Caps Lock Off",
--- a/meson.build
+++ b/meson.build
@@ -670,6 +670,7 @@ _noctalia_sources = files(
   'src/system/hardware_info.cpp',
   'src/system/icon_resolver.cpp',
   'src/system/internal_app_metadata.cpp',
+  'src/system/easyeffects_service.cpp',
   'src/system/day_night_schedule.cpp',
   'src/system/gamma_service.cpp',
   'src/system/location_service.cpp',
--- a/src/app/application.cpp
+++ b/src/app/application.cpp
@@ -106,6 +106,15 @@ namespace {
     };
   }
 
+  OsdContent effectsProfileOsdContent(AudioEffectsProfileKind kind, std::string_view profile) {
+    const char* labelKey = kind == AudioEffectsProfileKind::Input ? "osd.effects.input" : "osd.effects.output";
+    return OsdContent{
+        .icon = "adjustments",
+        .value = i18n::tr(labelKey) + ": " + std::string(profile),
+        .showProgress = false,
+    };
+  }
+
   OsdContent caffeineOsdContent(bool enabled) {
     return OsdContent{
         .kind = OsdKind::Caffeine,
@@ -991,6 +1000,9 @@ void Application::initServices() {
 
   try {
     m_pipewireService = std::make_unique<PipeWireService>();
+    m_easyEffectsService = std::make_unique<EasyEffectsService>();
+    m_easyEffectsService->refreshProfiles();
+    m_easyEffectsService->refreshActiveEffectsProfiles();
     m_pipewireSpectrum = std::make_unique<PipeWireSpectrum>(*m_pipewireService);
     m_soundPlayer = std::make_unique<SoundPlayer>(m_pipewireService->loop());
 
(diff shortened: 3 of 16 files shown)


LucasOe:
While on the topic, personally I think it would be also nice if the Easy Effects Sink could be hidden from the selectable output devices.

Ly-sec:
Nice feature, and the EasyEffects side discovery looks well thought out (XDG paths, Flatpak, `easyeffectsrc`, the runtime socket). I'd use this. Just wouldn't merge it as-is, mostly where the code lives and how often it runs.

## What works

The Control Center rows make sense: show output/input presets when EasyEffects is actually there, hide them otherwise. IPC + OSD for `effects-profile-set` fits how we already expose audio stuff. Showing the active profile in the volume widget tooltip is a good bonus.

## What I'd change

**Pull it out of `PipeWireService`.** EasyEffects isn't PipeWire, this is ~500 lines of socket I/O, INI parsing, and directory walks bolted onto a file that's already huge. A small dedicated service (whatever we want to call it) wired from `Application`, with `AudioTab` and IPC calling into it, would be much easier to live with.

**Stop scanning the filesystem every frame.** `syncEffectsProfileControls()` calls `effectsProfiles()` on every `doUpdate`, which stats the socket, walks config/data dirs, and may parse `easyeffectsrc`. No cache. That's a lot of I/O just because Control Center is open.

Ironically the *active* profile goes the other way - it's cached and only refreshed when the tab is created or you switch from Noctalia. So the list is hammered constantly but the selection can be wrong if you changed presets in EasyEffects itself.

**Socket I/O on the UI thread is probably fine for a click** (`loadEffectsProfile` has that ~250ms poll), but discovery shouldn't be on a hot path.

**Smaller nits:**

- `"EasyEffects"` is hardcoded in the volume tooltip, needs i18n
- the `dismissTransientUi()` flip is unrelated, would drop it from this PR

LucasOe's point about hiding the EasyEffects virtual sink from the device picker is fair, doesn't need to block this, but it's a natural follow-up.

## Before merge

1. Extract out of `pipewire_service.cpp`
2. Cache profile discovery, refresh on tab open, after a successful load, maybe a lazy timer. Not every frame.
3. Actually refresh the active profile when it might've changed externally
4. i18n + IPC docs
5. Fix merge conflicts

Feature: yes. As written: needs another pass.

Ape:
@ItsLemmy Can you reopen this so that I can change the base branch to `main`?

ItsLemmy:
I had to recreate a v5 branch, so please do it quick if you can

Ape:
> I had to recreate a v5 branch, so please do it quick if you can

Base branch changed! However, I won't have time to finish this PR at least for a couple of days now.

Ly-sec:
> > I had to recreate a v5 branch, so please do it quick if you can
> 
> Base branch changed! However, I won't have time to finish this PR at least for a couple of days now.

I'll mark is as a draft for now then :).

Ape:
@Ly-sec, new version ready for review.

Ly-sec:
Sorry for the late reply @Ape , here are the last few things :)

## 1. `activeEffectsProfile()` must not discover on read

`activeEffectsProfile()` falls back to `discoverActiveEffectsProfile()` when the cache is empty, which walks config roots and parses `easyeffectsrc` synchronously.

It is called from:

- `AudioTab::syncEffectsProfileControls()` (every audio-tab update until the list key stabilizes)
- `VolumeWidget::syncState()` (every bar widget update)

If `refreshActiveEffectsProfiles()` failed at startup (e.g. EasyEffects not running yet), this can repeat filesystem work whenever the bar or control center refreshes.

**Fix:** getters return only cached data (empty string if unknown). Discovery belongs in explicit `refresh*` calls, not in const accessors used from UI hot paths.

---

## 2. Active profile goes stale when changed outside Noctalia

The active preset is refreshed on app init, audio tab `create()`, and after a successful `loadEffectsProfile()` / IPC set. If the user switches presets inside EasyEffects, the control-center select and volume tooltip stay wrong until the tab is torn down and recreated.

**Fix:** refresh active profiles when the audio tab becomes active, and/or watch `easyeffectsrc` or poll the EasyEffects socket while the audio tab is visible.

---

## 3. Rebase and verify against current `main`

The PR branch is behind `main` (conflicts likely in `application.cpp`, `bar.*`, `audio_tab.*`). Rebase and re-test before merge:

- native EasyEffects install
- Flatpak EasyEffects (`com.github.wwmm.easyeffects`)
- EasyEffects not running at startup
- EasyEffects started after Noctalia is already running

Ape:
Rebased on the current main fixing minor merge conflicts.

> 2. Active profile goes stale when changed outside Noctalia

I don't see any easy way to get events from the profile changing. I don't think it's a good idea to poll easyeffects for this. The state is already updated when the audio tab is opened. I think this is good as is.

Ape:
> 1. activeEffectsProfile() must not discover on read

Valid. Fixed now.

Ly-sec:
Looks good to me now @Ape :) thank you very much. I'll merge it!

LucasOe:
The active profile should also be reloaded in case the device is changed, in case Autoload is enabled in EasyEffects. Right now with Autoload enabled, the active profile isn't updated until the panel is recreated. Maybe this requires a small delay to avoid a race condition with EasyEffects profile switching.

PR #2925, opened 2026-06-06
feat(keyboard-layout): custom labels support (v5 port)

fennstef:
## Summary

Port of the keyboard layout custom labels feature from v4 to v5.

- **`DisplayMode::Custom`**: New display mode that shows a user-defined label for each layout. Falls back to the layout name if no mapping is configured.
- **`custom_labels` config key**: TOML sub-table under `[widget.keyboard_layout]` mapping layout names to display labels:
  ```toml
  [widget.keyboard_layout]
  display = "custom"
  [widget.keyboard_layout.custom_labels]
  de = "DE"
  us = "EN"
  ```
- **Settings GUI**: All keyboard layout settings are now exposed in the widget inspector (cycle_command, hide_when_single_layout, show_icon, show_label, display, custom_labels). The custom_labels editor pre-populates rows for all currently active compositor layouts.
- **OSD pill**: The layout-change notification pill also uses custom labels.

## Implementation notes

`custom_labels` is a string→string map. Because toml++ statically introspects the `WidgetSettingValue` variant and rejects non-native TOML types, the map cannot be added to that variant. Instead, a separate `tables` field (`unordered_map<string, unordered_map<string,string>>`) was added to `WidgetConfig` to hold TOML sub-tables. Override persistence works via `widgetConfigEqual` which now compares `tables` as well.

## Test plan

- [ ] Set `display = "custom"` in widget inspector, verify segmented control shows short/full/custom
- [ ] Active layout names appear pre-populated in the custom labels editor
- [ ] Enter a label, press Enter → verify saved in `~/.local/state/noctalia/settings.toml`
- [ ] Close and reopen settings → labels still present
- [ ] Switch keyboard layout → OSD pill shows custom label
- [ ] With no custom label set for a layout → OSD falls back to display mode formatting


ItsLemmy:
I cherry-picked the `scripted_widget_bindings.cpp` change, thanks for that.

For the core PR, I’m not going to merge this as-is. The TOML shape is good, but the implementation spreads too far through config/settings for a small widget feature and leaves table-backed settings only partially wired up.

  The main blockers are:
  - `custom_labels` is a TOML table, but the schema validates it as a string.
  - Override state only checks `WidgetConfig::settings`, while `custom_labels` lives in `WidgetConfig::tables`.
  - `display = "custom"` seems unnecessary; custom labels should just override `short`/`full` labels when present.

So thanks again,  I’m going to close this and rework the feature with the same clean TOML, but a narrower implementation.

PR #2928, opened 2026-06-07
feat: add support for ics (subscription) calendars

danielepintore:
## Motivation
Add support for ics (subscription calendars). This kind of calendars are usually shared via a single url (no auth) and use the same ical format.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring

## Testing
Tried to add a calendar and checked if the events are present in the ui

- [ ] Tested on niri
- [x] Tested on Hyprland
- [ ] Tested on sway

## Checklist
- [x] Self-reviewed my code

## Additional Notes
Docs are needed, I will write them if everything is ok with this pr

### DIFF
--- a/assets/translations/en.json
+++ b/assets/translations/en.json
@@ -619,6 +619,8 @@
       "delete-error": "Could not delete calendar account.",
       "edit-title": "Edit Calendar Account",
       "id-label": "Account ID",
+      "ics-url-label": "ICS URL",
+      "ics-url-placeholder": "https://example.com/calendar.ics",
       "invalid": "Check the highlighted calendar account fields.",
       "name-label": "Display Name",
       "name-placeholder": "Personal Calendar",
@@ -629,7 +631,8 @@
       "provider": {
         "custom": "CalDAV",
         "google": "Google",
-        "icloud": "iCloud"
+        "icloud": "iCloud",
+        "ics": "ICS URL"
       },
       "provider-label": "Provider",
       "save": "Save",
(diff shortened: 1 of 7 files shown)


ItsLemmy:
Thanks for the PR!

## 1. Recurring events are not expanded (significant)

parseICalEvents explicitly does not expand recurrence — its header says so, because CalDAV (<C:expand>) and 

Google expand server-side, so every VEVENT it  receives is already a single concrete instance:
  // Recurrence is not expanded here: callers request server-side expansion (CalDAV <C:expand>)

 A static .ics subscription is the opposite, it relies on RRULE for recurring events (holidays, weekly meetings, birthdays). The parser has  zero RRULE/EXDATE handling. So with this PR, each recurring event renders once, at its original DTSTART, often years in the past and never within the  visible window. Since subscription calendars are heavily recurrence-based, a large fraction of events will silently go missing or appear wrong. This is  the headline limitation. Ideally expand RRULE in the parser (or a windowed expansion step); at minimum it must be called out as a known limitation in the  docs, but it would be great to see it fixed.

## 2. ICS path skips the time-window filter (consistency + memory/cache)

CalDAV and Google constrain server-side to [now − 31d, now + 90d] (kWindowBefore/kWindowAfter). fetchIcs parses the entire file and hands all of it to accountDone → rebuildSnapshot → saveCache, with no window filter anywhere downstream (I checked rebuildSnapshot — it only sorts/merges). A multi-year  holiday feed or busy calendar dumps thousands of events spanning years into memory and the on-disk cache. The ICS path should filter parsed events to the  same [now - kWindowBefore, now + kWindowAfter] window after parsing, for parity with the other providers.

ItsLemmy:
PR was automatically closed due to the branch deletion, dont hesitate to come again with a new one on Main

PR #2943, opened 2026-06-09
feat(session menu): change default session menu to use lock_and_suspend

mattvilim:
# Pull Request

## Motivation
Just my 2c...feel free to ignore this PR if you disagree. The default session menu item to suspend does not lock, which I think is a less common choice than `lock-and-suspend`. I changed the default option to `lock-and-suspend` while users with a custom session config would not be affected. I also changed the glyph for `lock-and-suspend` to use the suspend symbol (pause).

Ideally, I think `suspend` and `lock-and-suspend` would make more sense as `suspend` (defaults to lock) and `suspend-no-lock` if a user wants it for some reason (seems a less common choice). But that would be a breaking change that needs IPC changes, doc updates, i18n, etc., so I'd want buy-in from you guys before doing anything.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring

## Testing

- [X] Tested on niri
- [ ] Tested on Hyprland
- [ ] Tested on sway

## Screenshots / Videos

## Checklist
- [X] Self-reviewed my code

### DIFF
--- a/src/config/config_types.cpp
+++ b/src/config/config_types.cpp
@@ -66,7 +66,7 @@ std::vector<SessionPanelActionConfig> defaultSessionPanelActions() {
           KeyChord{XKB_KEY_2, 0}
       },
       SessionPanelActionConfig{
-          "suspend", true, std::nullopt, std::nullopt, std::nullopt, SessionActionButtonVariant::Default,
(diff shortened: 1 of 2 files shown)


ItsLemmy:
Thanks for the PR, this also gave me an idea for a minor refactor.

PR #2945, opened 2026-06-09
Fix issue with enabling Bluetooth when soft blocked

notiant:
Enabling Bluetooth when it's soft blocked doesn't work at the moment since `rfkill_helper` reads the type file as an integer although it's a string.


Ly-sec:
Looks good to me, I'll merge it. Thank you very much!

PR #2947, opened 2026-06-09
feat(niri): add layer-rule for noctalia-backdrop namespace

YuvrajSinghGitbub:
Added the missing layer rule that will allow the niri backdrop to be set to current noctalia-wallpaper


ItsLemmy:
We cannot really add that to the default template, as there are 3 different way to handle this.

https://docs.noctalia.dev/v5/compositor-settings/niri/

YuvrajSinghGitbub:
Sorry, I didn't thought about that. That does makes sense.

PR #2953, opened 2026-06-09
feat(hyprland): add color exports to the template

Hermanest:
https://github.com/noctalia-dev/noctalia/issues/2927

This PR allows to customize what noctalia does to hyprland. With these changes importing will change to the following: `require("noctalia").apply_theme()` where `apply_theme` can be omitted if you only need the colors.

### DIFF
--- a/assets/templates/hyprland/apply.sh
+++ b/assets/templates/hyprland/apply.sh
@@ -44,7 +44,7 @@ detect_mode() {
 
 apply_lua() {
   local include_line='-- For Noctalia Color templates
-require("noctalia")'
+require("noctalia").apply_theme()'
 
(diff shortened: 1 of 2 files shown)


ItsLemmy:
But this will silently break all hyprland user's config as the colors wont be automatically applied anymore.

Hermanest:
> But this will silently break all hyprland user's config as the colors wont be automatically applied anymore.

The script generates a version with `apply_colors` by default, so this shouldn't be a problem. Regarding existing configs: i think it's okay to change the schema since v5 isn't released yet. If you think its not, we could add a migration.

ItsLemmy:
Thanks, added a migration so no-one is left behind with non updating colors

PR #2954, opened 2026-06-09
feat: add lock screen login scale setting

tomsch:
## Summary
- add `[lockscreen].login_scale` with schema validation and example config
- scale the real lock screen login panel, password field, padding, button, glyph, and panel radius
- expose the setting in Settings > Security > Lock screen
- cover the new config path and clamp behavior in existing config tests

## Tests
- `nix build --impure --expr 'let flake = builtins.getFlake (toString ./.); pkg = flake.packages.x86_64-linux.default; in pkg.overrideAttrs (old: { doCheck = true; checkPhase = "meson test --print-errorlogs config_schema_roundtrip config_path_resolution"; })' --extra-experimental-features 'nix-command flakes'`

### DIFF
--- a/assets/translations/en.json
+++ b/assets/translations/en.json
@@ -1518,6 +1518,10 @@
           "description": "Choose which monitors show the lock screen; leave empty to show on all monitors",
           "label": "Monitors"
         },
+        "login-scale": {
+          "description": "Size multiplier for the lock screen login and password field",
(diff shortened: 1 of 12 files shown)


tomsch:
maybe its better suited for die lockscreen widget editor settings, i quick added it for now.

PR #2960, opened 2026-06-09
feat(i18n): add Dutch (nl) translation

dotsem:
Add translations for Dutch. This gives a good baselayer for most translations.
I'll continue to check and review any potential translation errors and further cross reference them with other WM and MS Windows.
This overrides the 3% already present dutch translation (translated by me).

Future changes will be done via the website.

## AI transparency
This is an AI‑assisted translation with a human review pass and automated structural/placeholder verification against the official Context.


notiant:
I wouldn't necessarily use Windows as a prime reference since the target user base is quite different from Noctalia and people might be more used to certain English terms (like "Launcher" instead of "Start"). Did you use the prompt context for the initial creation?

dotsem:
I did use the prompt context. I explicitly told the model to change those words as they fitted better. 
If preferred I could use KDE as a reference or just leave them as launcher & tray.

Mathew-D:
I think you need to do the translation on https://i18n.noctalia.dev/projects/noctalia

dotsem:
> I think you need to do the translation on https://i18n.noctalia.dev/projects/noctalia

I already did some, but it takes a long time cross referencing them. Via this way i can write them all & test them directly in the shell.
After an initial PR any future translations will be shifted to [https://i18n.noctalia.dev/projects/noctalia ](https://i18n.noctalia.dev/projects/noctalia)

PR #2963, opened 2026-06-09
Support for Non-GNU systems

RangHo:
I'm daily-driving Chimera Linux, which is a Non-GNU Linux distro with clang and BSD coreutils as main components. Noctalia fails to build on such Non-GNU systems as of now.

Currently, Noctalia depends on some GNU-specific extensions and assumptions to build, and this PR attempts to address those.

Currently, the following errors are identified (those with fixes are ticked):

- [x] GNU extension behavior of the `-i` option in `sed` (in-place replacement must provide backup file suffix in POSIX standard)
- [x] Clang requires a separate feature flag for ABI-unstable experimental features (like `<chrono>` and `<format>`)
- [x] libc++ declares filesystem timestamps to be `__int128_t`, requiring extra work to format using `std::to_string`
- [x] libc++ includes `<cmath>` in `<format>`, causing name clash of `log` in certain source files

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -11,6 +11,16 @@ project('noctalia', ['c', 'cpp'],
 cc  = meson.get_compiler('c')
 cpp = meson.get_compiler('cpp')
 
+
+# Clang requires an opt-in flag to properly use experimental C++ libraries.
+# GCC/libstdc++ doesn't need this since these libraries are unconditionally
+# exposed.
+is_libcxx = cpp.get_define('_LIBCPP_VERSION', prefix: '#include <version>' ) != ''
+if is_libcxx
+  add_project_arguments('-fexperimental-library', language: 'cpp')
+  add_project_link_arguments('-lc++experimental', language: 'cpp')
+endif
+
 # Sanitizers are enabled via meson's built-in -Db_sanitize=address,undefined.
 # pipewire_lsan.cpp is only needed when address sanitizer is active.
 sanitize = get_option('b_sanitize') != 'none'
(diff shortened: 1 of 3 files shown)


RangHo:
Note that libc++ including `<cmath>` problem seems to be resolved in llvm/llvm-project@c2e739f3bb2d724170c21385a8df46c69983c514, but this fix hasn't landed in any versions of libc++ so far.

RangHo:
The name clash issue has been resolved by renaming the logger constant from `log` to `kLog`, matching the naming convention of other source files in the tree. Now it builds fine using clang and libc++.

ItsLemmy:
Thanks, the fixes look good! One adjustment before merge: the experimental-library flags are really a libc++ requirement, not a clang one.

Clang defaults to libstdc++ (libc++ is opt-in), so building with CXX=clang++ on a glibc distro would inject -lc++experimental and fail to link on a config that builds fine today.

Could you gate on the stdlib instead of the compiler?  Something like: 

```
  is_libcpp = cpp.compiles('#include <version>\n#ifndef _LIBCPP_VERSION\n#error\n#endif', name: 'libc++ in use')

  if is_libcpp
    add_project_arguments('-fexperimental-library', language: 'cpp')
    add_project_link_arguments('-lc++experimental', language: 'cpp')
  endif
```
  That should keep Chimera/libc++ working without breaking clang-on-libstdc++.

RangHo:
Ah, I missed that part, thanks!

I've followed the pattern that was already there to check for glibc (`cpp.get_define` instead of `cpp.compiles`) but the rest of the logic is the same.
Technically one _could_ choose to use libc++ with g++ but I guess that kind of exotic requirement is for them to solve themselves.

PR #2964, opened 2026-06-10
fix(nightlight): smoother transition on/off with easing

mattvilim:
# Pull Request

## Motivation
When nightlight is enabled/disabled, the gamma update happens at discrete 100ms ticks. Visually, the discrete steps are very obvious which is jarring. I reduced the time between animation ticks to 20ms (which is still greater than 16ms frame interval at 60Hz) and added easing at both ends. Visually, the transition looks very smooth now.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring

## Testing
I tested on both 60Hz and 240Hz monitors with Niri. The transitions on both appear visually smooth, and I can't distinguish the discrete steps anymore.

- [X] Tested on niri
- [ ] Tested on Hyprland
- [ ] Tested on sway

## Screenshots / Videos

## Checklist
- [X] Self-reviewed my code

### DIFF
--- a/src/system/gamma_service.cpp
+++ b/src/system/gamma_service.cpp
@@ -2,6 +2,7 @@
 
 #include "core/log.h"
 #include "ipc/ipc_service.h"
+#include "render/animation/animation.h"
 #include "system/day_night_schedule.h"
 #include "wayland/wayland_connection.h"
 #include "wlr-gamma-control-unstable-v1-client-protocol.h"
@@ -19,7 +20,8 @@ namespace {
   constexpr Logger kLog("gamma");
 
   constexpr float kTransitionDurationMs = 1500.0f;
-  constexpr int kTransitionIntervalMs = 100;
+  constexpr int kTransitionIntervalMs = 20;
+
   constexpr auto kScheduleRecheckInterval = std::chrono::minutes(1);
 
   const zwlr_gamma_control_v1_listener kGammaControlListener = {
@@ -362,8 +364,9 @@ void GammaService::tickTransition() {
   m_transitionProgress = std::min(
       1.0f, static_cast<float>(std::chrono::duration<double, std::milli>(elapsed).count()) / kTransitionDurationMs
   );
+  const auto easeProgress = applyEasing(Easing::EaseInOutCubic, m_transitionProgress);
   const int interpolated = static_cast<int>(
-      std::lerp(static_cast<float>(m_transitionFromKelvin), static_cast<float>(m_targetKelvin), m_transitionProgress)
+      std::lerp(static_cast<float>(m_transitionFromKelvin), static_cast<float>(m_targetKelvin), easeProgress)
   );
   if (interpolated != m_currentKelvin) {
     applyGammaToAll(interpolated);


ItsLemmy:
This is not a bug but a feature.

On some (AMD) dGPU Niri/Smithay is super slow to apply the gamma. Making it tick more often just stack the requests make the entire desktop lag super bad you basically cannot move the mouse anymore during the transitions. You can also see it if starting the shell at night, the wallpaper animation will struggle on startup.

So for now the default should not be changed.  If Smithay ever fix the underlying problem we can revisit. 

We could however go with a setting in the GUI that defaults to the current behaviour so that  people can opt-in to the smoother behavior.

mattvilim:
Interesting, apologies for the regression then. I think better to avoid more configs (where will the config options end if you're allowed to set minutiae for every random feature... 😝).

Maybe there's some cleaner solution that would allow smaller intervals for all cases without causing issues...would have to do some research. I do know V4 transition with wlsunset was smooth (on my HW at least), so I'd be curious to look what they do there (assuming AMD/Niri also worked).

ItsLemmy:
wlsunset does not do transition on startup it just jump there and I hate that.

it does transition smoothly when already running and it's time for nightlight, but it does it over one minute, so there is plenty of time for Smithay to process each change.

We could do something like that, but it would mean that if you start you computer at night time it would take one minute or so to reach the final gamma.

mattvilim:
@ItsLemmy I see. :+1: 

If you're interested, I was able to reproduce and fix the AMD issue on my integrated graphics. The 20ms updates were causing UI lag on the transition like you said. So I ended up using `wl_display_sync` to get a callback when the compositor processes each gamma update, so Noctalia can rate limit updates at the compositor sink rate. I tested it on integrated AMD and discrete NVIDIA and don't see the lag anymore on AMD. I also added some telemetry to measure the round-trip update time and am seeing 30ms on integrated AMD @ 60Hz and 8ms on NVIDIA @ 240Hz.

The branch is here if you're interested: https://github.com/mattvilim/noctalia/commit/7e65f46824111b65fffe0fce6506c4f1b6b756c8 Although, I understand if you want to keep things simple with a slow, dumb timer. :smile:

ItsLemmy:
Tried your branch and it is still laggy for me on 7800xt @ 165Hz
I had big hopes xD

If you are interested here is a discussion I had with the author of a similar software that we planned to use with noctalia v3 but never did the switch. https://github.com/psi4j/sunsetr/issues/21

mattvilim:
:sob: Interesting, this is a lot more complicated that I had expected. Hehe Still not clear if it's a compositor or driver issue.

I also found [this thread](https://gitlab.freedesktop.org/drm/amd/-/work_items/623). The original thread is old, but an update in 2023 indicates Sway/wl_roots don't stutter anymore which means driver might be OK. (who knows which card/driver combination though and things may have regressed since then)

I wonder if it stutters in Sway. If not, then maybe there is a possible Smithay fix. The thread suggests a wlroots change that might be relevant (but who knows). I can't debug further without an AMD card, and you have more important things to work on. :smiley:

PR #2965, opened 2026-06-10
launcher: don't fallback to double-forking if systemd-run fails

jinliu:
Since the systemd-run path is only used on an option which has to be explicitly enabled by the user, if it fails, it's better to just fail the launch, so the user can be aware of the issue.

Still fallback to double-forking if systemd is not detected, in case there's a leftover in the config file enabling the systemd launcher on a non-systemd setup. In this case, there's no such option in settings UI, so the user won't even know that the option is enabled, anyway.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -921,6 +921,7 @@ process_test = executable('process_test',
   sources: files(
     'tests/process_test.cpp',
     'src/core/process.cpp',
+    'src/core/log.cpp'
   ),
   include_directories: _noctalia_inc,
 )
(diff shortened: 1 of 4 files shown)


Ly-sec:
Hey @jinliu I'd merge it like that but perhaps we should inform the user with a notification toast or something so it's very obvious? Not sure

jinliu:
> Hey @jinliu I'd merge it like that but perhaps we should inform the user with a notification toast or something so it's very obvious? Not sure

Good idea. But that would be like a `runAsync(systemd-run)` with a callback, but since the callback won't be in the main thread, I don't know how to make a toast from it.

PR #2967, opened 2026-06-10
bluetooth: try getting battery percentage from upower if missing from bluez

jinliu:
Bluetooth widget / tab will now also query UPowerService for battery percentage if BluetoothService doesn't provide it.

An example of this is the Playstation 5 controller.

### DIFF
--- a/src/shell/bar/widget_factory.cpp
+++ b/src/shell/bar/widget_factory.cpp
@@ -198,7 +198,8 @@ std::unique_ptr<Widget> WidgetFactory::create(
   if (type == "bluetooth") {
     const bool showLabel = wc != nullptr ? wc->getBool("show_label", false) : false;
     const bool hideWhenNoConnectedDevice = wc != nullptr ? wc->getBool("hide_when_no_connected_device", false) : false;
-    auto widget = std::make_unique<BluetoothWidget>(m_bluetooth, output, showLabel, hideWhenNoConnectedDevice);
+    auto widget =
+        std::make_unique<BluetoothWidget>(m_bluetooth, m_upower, output, showLabel, hideWhenNoConnectedDevice);
     widget->setContentScale(contentScale);
     return widget;
   }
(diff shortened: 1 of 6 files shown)


ItsLemmy:
Thanks for this! a few things before merge:
  
Blocker: the toLower will break the lookup. deviceForSelector → selectorMatchesField does a case-sensitive match and doesn't normalize the stored UPower fields. BlueZ reports Address in uppercase and UPower mirrors that uppercase MAC into the device serial, so a lowercased selector (aa:bb:...) won't match the uppercase serial (AA:BB:...),  the lookup returns null and the feature is inert. Please pass d.address verbatim instead of StringUtils::toLower(d.address) at both call sites (and you can drop the two new util/string_utils.h includes). Verbatim matches serial exactly, and it keeps the shared selector matcher case-sensitive, making it case-insensitive would change behavior for the user-facing battery device selector too.

Staleness of the fallback value. The UPower-derived percentage isn't part of any invalidation key, and luetoothService emits no signal when UPower's percentage changes, so the value can drift after it's first shown:
- BluetoothTab::listKey() includes bluez batteryPercent but not the UPower fallback, so it rebuilds on bluez battery changes but not UPower ones.
- BluetoothWidget::syncState()'s early-return key doesn't include battery at all.

 It'll render ok on connect (that triggers a rebuild), but won't update afterward. A complete fix needs both: subscribe to UPowerService's change callback to pump requestUpdate()/rebuild, and fold the resolved percentage into the keys so the rebuild isn't suppressed. It's fine to scope that to a follow-up if you'd rather.

Minor:
  - control_center_panel.cpp still has a stale (void)upower; now that upower is actually used.
  - The new upower_service.h includes in the two headers can be forward declarations (class UPowerService;) — only used as pointers there.
  - static_cast<int>(percentage) truncates rather than rounds (minor, and inconsistent with the uint8_t cast in the widget).

jinliu:
`toLower` is required for the PS5 controller. I haven't researched why, but maybe we need to try both variants.

I think maybe I should move the whole logic to `BluetoothService`, i.e., let it watch `UPowerService` and update `batterPercentage`?

ItsLemmy:
The logic should definitely be in the service.

ItsLemmy:
thanks for the PR, but it's gone stale, come again some other time if you feel like it. ty!

PR #2973, opened 2026-06-10
feat(i18n): add & rename chinese translations

MetsukiMio:


jinliu:
Looks good. Maybe rename to `zh-CN.json`, or do we want to make this the fallback for all `zh-*` locales?

Obelusod:
While reviewing the PR, I noticed that many translations appear to be machine-translated or generated with the help of an LLM, and some entries do not seem to have been fully proofread in their actual UI context. It might be worth doing another review pass before merging.

Some examples:

* `caffeine` → `咖啡因`

  In Linux desktop environments, this usually refers to preventing sleep / idle inhibition rather than the chemical substance itself. Something like `保持唤醒` would be more appropriate.

* `capsule` → `药丸`

  The translation currently mixes `药丸` and `胶囊`. In Chinese UI/UX terminology, `胶囊` is much more common and should be used consistently.

* `Total width budget in pixels` → `总宽度预算`

  `总宽度限制` would sound more natural.

* `backdrop` → `背景`

  This conflicts with `background`. In Noctalia, `backdrop` refers to the layer behind all workspaces and visible in Overview mode, so it should be distinguished from a normal background.

* `fancy-audio-visualizer` → `花式音频可视化`

  `花式` feels awkward in Chinese and does not fully convey the intended meaning.

* `ghost` → `幽灵`

  As a UI style variant, this is closer to a transparent/backgroundless button style rather than a literal ghost.

* `destructive` → `破坏性`

  As a button/style variant, this wording feels unnatural in Chinese UI terminology.

* `ultra-heavy` → `超重`

  This refers to font weight rather than physical weight.

* `Search glyphs…` → `搜索图形…`

  The translation mixes `图形` and `图标`. These should be translated consistently.

* `Audio Overdrive` → `音频超驱`

  This option appears to allow volume levels above 100%; the current translation may not clearly convey that behavior.

* `Night Light` → `夜灯`

  Most Linux desktop environments and operating systems use terms closer to `夜间模式`.

There are also several other translations that feel overly literal or inconsistent with the surrounding context. Some technical terms may be better left untranslated as well.

For reference, several contributors are currently proofreading and translating strings manually on:

`https://i18n.noctalia.dev/`

Feel free to join us there if you'd like to help improve the Chinese localization.

ItsLemmy:
translations must be contributed via our webapp @ https://i18n.noctalia.dev/projects/noctalia

PR #2974, opened 2026-06-10
feat(nix): add config validation to hjem module

Nivalux:
Since there is an cli option to validate the noctalia config, I added it as an option to the hjem module to validate the config.toml at build time.

While testing I could not yet make it error out, but it did provide warnings for wrong config options, which you can see in the build log.

The same can probably be added to the home-manager module, but I can't test that.

@linusammon

An example of the build output with a wrong section (misspelled theme):
```
noctalia-config> WARN  theAsme: unknown section
noctalia-config>
noctalia-config> ✓ Config is valid (1 warning(s))
```

### DIFF
--- a/nix/hjem-module.nix
+++ b/nix/hjem-module.nix
@@ -14,7 +14,19 @@ let
   toml = pkgs.formats.toml { };
   json = pkgs.formats.json { };
 
-  configToml = toml.generate "config.toml" cfg.settings;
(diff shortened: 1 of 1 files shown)


linusammon (approved):
That is pretty neat, looks good to me for now, thanks! I'll look into this some more once `noctalia config validate` supports files.

PR #2981, opened 2026-06-10
feat(lockscreen): add lock_on_startup config option

allmeta:
## Summary

Adds a `[lockscreen] lock_on_startup` config option that locks the session immediately when noctalia starts.

## Motivation

Getting a locked session at startup today means running a separate lock screen (greeter) in front of the compositor, which spawns two Wayland sessions and makes startup into niri noticeably stuttery. Just starting the session normally and having noctalia lock it on startup is a lot smoother.

Doing that externally doesn't work cleanly either: spawning `noctalia msg session lock` from the compositor's autostart races shell startup — the IPC client makes a single `connect()` attempt (`src/ipc/ipc_client.cpp`), and at that point the socket doesn't exist yet, so the command fails with `error: noctalia is not running`. Working around it needs a shell retry loop, and even then the desktop is briefly visible and unlocked before the lock lands.

With this option the lock engages in a startup phase right after `initUi`, so the lock request is flushed to the compositor before plugins, IPC, and hooks initialize — no second session, no race, no unlocked window.

## Implementation notes

- The compositor's `locked` event (which fires the `session_locked` user hook) arrives in the main loop, after `HookManager::reload`, so hooks behave normally for the startup lock.
- `LockScreen::lock()` already defers when no outputs are connected, so startup ordering stays safe on slow output enumeration.
- Settings toggle under **Security → Lock Screen**, documented in `example.toml`, covered by the schema roundtrip test.
- Translations added for the locales that already translate the settings schema (de, es, hu, it, nl, pl, pt-BR, ru, uk-UA, vi, zh-Hans); be/fr/nn fall back to English like the rest of their settings strings.

## Testing

- `meson test` passes (21/21) on debug and release builds.
- `tools/i18n-check.py` passes.
- Tested on niri: with `lock_on_startup = true` and a plain `spawn-at-startup "noctalia"`, the session comes up locked.


ItsLemmy:
Hey thanks for the PR!

We have a decent greeter now, so I don't think this hack is such a good idea.
https://github.com/noctalia-dev/noctalia-greeter
 
Also please note for future PRs, don't machine translate all languages. Just add the english strings. The translation team will take care of the rest as we have a specific web app for that. 

https://i18n.noctalia.dev/projects/noctalia

Thanks again.

Ly-sec:
I have to agree with @ItsLemmy , this is a hacky workaround and given that we already have a greeter by now I don't really think this is needed.

PR #2982, opened 2026-06-10
feat(bar): add per-bar and per-widget icon color setting

calmbook1300:
This PR adds an optional color to the bar settings and the bar widget common settings.  

In most cases this only required replacing `widgetForegroundOr` with a similar `widgetIconColorOr`.  The passed color role is unchanged so with default settings there will be no visible change.  When a foreground color is set via the "Widget Color" setting it now only affects the text/dynamic areas of bar widgets.  The new "Widget Icon Color" setting can be used to independently change the color of glyph/static areas of widgets.

In widgets with two states `widgetIconColorOr` is applied to the state that uses `ColorRole::Primary` or the state that best represents "on" rather than "off" (e.g. widgets that toggle between `ColorRole::OnSurface` and `ColorRole::OnSurfaceVariant`).  In widgets with three states generally the color is applied only to the most logical "on" state.


ItsLemmy:
Thanks for the PR, I like the idea. I'm gonna merge as is and rework it a bit so the colors precedence are more unified and can actually be documented.   Right now icon color and text color are inheriting stuff differently.

PR #2985, opened 2026-06-11
settings: add an "force en translation" toggle on top

jinliu:
This should be convenient for translators when proofreading translations.
When you find something in doubt, just turn on this toggle to check the
original English text, without needing to go to the translation website
and search.

To enable it, set the environment variable NOCTALIA_TRANSLATOR=1 before
launching Noctalia. E.g.:

    NOCTALIA_TRANSLATOR=1 noctalia

### DIFF
--- a/src/shell/settings/settings_window.h
+++ b/src/shell/settings/settings_window.h
@@ -239,6 +239,7 @@ private:
   std::string m_selectedSection;
   std::string m_statusMessage;
   std::string m_pendingResetPageScope;
+  bool m_forceEnTranslation = false;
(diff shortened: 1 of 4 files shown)


ItsLemmy:
I like the idea, but I think it could be better. Right now this assumes translator run the debug build, which most probably don't as they are not developers.

Maybe via a ENV var would be better.

jinliu:
> Maybe via a ENV var would be better.

Good idea. Will do.

ItsLemmy:
Thanks

PR #2987, opened 2026-06-11
feat(greeter): copy avatar path if exists to staging directory for greeter appearance helper

SaverinOnRails:
Syncs the avatar path if set to the greeter config so that it can be used later in place of the user glyph in the greeter:
Screenshot from 2026-06-11 13-31-06


Ly-sec:
Hey @SaverinOnRails , as mentioned in discord I implemented proper AccountsService support so it should be working very well now, no matter how many users :). I'll close this one. Thank you very much though :heart:

PR #3001, opened 2026-06-13
fix(lockscreen): run PAM on background thread to avoid blocking UI

mattvilim:
## Summary

The PAM call on the lock screen is blocking and when a distro has `pam_faildelay` configured, typing a wrong password will cause Noctalia's main thread to block for a couple seconds. Before, blocking the UI thread naturally made the UI freeze to give a visual indication a failed auth attempt is in-progress. Now that it runs async, I updated the lockscreen UI to unfocus and disable the lock text box and button to visually indicate the lock request is in-flight (this only applies if password was wrong and there's a `faildelay`).

I think this may have also fixed a difficult to reproduce issue where the lock screen would very rarely freeze indefinitely because PAM call never returns. I think it might be related to some circular dependence (maybe D-Bus?) where forward progress of the PAM call is blocked because the main Noctalia thread can't sink D-Bus messages. (Uncertain because it happens to me so rarely)

## Motivation

Stalling the main thread for seconds is not good and may be the cause of intermittent deadlock I was seeing.

The UI changes are just for the case where password is incorrect and user should wait a couple seconds before trying again. Without these changes, the password box can be focused, changed, etc without visual indication user can't resubmit yet. The code does guard against re-submission if the async PAM thread hasn't finished yet.

The PAM thread shares no mutable state with the main thread. The PAM thread schedules the auth result handling back on the main thread which runs out of the regular event queue.

## Type of Change

- [X] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Build / packaging

## Related Issue

N/A

## Testing

I tested repeated unlock attempts (both success and failure). I did not test with multiple monitors attached (might be relevant to the UI changes). I also didn't test with fingerprint reader (is that supported on V5 and would there be UI issues?).

## Manual Coverage

- [X] Tested on Niri
- [ ] Tested on Hyprland
- [ ] Tested on Sway

## Checklist

### DIFF
--- a/src/auth/pam_authenticator.cpp
+++ b/src/auth/pam_authenticator.cpp
@@ -96,7 +96,8 @@ namespace {
 
 } // namespace
 
-PamAuthenticator::Result PamAuthenticator::authenticateCurrentUser(std::string_view password) const {
+// Potentially not called from the main thread so should not mutate shared state (class method).
+PamAuthenticator::Result PamAuthenticator::authenticateCurrentUser(std::string_view password) {
   std::string user = currentUsername();
   if (user.empty()) {
     return Result{.success = false, .message = i18n::tr("auth.pam.user-unavailable")};
--- a/src/auth/pam_authenticator.h
+++ b/src/auth/pam_authenticator.h
@@ -10,6 +10,6 @@ public:
     std::string message;
   };
 
-  [[nodiscard]] Result authenticateCurrentUser(std::string_view password) const;
+  [[nodiscard]] static Result authenticateCurrentUser(std::string_view password);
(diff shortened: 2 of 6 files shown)


Ly-sec:
Thanks for the PR!

The async PAM work is already on `main`. `LockScreen::tryAuthenticate()` runs `PamAuthenticator::authenticateCurrentUser()` on a background `std::thread`, posts the result back with `DeferredCall::callLater`, and uses `m_authenticating` to block re-submission and disable the password field / login button while auth is in flight.

Given that, I don't think there's anything left to merge from this branch, happy to close as superseded. Appreciate you digging into the `pam_faildelay` freeze issue.

PR #3007, opened 2026-06-13
fix(desktop-clock): stop digital clock jitter on second changes

odtgit:
## Problem

The digital desktop clock visibly shifts every second. With a seconds format (e.g. `{:%H:%M:%S}`) and a proportional font, the digits and the colon jump left/right on each tick.

## Root cause

`DesktopClockWidget::doUpdate()` re-measures the label every second, so the content width changes as digit advance widths change. That reflows the widget:

- **Boxed** widget: `intrinsicWidth()` returns the fixed box width, but the content is re-centered *inside* the box using the live width (`DesktopWidget` content centering), so the glyphs drift within the tile.
- **Unboxed** widget: the intrinsic width itself changes, so the whole widget re-centers at its anchor.

## Fix

Pin the digital label to the width of its **widest-digit** rendering (`setMinWidth == setMaxWidth`) and **left-align** it:

- The reserved width is constant for a given format/font, so the box never reflows on a digit change.
- Left alignment keeps the leading glyphs (e.g. the first colon) anchored; only the trailing edge breathes into the reserved slack.
- The stable width is measured off-label via `Renderer::measureText` (so it never clobbers the live label) and is recomputed only when the font identity or the non-digit text changes — never on the per-second path.

Analog style is unaffected.

## Testing

Built with `just build` and run live on niri with a multi-monitor desktop. Verified with a proportional font (Lobster) and `{:%H:%M:%S}`: the clock no longer jitters as the seconds advance, on both boxed and unboxed widgets.

## Note

For a proportional font the text now sits anchored within the widest-digit box rather than perfectly optically centered (the slack sits on the trailing side). This is stable and, for a clock, reads better than the previous per-second jitter. Fully centered + stable would additionally require tabular figures, which depends on the font.

### DIFF
--- a/src/shell/desktop/widgets/desktop_clock_widget.cpp
+++ b/src/shell/desktop/widgets/desktop_clock_widget.cpp
@@ -197,7 +197,10 @@ void DesktopClockWidget::create() {
       .fontSize = clockFontSize(contentScale()),
       .color = m_color,
       .fontWeight = FontWeight::Bold,
-      .textAlign = TextAlign::Center,
+      // Left-aligned inside the widest-digit reserved width (see updateStableDigitalWidth):
+      // the box stays put AND the leading glyphs (the first colon) don't drift as digits
+      // change advance width — only the trailing edge breathes into the reserved space.
+      .textAlign = TextAlign::Start,
   });
   m_digitalRoot->addChild(std::move(label));
   rootNode->addChild(std::move(digitalRoot));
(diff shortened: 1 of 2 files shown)


jinliu:
In addition to this, you can also request the "tnum" font feature, so that in supported fonts, digits will have equal widths.

ItsLemmy:
Thanks for the PR, merging.   "tnum" would be a nice addition even if it requires more plumbing, and a lot of font don't implement it.  Adding it to the todo for later.

PR #3008, opened 2026-06-13
feat(desktop-widgets): add show/hide/toggle IPC verbs

odtgit:
Exposes the **Desktop > Widgets** enable toggle over IPC with three new verbs:

- `desktop-widgets-show` — enable desktop widgets
- `desktop-widgets-hide` — disable desktop widgets
- `desktop-widgets-toggle` — flip current state (prints `enabled`/`disabled`)

### Why

Today the only desktop-widget IPC verbs are the edit-mode ones (`desktop-widgets-edit/-exit/-toggle-edit`); there is no way to turn widgets on/off from a script. That makes it impossible to, e.g., reveal widgets only while the desktop is actually visible (a "peek desktop" keybind) instead of having them render behind windows the rest of the time — the audio visualizer in particular keeps doing GPU/PipeWire work while fully occluded.

### How

The verbs flip the same persisted `desktop_widgets.enabled` setting the Control Center toggle uses, via `ConfigService::setOverride({"desktop_widgets","enabled"}, …)`. Because they go through the exact same config path, the IPC state and the settings UI can never disagree. The existing reload callback (`handleConfigReload → applyVisibility`) drives the actual show/hide, and disabling runs `host->hide()` which tears down the widget instances — so it genuinely stops their rendering/compute, not just visually hides them.

No new state is introduced (no orthogonal "hidden" flag), keeping behavior consistent with the existing toggle.

### Testing

Built locally (release) and exercised live on niri:
- `desktop-widgets-hide` writes `enabled = false`; widgets and their compute stop.
- `desktop-widgets-show` restores the default (key removed); widgets reappear.
- `desktop-widgets-toggle` alternates and reports `enabled`/`disabled`.
- Control Center toggle stays in sync with the IPC-driven changes.

Driving it from a peek-desktop script (show on peek, hide on restore) works as intended.


ItsLemmy:
Hey and thanks for the PR!

Two issues here.  First, the Control Center does not use anything desktop widgets related. It's the Settings/Desktop/Desktop Widgets toggle, so these verbs are rewriting a user's saved preference in settings.toml. 

Second, because it persists, hiding widgets via IPC sticks across restart, and a peek-desktop keybind would write to
settings.toml on every keypress.

Basically we don't want any IPC call that write to settings.

The bar-show/bar-hide/bar-toggle verbs are the precedent to follow: they flip an ephemeral runtime visibility flag layered on top of config, so the IPC action tears things down without clobbering the saved setting. You should mirror that for desktop widgets rather than calling setOverride.

odtgit:
Thanks — fully agree on not writing to settings from IPC, and you’re right that a peek keybind hitting `setOverride` would thrash `settings.toml` and clobber the saved preference. I’ve reworked it to follow the bar precedent: an ephemeral runtime override, never persisted, reset on restart.

One deliberate difference I want to flag, since it changes the surface slightly. I made the override **bidirectional** rather than hide-only:

```cpp
enum class RuntimeVisibility { FollowConfig, ForceShown, ForceHidden };
```

- `desktop-widgets-hide` → `ForceHidden`
- `desktop-widgets-show` → `ForceShown`
- `desktop-widgets-toggle` → flips based on current effective visibility
- default `FollowConfig` (restored on restart) honors the saved `desktop_widgets.enabled`

The reason for `ForceShown` being able to reveal widgets even when the saved setting is **disabled**: the motivating use case is *opt-in*, not opt-out. The user keeps desktop widgets **off** as their saved default, and reveals them on demand (a peek-desktop keybind shows them while the desktop is visible, then hides them again). A strictly hide-only flag (`effective = enabled && !hidden`) can’t express that — `show` would be a no-op whenever the saved setting is off, so widgets never appear.

`ForceShown` is still purely ephemeral (no `settings.toml` writes, resets to `FollowConfig` on restart), so it doesn’t touch the saved preference — it just lets the runtime layer override config in *either* direction. That also keeps the bar-style "shown by default, hide during X" case working.

Verified locally (release build): saved `enabled = false`, `show` reveals widgets, `hide`/`toggle` work, and `settings.toml` is never written across any verb. Happy to make it hide-only if you’d prefer to keep the IPC surface minimal and handle opt-in another way — but bidirectional felt like the more complete ephemeral primitive. Branch force-pushed with the rework.

ItsLemmy:
Thanks. I merged it and applied a micro fix to reset to "FollowConfig" if the user changes the toggle in the settings.

Issue #2599, opened 2026-05-02
System tray icons not rendering for apps using themed IconThemePath directories


john-okeefe:
### Description

System tray icons render as blank space for applications that provide their `IconThemePath` using a standard hicolor directory structure. The tray item is registered and has a visible slot, but the icon image fails to load.

### Steps to Reproduce

1. Install and launch Gajim (XMPP client)
2. Observe the system tray in the noctalia-shell bar
3. A blank space appears where the Gajim tray icon should be

### Expected Behavior

The Gajim tray icon should render visibly, the same as other tray icons (Steam, Nextcloud, Heroic, Proton Mail Bridge, etc.).

### Screenshots

N/A -- icon area is blank/empty space.

### Environment

- Distro: Arch
- Compositor: Hyprland
- Noctalia-shell Version: 0.0.12
- Noctalia QS Version: 0.0.12
- Monitor resolution: N/A

### Additional Context

**Root cause:** The `?path=` handler in `Modules/Bar/Widgets/Tray.qml` (line ~422-429) constructs a flat file URL from the StatusNotifierItem's `IconName` and `IconThemePath`:

```qml
if (icon.includes("?path=")) {
    const chunks = icon.split("?path=");
    const name = chunks[0];
    const path = chunks[1];
    const fileName = name.substring(name.lastIndexOf("/") + 1);
    return `file://${path}/${fileName}`;
}
```

This works for apps like Steam that store icons as flat files directly in `IconThemePath` (e.g., `~/.local/share/Steam/public/steam_tray_mono.png`), but fails for apps like Gajim that use a standard hicolor theme directory structure:

```
IconThemePath/
  hicolor/
    index.theme
    scalable/
      status/
        gajim-status-message-new.svg
    32x32/
      status/
        gajim-status-message-new.png
```

The constructed path `file://{IconThemePath}/{IconName}` resolves to a nonexistent file because the actual icon is nested under `hicolor/scalable/status/`.

**D-Bus evidence for Gajim:**

```
IconName = 'gajim-status-message-new'
IconThemePath = '/usr/lib/python3.14/site-packages/gajim/data/icons'
```

[...]

PR #3011, opened 2026-06-13
fix(tray): resolve IconThemePath before clearing generic icon name

rebizzz:
## Summary

Resolve a tray item's `IconThemePath` icon before `looksGenericStatusItemName()` clears the icon name, so Electron/Chromium apps render their real tray icon instead of the generic fallback.

## Motivation

Electron/Chromium apps (Vesktop, Discord, Slack, VS Code, ...) do not send an `IconPixmap` over the StatusNotifierItem interface. They write the tray PNG into a private temp directory and advertise it by name:

```
IconName      = "status_icon_1"
IconThemePath = "/tmp/org.chromium.Chromium.XXXXXX"   (contains status_icon_1.png)
IconPixmap    = (not provided)
```

In `TrayWidget::resolveIconPath()` the name goes into `preferred`, then:

```cpp
if (tray::looksGenericStatusItemName(preferred)) {
    preferred.clear();
}
if (const auto themed = resolveFromTrayThemePath(item.iconThemePath, preferred); !themed.empty()) {
    ...
}
```

`looksGenericStatusItemName()` matches the `status_icon` substring, so `preferred` is cleared to `""`. `resolveFromTrayThemePath()` then hits its empty-name guard and returns nothing, and with no `IconPixmap` to fall back on the item renders the generic placeholder.

Clearing the name is correct for app-identity / pin matching, where `status_icon_1` carries no meaning. But `IconThemePath` is the app's own icon directory and the icon name is the literal file name inside it, so that file lookup should still use the original name. This patch moves the theme-path resolution ahead of the clear. The generic-name filter still applies to the desktop-icon and name matching that follows.

## Type of Change

- [x] Bug fix

## Related Issue

Same class of bug as #2599 (reported on v4, closed as deferred to v5). It still reproduces on current v5.

## Testing

Verified the root cause against the live StatusNotifierItem exported by Vesktop (Electron, v1.6.5) on Niri / NixOS:

1. Read the item's D-Bus properties and confirmed `IconName=status_icon_1`, `IconThemePath=/tmp/org.chromium.Chromium.*`, and no `IconPixmap`.
2. Confirmed `status_icon_1.png` exists in the advertised `IconThemePath`.
3. Confirmed `looksGenericStatusItemName("status_icon_1")` is true, which is what blanks the name before the theme-path lookup.

I was not able to run a full meson build of the shell in my environment, so the patched binary has not been run. The change is a straight reorder of two existing calls with no new logic.

## Manual Coverage

- [ ] Tested on Niri
- [ ] Tested on Hyprland
- [ ] Tested on Sway

Left unchecked: root cause was verified from source and live D-Bus, but the patched binary was not built/run.

### DIFF
--- a/src/shell/bar/widgets/tray_widget.cpp
+++ b/src/shell/bar/widgets/tray_widget.cpp
@@ -827,13 +827,13 @@ std::string TrayWidget::resolveIconPath(const TrayItemInfo& item) {
   } else {
     preferred = item.iconName;
   }
-  if (tray::looksGenericStatusItemName(preferred)) {
-    preferred.clear();
-  }
-
   if (const auto themed = resolveFromTrayThemePath(item.iconThemePath, preferred); !themed.empty()) {
     m_preferredIconPaths[item.id] = themed;
     return themed;
+  }
+
+  if (tray::looksGenericStatusItemName(preferred)) {
+    preferred.clear();
   }
 
   // Match the on-screen request size used when the icon is loaded (see rebuild).


Ly-sec:
Hey @rebiz-loves-milfs , looks good to me. I'll merge it.

PR #3018, opened 2026-06-14
idle: suppress behaviors while caffeine is enabled

1linhao:
## Summary

  Manual Keep Awake ("caffeine") was not included in the inhibit lock count used by
  `IdleManager`.

  That meant idle behaviors like screen off / lock / suspend could still trigger even while
  Keep Awake was enabled from the bar or control center.

  ## Fix

  Include `m_idleInhibitor.enabled()` in the inhibit lock calculation and resync it when:
  - Keep Awake changes
  - ScreenSaverService lock count changes
  - ScreenSaverService becomes unavailable

  ## Result

  With Keep Awake enabled, Noctalia now suppresses idle behaviors consistently instead of only
  registering the Wayland idle inhibitor.

  ## Notes

  I could not fully build locally because `sdbus-c++` is missing in my environment, so this was
  verified by code-path inspection and runtime behavior on the current shell.

### DIFF
--- a/src/app/application.cpp
+++ b/src/app/application.cpp
@@ -386,6 +386,12 @@ void Application::syncClipboardService() {
   }
 }
 
+void Application::syncIdleInhibitLocks() {
+  const std::int64_t screenSaverLocks = m_screenSaverService ? m_screenSaverService->inhibitLocks() : 0;
+  const std::int64_t keepAwakeLocks = m_idleInhibitor.enabled() ? 1 : 0;
+  m_idleManager.setScreenSaverInhibitLocks(screenSaverLocks + keepAwakeLocks);
+}
+
 void Application::run(std::function<void()> startupReadyCallback) {
   initLogFile();
   kLog.info("noctalia {}", noctalia::build_info::displayVersion());
@@ -671,6 +677,7 @@ void Application::initServices() {
   }
   m_idleInhibitor.initialize(m_wayland, &m_renderContext);
   m_idleInhibitor.setChangeCallback([this, shouldRefreshControlCenter]() {
+    syncIdleInhibitLocks();
     m_bar.refresh();
     if (shouldRefreshControlCenter()) {
       m_panelManager.refresh();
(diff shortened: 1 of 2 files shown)


Ly-sec:
Hey @1linhao , CI failed. Do you mind fixing the issue and let us know afterwards?

```
../src/app/application.cpp: In member function ‘void Application::initUi()’:
../src/app/application.cpp:1559:7: error: ‘syncIdleInhibitLocks’ was not declared in this scope
 1559 |       syncIdleInhibitLocks();
      |       ^~~~~~~~~~~~~~~~~~~~
../src/app/application.cpp:1562:7: error: ‘syncIdleInhibitLocks’ was not declared in this scope
 1562 |       syncIdleInhibitLocks();
      |       ^~~~~~~~~~~~~~~~~~~~
../src/app/application.cpp:1567:5: error: ‘syncIdleInhibitLocks’ was not declared in this scope
 1567 |     syncIdleInhibitLocks();
      |     ^~~~~~~~~~~~~~~~~~~~

```

1linhao:
Fixed and force-pushed. The helper is now an `Application` member function so both `initServices()` and `initUi()` can call it.

I use Codex to fix it.




lhao
***@***.***

Ly-sec:
> Fixed and force-pushed. The helper is now an `Application` member function so both `initServices()` and `initUi()` can call it. I use Codex to fix it. lhao ***@***.***

I do wonder, what is your compositor by the way? Just tested it on niri without your PR and it does seem to work properly.

1linhao:
> > Fixed and force-pushed. The helper is now an `Application` member function so both `initServices()` and `initUi()` can call it. I use Codex to fix it. lhao _**@**_.***
> 
> I do wonder, what is your compositor by the way? Just tested it on niri without your PR and it does seem to work properly.

I'm using niri as well. When using an XWayland application, such as "动漫共和国" launched through Wine, and the window is in the "fullscreen-window" state, enabling "keep awake" from the Control Center panel does not keep the system awake.

This PR fixes the issue on my system. An AI suggested that I submit this fix as a PR.

ItsLemmy:
This is not a Noctalia issue.

We rely on the compositor to report the idle time via a wayland protocol. If the compositor keeps reporting idle time when your full screen app is running then it should be fixed upstream and not in Noctalia.

If we accept this PR this creates a precedent where we take responsibility for idle detection and we don't want that.

This should be reported to xwayland-satellite first and see what they say.

Thanks!

PR #3025, opened 2026-06-14
feat: add workspace alert IPC

khughitt:
## Summary
- Add a generic `workspace-alert` IPC surface for marking and clearing workspace attention state.
- Thread canonical workspace keys through compositor workspace models and overlay alerts through the shared workspace read path.
- Resolve window-origin alerts through compositor window assignments, scope keys correctly across outputs, and clear alerts when the user visits the alerted workspace.

## Example Usage

```
noctalia msg workspace-alert-status
noctalia msg workspace-alert-add 4
noctalia msg workspace-alert-status
noctalia msg workspace-alert-clear 4
```

## Test Plan
- `just test workspace_alert_service`
- `just build`
- `git diff --check upstream/main..HEAD`

## Notes
- Alerts reuse the existing urgent/error workspace styling for the initial implementation.
- Direct `workspace-alert-add` expects the canonical key shown by `workspace-alert-status`; `workspace-alert-add-window` resolves window assignments automatically.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -737,6 +737,7 @@ _noctalia_sources = files(
   'src/wayland/text_input_service.cpp',
   'src/wayland/virtual_keyboard_service.cpp',
   'src/compositors/compositor_platform.cpp',
+  'src/compositors/workspace_alert_service.cpp',
   'src/compositors/compositor_detect.cpp',
   'src/compositors/compositor_runtime.cpp',
   'src/compositors/dwl/dwl_workspace_backend.cpp',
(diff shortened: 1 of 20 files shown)


ItsLemmy:
Thanks for this, the code itself is clean and well-tested. My concern is scope: the feature is useful but niche, while most of the diff threads a new Workspace.key through all seven compositor backends, which is our highest-maintenance surface.  So this is a no go.

Could the feature work without the new key, reusing an existing workspace identifier and accepting a documented single-output limitation? That shrinks it to a self-contained alert service + IPC + one overlay hook, no per-backend changes. I'd be glad to take another look in that form.

PR #3034, opened 2026-06-15
Add cusomizable fallback colors when using wallpaper for palette source

Nedra1998:
## Summary

This change adds a configuration field for wallpaper fallback colors, such that when picking a seed color from the wallpaper fails the user can configure the fallback color. It also allows multiple fallback colors which it picks one at random.

Then I also added a color list picker to allow the configuration of this field from the settings UI.

## Motivation

Currently when generating a palette from a wallpaper using the Material themes, the logic scores the colors from the wallpaper image to find one with high chroma to use for the seed color. For low chroma images (e.g. black & white) this fails to find any suitable seed color and uses a hard-coded fallback of #4285f4 for all wallpapers. This change allows user configuration of the fallback.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Build / packaging

## Testing

I ran the standard unit-tests and manual testing. For manual testing I tested these cases:

- Black & White images with a single fallback color, always uses that color
- Black & White images with no fallback color set uses the hard-coded color
- Black & White images with multiple fallback colors pick one at random
- Colorful images ignore the fallback and pick a color from the image

## Manual Coverage

- [ ] Tested on Niri
- [x] Tested on Hyprland
- [ ] Tested on Sway

## Screenshots / Videos

https://github.com/user-attachments/assets/594b147c-1b6d-4d3b-a3a9-d295c4b04e23

## Checklist

## Additional Notes

I made sure to keep the hard-coded color as the default value, and as a fallback if unset or set to an empty list, so behavior should be unchanged by default.

### DIFF
--- a/assets/translations/en.json
+++ b/assets/translations/en.json
@@ -695,6 +695,9 @@
       "list": {
         "add-entry-placeholder": "Add entry…"
       },
+      "color-list": {
+        "add-entry": "Add color"
+      },
       "path-browse": {
         "file-title": "Choose file",
         "folder-title": "Choose folder"
--- a/assets/translations/en.json
+++ b/assets/translations/en.json
@@ -1232,6 +1232,10 @@
         "wallpaper-generation-scheme": {
           "description": "Choose how wallpaper colors are converted into a palette",
           "label": "Wallpaper Generation Scheme"
+        },
+        "wallpaper-fallback-colors": {
+          "description": "Color to use for the palette when the wallpaper doesn't contains suitable colors",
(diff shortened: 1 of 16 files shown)


ItsLemmy:
Thanks for the PR but we are not willing to merge this feature.
This violate Material You schemes, plus add another layer of complexity in the GUI that is not worth it.
There are  already multiple ways to load custom and community palettes easily and you can favourite those palettes per wallpaper.

PR #3050, opened 2026-06-17
feat(wallpaper): add Span fill mode across monitors

AkusenArcade:
## Summary

Add a new `span` wallpaper fill mode that maps a single wallpaper across the whole multi-monitor desktop, so each output shows the slice that matches its position in the compositor layout (instead of every screen repeating the same framing).

## Motivation

The existing fill modes (`center`, `crop`, `fit`, `stretch`, `repeat`) all frame the image independently per output. There was no way to treat the monitors as one continuous canvas. `span` fills that gap: a wide wallpaper now flows continuously across monitors, honoring each output's logical position and any offset between them.

## How it works

- The image is cover-fit onto the desktop **bounding box** (the union of every ready output's logical geometry).
- Each output samples the sub-rectangle at its logical offset, so the slices line up across screens.
- Geometry (offset, monitor size, total desktop size) reaches the wallpaper shader through new uniforms via a small `WallpaperSpanParams` struct; the new branch in `calculateUV` reuses the existing cover math.
- Falls back to `crop` when span geometry is unavailable (e.g. the standalone `WallpaperRenderer` path, or before `xdg_output` geometry is known). Single-monitor setups naturally degenerate to `crop`.
- Output-layout changes refresh every instance so the slices stay coherent when a monitor is moved/added/removed.
- The `repeat` (tile) sampling path is now gated to fill mode `4` only, so `span` (`5`) uses clamped sampling.

The new mode is picked up automatically by the settings UI (it iterates `kWallpaperFillModes`) and the config schema. English string added to `assets/translations/en.json` only, per the translation workflow.

## Type of Change

- [x] New feature

## Testing

- Built with `meson` / `ninja` (C++23, clang-format v22.1.6 clean; `python3 tools/i18n-check.py` → OK).
- Ran on **Niri** with two monitors in an L-shaped layout: DP-1 `3440x1440` at `(0,0)` and HDMI-A-1 `1920x1080` at `(760,1440)`. With `fill_mode = "span"` the wallpaper renders as one continuous image across both outputs, and HDMI-A-1's `760px` horizontal offset is correctly reflected in the slice it shows. Verified the seam between the two outputs is continuous; areas of the bounding box not covered by any output simply stay empty as expected.
- Verified single-monitor behavior matches `crop`.

## Manual Coverage

- [x] Tested on Niri

### DIFF
--- a/assets/translations/en.json
+++ b/assets/translations/en.json
@@ -1125,6 +1125,7 @@
           "crop": "Crop",
           "fit": "Fit",
           "repeat": "Repeat",
+          "span": "Span",
           "stretch": "Stretch"
         },
         "order": {
--- a/example.toml
+++ b/example.toml
@@ -60,7 +60,7 @@ blacklist = []                      # ignore MPRIS players by bus/identity/deskt
 
 [wallpaper]
 enabled             = true
-fill_mode           = "crop"        # center | crop | fit | stretch | repeat
+fill_mode           = "crop"        # center | crop | fit | stretch | repeat | span
 fill_color          = ""            # optional fallback/fill color: prefer a color role token; fixed hex also works
 transition          = ["fade", "wipe", "disc", "stripes", "zoom", "honeycomb"]
 transition_duration = 1500          # milliseconds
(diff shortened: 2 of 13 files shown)


ItsLemmy:
thanks

PR #3052, opened 2026-06-17
fix(notification): support arbitrary notification actions and wrap them dynamically

### LINKED ISSUE #2934: [BUG] Notification action buttons capped at 2 (toast + center) — make configurable or show overflow
### Submission checklist

### Description

v5 silently drops notification actions beyond the first two. A notification sent with 4 actions shows only the first 2 as buttons; the rest are unreachable from the GUI — in **both** the toast and the notification center.

This affects any client that supplies more than two actions (allowed by the freedesktop Desktop Notifications spec). Because the extras are dropped without any affordance, there is no way for a user to reach them.

### Where (v5 branch)

- Toast: `kMaxActionButtons = 2` — `src/shell/notification/notification_toast.cpp:66` (collection loop at `:309` stops once 2 buttons are built).
- Notification center / history: `kHistoryMaxActionButtons = 2` — `src/shell/control_center/notifications_tab.cpp:46` (loops at `:72` and `:445`).

So neither surface exposes more than two actions.

### Steps to reproduce

```sh
notify-send "Test" "four actions" \
  --action="a=One" --action="b=Two" --action="c=Three" --action="d=Four"
```

Only **One** and **Two** render as buttons; **Three** and **Four** are silently dropped — in the toast and in the notification center.

### Expected behavior

All supplied actions should be reachable. Ideally one of:

- a configurable maximum (setting), and/or
- an overflow / "more" affordance for actions beyond the visible ones, and/or
- wrapping actions onto additional rows.

### Actual behavior

Only the first two actions are ever shown; any further actions are silently discarded in both the toast and the notification center.

### Environment information

```text
Noctalia Version: v5.0.0 (noctalia-git 5.0.0.r3175.g732922d96)
Compositor: niri
Distribution: Arch-based (CachyOS)
Installation Method: AUR (noctalia-git)
```

the-daonm:
## Summary

This PR fixes the issue where `noctalia` silently dropped notification actions beyond the first two in both the toast notification and the notification center/history tab. The hardcoded limits of `2` action buttons have been removed. Additionally, action buttons are now dynamically wrapped onto multiple horizontal rows when their cumulative widths exceed the card's available width limit. Within each wrapped row, buttons are configured with `flexGrow = 1.0f` to stretch and share the horizontal space evenly, matching modern UI standards.

## Motivation

This solves the problem where clients (such as `notify-send`) supplying more than two actions are unable to display them to the user, rendering critical notification controls unreachable.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Refactoring
- [ ] Build / packaging

## Related Issue

Closes #2934

## Testing

`just build`, `just test`, `just run` work

## Manual Coverage

- [x] Tested on Niri
- [ ] Tested on Hyprland
- [ ] Tested on Sway

## Screenshots / Videos

image

image

image

## Checklist


src/shell/desktop/widgets/desktop_button_widget.cpp
```diff
@@ -26,8 +26,7 @@ DesktopButtonWidget::DesktopButtonWidget(
     std::optional<ColorSpec> labelColor, ColorSpec hoverBackground
 )
     : m_glyph(std::move(glyph)), m_label(std::move(label)), m_command(std::move(command)), m_variant(variant),
-      m_labelColor(std::move(labelColor)), m_hoverBackground(std::move(hoverBackground)),
-      m_showBackground(showBackground) {}
+      m_labelColor(labelColor), m_hoverBackground(hoverBackground), m_showBackground(showBackground) {}
```
ItsLemmy, line 29: This file should not be in the PR. This would revert a clang tidy fix.

ItsLemmy:
## Main concern: the wrapping algorithm is copy-pasted 3×

  The same greedy-pack + flushRow lambda now appears nearly verbatim in:
  - measureHistoryActionsRowHeight() (measure pass)
  - NotificationHistoryCard::bind() (build pass)
  - layoutNotificationActionsRow() in notification_toast.cpp

The measure pass and the build pass must produce an identical row count or the card height won't match the rendered  content and they only stay in sync by hand-duplicating the loop. That's fragile. 

Extract one helper, e.g.

```
  // returns rows of buttons packed to fit maxWidth, gap between buttons
  std::vector<std::vector<std::unique_ptr<Button>>>
  wrapButtonsIntoRows(Renderer&, std::vector<std::unique_ptr<Button>>&, float maxWidth, float gap);
```

and have all three call sites use it. This also aligns with the project's "generic over hardcoded" preference and removes the measure/build drift risk  entirely.


##  History tab doesn't clamp a lone over-wide button

In the toast's flushRow, a single-button row clamps width:
``` if (rowButtons.size() == 1) btn->setMaxWidth(maxRowWidth); else btn->setMaxWidth(0.0f);```
The history-tab flushRow only does setFlexGrow(1.0f) with no setMaxWidth. A single action with a very long label that exceeds cardTextWidth can overflow the card there. Mirror the toast's clamp for consistency.

##  No upper bound on action count

The freedesktop spec doesn't bound actions, and a buggy/hostile client could send dozens. The toast card has clipChildren = true + maxToastCardHeight, so it's contained; the history card has no such clamp and could grow very tall. Worth a sane cap

the-daonm:
@ItsLemmy Thanks for the review! I've refactored the code to keep it DRY and addressed the issues you pointed out.

ItsLemmy:
Thanks!

PR #3065, opened 2026-06-18
feat(notifications): filter option added to disable permanent notifications

Nivalux:
## Summary & Motivation

My webbrowser sends notifications as permanent notifications (timeout = 0). This sometimes results in a big list of notifications, that I have to click away manually on browser startup.
This PR adds a filter setting, which allows disabling permanent notifications and have them timeout normally.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Build / packaging

## Testing

Running `notify-send -t 0 Test` before and after setting the `allow_permanent` option.

## Manual Coverage

- [x] Tested on Niri
- [ ] Tested on Hyprland
- [ ] Tested on Sway

## Checklist

### DIFF
--- a/assets/translations/en.json
+++ b/assets/translations/en.json
@@ -994,11 +994,13 @@
       "filter": {
         "add": "Add filter",
         "add-title": "Add Notification Filter",
+        "allow-permanent": "Allow permanent",
         "allowed-urgencies-label": "Allowed urgencies",
         "flag": {
           "history": "history",
           "sound": "sound",
-          "toast": "toast"
+          "toast": "toast",
+          "permanent": "permanent"
         },
         "flags-label": "When matched",
         "match-label": "App",
--- a/src/config/config_overrides.cpp
+++ b/src/config/config_overrides.cpp
@@ -522,6 +522,7 @@ namespace {
               row.insert_or_assign("show_toast", item.showToast);
               row.insert_or_assign("save_history", item.saveHistory);
               row.insert_or_assign("play_sound", item.playSound);
+              row.insert_or_assign("allow_permanent", item.allowPermanent);
               if (!item.allowedUrgencies.empty()) {
                 toml::array urgencies;
                 for (const auto& urgency : item.allowedUrgencies) {
--- a/src/config/config_types.h
+++ b/src/config/config_types.h
@@ -216,6 +216,7 @@ struct NotificationFilterConfig {
   bool showToast = true;
   bool saveHistory = true;
   bool playSound = true;
+  bool allowPermanent = true;
   /// Empty = allow low, normal, and critical. Otherwise only listed urgencies pass this filter.
   std::vector<std::string> allowedUrgencies;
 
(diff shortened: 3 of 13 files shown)


ItsLemmy:
Thanks for the PR.

Two blockers for me: 

1. We absolutely don't want to bring back any global filtering. All filtering should be done via the filters. This avoid confusion on what rule wins and also keep the logic centralised in one place.  Please remove  "Allow Permanent" toggle from the General group.

2. We should exempt Urgency::Critical from the coercion. The urgency argument is already available in addOrReplace, so it's just an extra guard, ex: 
```
 if (timeout == 0 && urgency != Urgency::Critical) {
      // … resolve filter and coerce timeout …
  }
 ```
That would keep the feature working for the noisy normal/low notifications from your browser, while leaving critical ones permanent, which is important to keep.

**Comment by Nivalux** [comment]
Thank you for the review.
I only thought of solving it with the filter afterwards. The global option was obsolete afterwards, I agree.
I have removed all the global option parts and added the Urgency::Critical exemption.

**Comment by Nivalux** [comment]
I did some more testing and I think the urgency check needs to be removed. In my case the browser notifications, which get send with `requireInteraction: true` result in this:
`[notification] notification added #5 origin=external from="Vivaldi" urgency=critical timeout=0ms`

If I understand it correctly the urgency check is not required, since you could define two filters for the same app with different `allowed urgencies` selected.

```toml
[notification]
filter_order = [ "filter", "filter-2" ]

    [notification.filter.filter]
    allow_permanent = true
    allowed_urgencies = [ "critical" ]
    enabled = true
    match = "vivaldi"
    play_sound = true
    save_history = true
    show_toast = true

    [notification.filter.filter-2]
    allow_permanent = false
    allowed_urgencies = [ "low", "normal" ]
    enabled = true
    match = "vivaldi"
    play_sound = true
    save_history = true
    show_toast = true
```

Nivalux:
I added the urgency to the filter matching.
This allows to define different filters depending on the urgency. Not sure if "Allowed urgencies" makes it clear, that it is part of the matching condition.

This also removes the hard coded behaviour if it is not part of the matching urgencies, since you can now just configure that normally. 

Something like this is now possible:
The following defines for vivaldi, that critical notifications are not permanent, but are shown as a toast message. Non critical notifications are not shown as a toast message, but are saved in the history.
It is a constructed example, but I tested it in the browser and it works.

```toml
[notification]
filter_order = [ "filter", "filter-2" ]

    [notification.filter.filter]
    allow_permanent = false
    allowed_urgencies = [ "critical" ]
    enabled = true
    match = "vivaldi"
    play_sound = true
    save_history = true
    show_toast = true

    [notification.filter.filter-2]
    allow_permanent = true
    allowed_urgencies = [ "low", "normal" ]
    enabled = true
    match = "vivaldi"
    play_sound = true
    save_history = true
    show_toast = false
```

ItsLemmy:
Thanks! added a tiny fix before merging.

PR #3079, opened 2026-06-19
feat(widget): show spinner when ethernet is resolving

allmeta:
## Summary

The network bar widget now shows a spinner in place of its glyph while a wired (ethernet) link is activating, then resolves to the normal wired icon once connected. The tooltip reports "Connecting" during this phase.

## Motivation

On my office network, ethernet takes ~5s to resolve (DHCP / IP negotiation). During that window the widget just showed the default disconnected glyph, giving no indication that a connection was actually in progress. The spinner makes that resolving state visible.

## Type of Change

- [x] New feature

## Testing

- Verified live against NetworkManager over D-Bus (`busctl`): during an ethernet down/up cycle the physical device transitions `30 → 70 (IP_CONFIG) → 100`, with the device's active connection in `Activating` state during the resolving window — which is what now drives the spinner.

## Manual Coverage

- [x] Tested on another compositor: Niri (observed the spinner during a real ethernet reconnect)

## Screenshots / Videos

image

https://github.com/user-attachments/assets/6394afd9-f4a8-4b1d-b865-9fcb339693c1

## Checklist

## Additional Notes

- Detecting the resolving state is not as simple as reading NetworkManager's `ActivatingConnection` property: while a real link is activating, `PrimaryConnection` stays `/` and `ActivatingConnection` tracks the loopback connection rather than the physical link. Instead, `rebindActiveConnection()` falls back to scanning for a physical ethernet/wifi device whose `State` is between Prepare (40) and Activated (100), and uses that device's active connection. Virtual devices (loopback, bridge, vlan, …) are excluded by device type, so they never surface as a connection.
- The global spinner revolution was slowed from 800ms to 1200ms; at the previous speed the rotation outran the compositor's frame cadence and read as jitter (visible on the bluetooth/scan spinners too).
- Tooltip strings ("Connecting"/"Not connected") follow the existing hard-coded strings in this widget, which are not routed through `en.json`; no new translatable strings were introduced.

### DIFF
--- a/src/dbus/network/network_manager_service.cpp
+++ b/src/dbus/network/network_manager_service.cpp
@@ -38,7 +38,12 @@ namespace {
   constexpr std::string_view kNmWiredConnectionType = "802-3-ethernet";
 
   // NMDeviceType values from NetworkManager D-Bus API.
+  constexpr std::uint32_t kNmDeviceTypeEthernet = 1;
   constexpr std::uint32_t kNmDeviceTypeWifi = 2;
+
+  // NMDeviceState: a device between Prepare and Activated is mid-activation.
+  constexpr std::uint32_t kNmDeviceStatePrepare = 40;
+  constexpr std::uint32_t kNmDeviceStateActivated = 100;
 
   // NMActiveConnectionState
   constexpr std::uint32_t kNmActiveConnectionStateActivating = 1;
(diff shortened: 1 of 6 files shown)


ItsLemmy:
Thanks!

PR #3084, opened 2026-06-19
feat(launcher): tab select in menus

allmeta:
## Summary

Add Tab / Shift+Tab navigation to the launcher and clipboard panels. Tab moves selection to the next entry, Shift+Tab (or `ISO_Left_Tab`) to the previous one. Both wrap around the ends.

## Motivation

Selection could only be moved with the arrow keys. Tab is a common expectation for cycling through list entries and keeps the hands on the home row.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Build / packaging

## Testing

Manually tested in the launcher and clipboard panels: Tab cycles forward, Shift+Tab cycles backward, both wrap at the ends. Tab combined with any modifier other than Shift is left untouched.

## Manual Coverage

- [x] Tested on Niri
- [ ] Tested on Hyprland
- [ ] Tested on Sway

## Checklist

## Additional Notes

No new config keys or user-facing strings. Tab only triggers when no modifiers other than Shift are held, so existing keybinds are unaffected.

### DIFF
--- a/src/shell/clipboard/clipboard_panel.cpp
+++ b/src/shell/clipboard/clipboard_panel.cpp
@@ -2,6 +2,8 @@
 
 #include "config/config_service.h"
 #include "core/deferred_call.h"
+#include "core/key_modifiers.h"
+#include "core/key_symbols.h"
 #include "core/keybind_matcher.h"
 #include "core/log.h"
 #include "core/process.h"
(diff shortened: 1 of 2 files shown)


ItsLemmy:
Hey and thanks for the PR.

I'm not willing to merge this one even if it works fine.

Basically we want to totally move away from hard coded keybinlds in every panels to instead rely on our custom Keybinds (user definable in Settings / Shell).  This is where people should add Tab and Shift+Tab to navigate Up and Down if they want to.  

In noctalia V4 we had the same situation and everyone was trying to PR they favorite keybinds in everypanel  (ex: VIM, Emacs, etc... ) which is clearly not the way to go.

This would also be part of an upcoming global effort, where we want to provide full keyboard navigation (which usually use tab and shift-tab to move focus).

allmeta:
@ItsLemmy That sounds like a much better solution, yes! Thanks for the reply 🙌🏼

PR #3113, opened 2026-06-22
Fix window cycling when pressing the dock item icon

whiskeyPeak:
## Summary

When left clicking on a docker item icon with more than 1 window open, one would expect it to cycle between windows. There is a bug with the current code where the cycle gets stuck because some windows have the same identifier and are treated as one and the same window.

## Motivation

Makes cycling between windows work.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Build / packaging

## Related Issue

N/A

## Manual Coverage

- [x] Tested on Niri
- [ ] Tested on Hyprland
- [ ] Tested on Sway

## Screenshots / Videos

Before:

https://github.com/user-attachments/assets/477eb940-963a-489c-8553-9a339310f2e3

After:

https://github.com/user-attachments/assets/5570813e-2cc7-4f9b-858e-0c297c6b8110

## Checklist

- [ ] This PR is ready for review, or it is marked as Draft.
- [ ] I read and followed the relevant guidance in `CONTRIBUTING.md`.
- [ ] I ran `just format` with clang-format v22+ installed, or this PR has no code changes.
- [ ] I ran the relevant build or test commands, or explained why they were not run.
- [ ] I self-reviewed the changes.
- [ ] I checked for new warnings or errors.

### DIFF
--- a/src/shell/dock/dock.cpp
+++ b/src/shell/dock/dock.cpp
@@ -112,11 +112,24 @@ namespace {
         || (compositors::isKde() && (!window.title.empty() || !window.appId.empty()));
   }
 
-  [[nodiscard]] bool matchesActiveWindow(const ToplevelInfo& window, const ActiveToplevel& active) {
+  [[nodiscard]] bool matchesActiveWindow(
+      const ToplevelInfo& window, const ActiveToplevel& active, const std::vector<ToplevelInfo>& windows
+  ) {
     if (active.handle != nullptr && window.handle == active.handle) {
       return true;
     }
-    return false;
+
+    if (active.identifier.empty() || window.identifier.empty() || active.identifier != window.identifier) {
+      return false;
+    }
+
+    int count = 0;
+    for (const auto& w : windows) {
+      if (w.identifier == active.identifier && ++count > 1) {
+        return false;
+      }
+    }
+    return true;
   }
 
   const ToplevelInfo* nextActivatableWindow(
@@ -129,7 +142,7 @@ namespace {
 
     if (active.has_value()) {
       for (std::size_t i = 0; i < windows.size(); ++i) {
-        if (!matchesActiveWindow(windows[i], *active)) {
+        if (!matchesActiveWindow(windows[i], *active, windows)) {
           continue;
         }
         for (std::size_t offset = 1; offset <= windows.size(); ++offset) {
--- a/src/shell/dock/dock.cpp
+++ b/src/shell/dock/dock.cpp
@@ -116,7 +116,7 @@ namespace {
     if (active.handle != nullptr && window.handle == active.handle) {
       return true;
     }
-    return !active.identifier.empty() && !window.identifier.empty() && active.identifier == window.identifier;
+    return false;
   }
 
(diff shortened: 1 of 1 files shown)


Ly-sec:
Thanks for the PR and the before/after videos, they made the bug easy to follow.

You're right about the root cause. On the wlr foreign-toplevel path we build `identifier` as `appId + ":" + title`, which is not unique when an app has several windows with the same title. `matchesActiveWindow()` then treats them as one window, so cycling gets stuck on the first match.

The one-line fix should work on Niri, Hyprland, and Sway where `ActiveToplevel` usually has a real handle, which matches what you tested.

My concern is that the identifier fallback was not useless. On KDE we use KWin UUIDs, and on some Hyprland/ext paths handles can be null while the identifier is still a proper per-window id. Removing identifier matching entirely means `nextActivatableWindow()` may never find the active index on those sessions. It would fall back to `preferredIdentifier` or the first window in the list, which sounds like "focus the same window again" rather than cycle. KDE was not checked on the PR template, so I would want that verified before merge.

Suggested direction: keep handle matching as the primary path, but only match on identifier when it is unique within the current `windows` list. That fixes duplicate `appId:title` on wlr compositors without breaking KDE or ext-only rows.

Happy to re-review once that is in. Good catch on the cycling bug either way.

whiskeyPeak:
@Ly-sec 

```
  [[nodiscard]] bool matchesActiveWindow(
      const ToplevelInfo& window, const ActiveToplevel& active, const std::vector<ToplevelInfo>& windows
  ) {
    if (active.handle != nullptr && window.handle == active.handle) {
      return true;
    }

    if (active.identifier.empty() || window.identifier.empty() || active.identifier != window.identifier) {
      return false;
    }

    int count = 0;
    for (const auto& w : windows) {
      if (w.identifier == active.identifier && ++count > 1) {
        return false;
      }
    }

    return true;
  }
```

Thoughts on this? I can't test it on KDE right now but this works on my end and should work there too.

whiskeyPeak:
So I managed to test it on Kwin but the dock doesn't show active windows, only pinned ones so the window cycling fix can't be tested. Maybe it's missing some IPC?

Ly-sec:
Thanks for iterating on this @whiskeyPeak . The updated `matchesActiveWindow()` looks good to me.

The root cause is right: on the wlr path we build `identifier` as `appId:title`, which collides when multiple windows share the same title. Identifier matching then treats them as one window and cycling gets stuck.

The fix keeps handle matching as the primary path and only falls back to identifier matching when that identifier is unique in the current window list. That fixes the duplicate-title case without breaking paths that rely on identifier fallback.

I re-tested on Plasma with this branch checked out. The dock still shows running apps, the active entry is still highlighted, and left-click cycling still works on my end so I guess it's good to go.

I'll merge it, thank you very much!

whiskeyPeak:
No problem!

Also my issue with Kwin was due to having run Kwin on a tty after niri, so Noctalia thought the platform was Niri and so the IPC didn't work.

PR #3134, opened 2026-06-24
feat(workspace): add workspace alert IPC

khughitt:
## What this does

Adds a workspace-alert feature. Over IPC, a user can flag a workspace — or the workspace containing a given window — so it renders with the existing `urgent` style until they switch to it. The alert auto-clears when the user activates that workspace.

IPC commands:
- `workspace-alert-add <workspace-id>`
- `workspace-alert-add-window <window-id>`
- `workspace-alert-clear <workspace-id>`
- `workspace-alert-clear-all`
- `workspace-alert-status`

## Why this is a rewrite of #3025

The original PR (#3025) was declined on scope: it introduced a new canonical `Workspace.key` field and threaded it through all seven compositor backends — the highest-maintenance surface in the project — plus output-name-resolver plumbing and widget/snapshot changes.

This version reuses the compositor's **existing workspace identifiers** as the alert token, so the feature collapses to a self-contained unit:
- a `WorkspaceAlertService` (set of tokens + overlay logic),
- the five IPC handlers,
- one overlay hook in each `CompositorPlatform::workspaces()` overload,
- the focused-output auto-clear wiring.

An alert token is matched against a workspace's **id, name, or visible index**, because backends populate different ones (niri only sets the visible index; Sway uses the name as the id; tag compositors use the numeric id). This keeps the feature usable across compositors without threading a new canonical key through every backend. **No per-backend changes.** The only cross-cutting edit is relocating the plain `WorkspaceWindowAssignment` struct between two headers.

## Known limitation

None of these identifiers is guaranteed unique across outputs on compositors that number workspaces per monitor (niri index, mango/dwl tags), so an alert there applies to the matching workspace on every monitor (and activating it on any monitor clears it). This is the accepted per-output trade-off for not introducing a per-backend canonical key; it's documented in the service header.

## Notes

- The overlay reuses the `urgent` flag and only marks **inactive** rows, so alerting the active workspace simply shows once you leave it and clears on return; on per-output compositors this also lets an inactive duplicate alert even when the same workspace is active on another monitor.
- The id/name/index matching runs *after* the metadata backend's `apply()` in `workspaces()`, so the visible index is populated at match time (this is what makes niri — which doesn't expose a usable `id` — work).
- `m_workspaceAlertService` is declared before `m_compositorPlatform` so the raw pointer handed to the platform outlives it.

## Testing

- `just build` passes.
- `workspace_alert_service` unit test: 31/31 (covers overlay suppression of active rows, matching by index/name/id, per-output duplicate-index handling, window→workspace resolution, and validation).
- Verified live on niri: `workspace-alert-add <n>` by visible number, `add-window <id>`, `clear`, `clear-all`, and `status` all behave as expected.
- `git diff --check` clean.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -753,6 +753,7 @@ _noctalia_sources = files(
   'src/wayland/text_input_service.cpp',
   'src/wayland/virtual_keyboard_service.cpp',
   'src/compositors/compositor_platform.cpp',
+  'src/compositors/workspace_alert_service.cpp',
   'src/compositors/compositor_detect.cpp',
   'src/compositors/compositor_runtime.cpp',
   'src/compositors/dwl/dwl_workspace_backend.cpp',
@@ -938,6 +939,16 @@ notification_filter_test = executable('notification_filter_test',
 )
 
 test('notification_filter', notification_filter_test)
+
+workspace_alert_service_test = executable('workspace_alert_service_test',
+  sources: files(
+    'tests/workspace_alert_service_test.cpp',
+    'src/compositors/workspace_alert_service.cpp',
+  ),
+  include_directories: _noctalia_inc,
+)
+
+test('workspace_alert_service', workspace_alert_service_test)
 
 animation_manager_test = executable('animation_manager_test',
   sources: files(
--- a/src/app/application.cpp
+++ b/src/app/application.cpp
@@ -717,7 +717,16 @@ void Application::initServices() {
       m_panelManager.refresh();
     }
   });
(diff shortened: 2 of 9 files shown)


ItsLemmy:
Thanks for the PR, I had to fix the merge since I refactored application.cpp a couple days ago, unfortunately could not push to your branch (you should enable "allow edits from maintainers" it makes things easier)

Anyways, it's committed in 2b524e521ca1baab27ffaf4049ba58539c2b3171

PR #3148, opened 2026-06-25
fix(panel): restore window focus on Hyprland after panel close

Kranzes:
## Problem

On Hyprland with `input:follow_mouse != 1`, closing a noctalia panel leaves the
previously-focused window reported as *active* but unable to receive keyboard
input until another focus transition occurs
(hyprwm/Hyprland#14285, hyprwm/Hyprland#14730).

Root cause for v4's architecture: panels live in a single **persistent**
per-screen `MainScreen` layer-shell surface that never unmaps — it only toggles
`WlrLayershell.keyboardFocus`. When the last panel closes, focus drops to `None`
on a still-mapped surface, which trips the Hyprland refocus issue.
DankMaterialShell fixed the same bug (AvengeMedia/DankMaterialShell#2655) by
holding keyboard focus until its **per-popout** surface unmaps — the unmap is
what makes the compositor refocus cleanly. noctalia's shared/persistent surface
can't release-on-unmap, so it nudges focus back explicitly instead.

## Fix

- Capture the toplevel that holds keyboard focus *before* the first panel grabs
  the keyboard (`PanelService.willOpenPanel`).
- When the active panel fully closes via a normal animated close, re-focus that
  window (`PanelService.closedPanel` → `restoreFocusAfterClose`).
- Gated to **Hyprland + `follow_mouse != 1`**, and skipped for immediate
  (app-launch) closes. The default `follow_mouse=1` and all non-Hyprland
  compositors see **no behavior change**.

Implementation:

- `HyprlandService`: query/expose `input:follow_mouse` (re-queried on `configreloaded`).
- `CompositorService`: mirror it on the facade as `hyprlandFollowMouse`.
- `PanelService`: capture-on-open / restore-on-close via the existing `focusWindow()` path.

## Testing

Set `input:follow_mouse` to `0` (or `2`/`3`), open a bar panel, close it (Escape
or click-out), and confirm keyboard input returns to the underlying app. With
`follow_mouse=1` (the Hyprland default) behavior is unchanged.

## Notes / limitations

- Uses capture-on-open (deterministic) rather than reading focus at close, since
  the cached focus model can be invalidated while the layer surface holds the
  keyboard grab.
- If focus is changed (e.g. alt-tab) *while a panel is open*, close restores the
  pre-panel window. The overview-launcher overlay path (a separate surface) is
  not covered.

### DIFF
--- a/Services/Compositor/CompositorService.qml
+++ b/Services/Compositor/CompositorService.qml
@@ -31,6 +31,10 @@ Singleton {
   // Overview state (Niri-specific, defaults to false for other compositors)
   property bool overviewActive: false
 
+  // Hyprland input:follow_mouse value (default 1, mirrored from the Hyprland backend).
+  // Lets panel logic decide whether it must explicitly restore focus when a panel closes.
+  property int hyprlandFollowMouse: 1
+
   // Global workspaces flag (workspaces shared across all outputs)
   // True for LabWC (stacking compositor), false for tiling WMs with per-output workspaces
   property bool globalWorkspaces: false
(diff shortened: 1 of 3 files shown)


Ape:
Try using noctalia v5 alpha instead. This Hyprland issue should be mostly avoided there.

Kranzes:
I don't like noctalia v5, i think it's really ugly

PR #3155, opened 2026-06-26
Improve media ipc

Ape:
This adds
```
media <..|play|next-player|previous-player>
```
and changes some logic in the mpris service so that it's a lot more convenient and behaves the way I would expect.

## Type of Change
- [x] Bug fix
- [x] New feature

## Manual Coverage
- [x] Tested on Hyprland

## Checklist
- [ ] I will update end-user documentation after merge.

### DIFF
--- a/src/dbus/mpris/mpris_service.cpp
+++ b/src/dbus/mpris/mpris_service.cpp
@@ -708,7 +708,7 @@ void MprisService::registerIpc(IpcService& ipc) {
           return playActive() ? "ok\n" : "error: no active player or Play unsupported\n";
         }
         if (action == "stop") {
-          return stopActive() ? "ok\n" : "error: no active player or Stop unsupported\n";
+          return stopActive() ? "ok\n" : "error: no active player or Pause/Stop unsupported\n";
         }
 
         return "error: invalid media action (use next, previous, toggle, play, stop)\n";
@@ -791,12 +791,26 @@ bool MprisService::stop(const std::string& busName) {
   if (it == m_players.end()) {
     return false;
   }
+
+  if (it->second.playbackStatus != "Playing") {
+    return true;
+  }
+
+  if (it->second.canPause) {
+    if (!callPlayerMethod(busName, "Pause")) {
+      return false;
+    }
+    dismissPlayer(busName);
+    return true;
+  }
+
   if (!canInvoke(it->second, "Stop")) {
     return false;
   }
   if (!callPlayerMethod(busName, "Stop")) {
     return false;
   }
+
   dismissPlayer(busName);
   return true;
 }
--- a/src/dbus/mpris/mpris_service.cpp
+++ b/src/dbus/mpris/mpris_service.cpp
@@ -556,7 +556,6 @@ void MprisService::applyPositionSample(const std::string& busName, int64_t rawPo
       } else if (pausedJumpUs < kPauseRecoveryMinJumpUs) {
         return;
       } else {
-        playerIt->second.playbackStatus = "Playing";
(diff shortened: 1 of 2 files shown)


ItsLemmy:
Review: PR #3155 — "Improve media ipc"

What it does: Adds media play, media next-player, and media previous-player IPC actions; reworks stop() to prefer Pause over Stop (and no longer dismiss the player); and reworks applyPlayerSnapshot so the active/pinned player auto-follows whichever player starts/stops playing. It also drops two pause-recovery playbackStatus = "Playing" overrides. Overall a reasonable UX improvement, but it leaves orphaned subsystems and reverts a deliberate heuristic without explanation.

Findings (most severe first)

1. mpris_service.cpp:785 (removed) — Removing the sole dismissPlayer() call orphans the entire "dismissed players" subsystem.
stop() was the only caller of dismissPlayer(), and dismissPlayer() is the only place that inserts into m_stoppedPlayers (line 2185). After this PR m_stoppedPlayers can never become non-empty, so every isDismissed() guard in chooseActivePlayer() (lines 2062, 2071, 2091, 2100, 2108, 2115), the erases at 770/1746/2034, and dismissPlayer() itself (2179) all become unreachable dead code. Either the dismiss feature is being intentionally retired — in which case this dead machinery should be deleted in this PR — or it's an unintended loss of "stop removes the player from active rotation." Worth confirming which, and cleaning up accordingly.

2. mpris_service.cpp:559 and :1606 (removed) — Dropping playbackStatus = "Playing" reverts the no-signal-pause recovery heuristic.
Both removed lines corrected a player that falsely reports Paused while its position keeps advancing past kPauseRecoveryMinJumpUs (a known quirk on some players/compositors). With them gone, the model leaves such a player at Paused (only the m_recentNoSignalPauseAt erase remains). Since chooseActivePlayer()'s "most recent playing" path (line 2071) only considers playbackStatus == "Playing", an actually-playing-but-stuck-Paused player drops out of active selection — the exact bug that heuristic was added to fix. This may be intentional given the new active-follow logic, but it's a deliberate heuristic being reverted with no mention in the PR; the author should confirm intent.

3. mpris_service.cpp (new } else if (clearedPinnedPlayerPreference) tail of applyPlayerSnapshot) — Unreachable branch.
clearedPinnedPlayerPreference is only ever set inside the becamePlaying or transitionedFromPlaying paths, both of which require playbackStatus to differ between existing->second and info — which guarantees existing->second != info, so the outer if always wins. The else if (the existing->second == info case) can never run with clearedPinnedPlayerPreference == true. Harmless defensive code, but dead; can be dropped.

Design note (not a bug)

The new becamePlaying clear logic resets an explicit user pin (media next-player) as soon as any other player starts playing while the pinned one isn't playing. That matches the "active follows playback" intent, but it does mean an explicit pin to a paused player is silently overridden by background playback — worth a sentence in the PR/commit so it's a documented choice rather than a surprise.

Ape:
> 1. mpris_service.cpp:785 (removed) — Removing the sole dismissPlayer() call orphans the entire "dismissed players" subsystem.

Removed the dead code.

> 2. mpris_service.cpp:559 and :1606 (removed) — Dropping playbackStatus = "Playing" reverts the no-signal-pause recovery heuristic.

Not sure what to do about this one. I repeatly had false triggers caused by that code I removed. I haven't used any player myself that had the inverse issue that this might fix. I guess we could drop that commit from this PR and figure it out separately.

> 3. mpris_service.cpp (new } else if (clearedPinnedPlayerPreference) tail of applyPlayerSnapshot) — Unreachable branch.

Removed the dead code.

PR #3160, opened 2026-06-26
feat(brightness): add explicit backlight device override per monitor

ppenguin:
## Disclaimer

Entirely AI generated PR, currently lack the bandwidth to go deep, but it seems to work (see *Tested*).

## Summary

- Adds `backlight_device` config option to `[brightness.monitor.<name>]` sections, allowing users to pin a specific sysfs backlight device (e.g. `intel_backlight`) to a monitor
- When set, the brightness service skips all other backlight devices for that connector, preventing ambiguous multi-device selection
- Adds a new IPC command `brightness-list-backlight-devices` to enumerate available sysfs device names, making it easy to discover the right value to configure
- Documents the option in `example.toml` with a usage hint pointing to the discovery command

## Tested

Tested on NixOS apple-silicon (asahi base) on MBP M1, which doesn't work without this fix since it targets the touchbar for brightness.

Uses this (nix) config:
```nix
      programs = {
        noctalia.settings.brightness.monitor = {
          "eDP-1" = {
            backlight_device = "/sys/class/backlight/apple-panel-bl";
          };
          "touchbar" = {
            # dummy output, not actually reported by niri
            backlight_device = "/sys/class/backlight/228600000.dsi.0"; # touchbar
          };
     ....
```


ItsLemmy:
Thanks

PR #3171, opened 2026-06-27
docs(example): drop removed "centered" panel placement, document *_position

f-o-o-g-s:
hey dusty, tiny docs thing i bumped into while updating my install.

after 7c50bbf84 dropped the Centered placement (it's `floating` + a per-panel `*_position` now), `example.toml` still lists "centered" in the `[shell.panel]` comments and uses it as the launcher/clipboard default. so if you copy the example as-is, `config validate` grumbles:

```
WARN  shell.panel.launcher_placement: unknown value "centered"
WARN  shell.panel.clipboard_placement: unknown value "centered"
```

this just syncs the example with the real schema defaults (launcher/clipboard default to `floating` + `center`, the rest to `attached` + `auto`) and documents the new `*_position` keys. `center` reproduces the old centered behavior, so nothing changes for anyone copying it.

before/after with `noctalia config validate example.toml`:

```
before: 8 warnings (the two above + unrelated illustrative keys)
after:  6 warnings, placement/position lines clean
```

docs only, no code touched. no rush at all on merging, just did it for fun while i had the schema open. cheers

### DIFF
--- a/example.toml
+++ b/example.toml
@@ -43,14 +43,22 @@ direction = "down"                  # center, up, down, left, right, up_left, up
 alpha     = 0.55                    # multiplied by each component's background opacity
 
 [shell.panel]
-transparency_mode     = "solid"     # solid | soft | glass; controls floating/centered opacity and card translucency
+transparency_mode     = "solid"     # solid | soft | glass; controls floating opacity and card translucency
 borders               = true        # panel shell outline and in-panel section cards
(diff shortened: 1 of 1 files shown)


f-o-o-g-s:
ah, closing this one, i pushed it from the wrong repo/branch on my end. sorry for the ping. cheers

PR #3193, opened 2026-06-27
refactor(settings): improve search performance to avoid keyboard input lag

mattvilim:
## Summary

The shell settings panel lags when typing the first few letters because there are so many matches to build in the node tree. This PR makes two changes:

1) Limit number of matches to 50 total and show message at the end of search results explaining the limit (new i18n key).
2) Optimize visibility check to use O(1) lookup so `addSettingsContentSections` goes from O(N^2) to O(N) in registry size.

## Motivation

The lag at beginning of search is mildly annoying. I don't have anymore lag after this PR (on my CPU at least).

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [X] Refactoring
- [ ] Build / packaging

## Testing

I tested searching for settings keys that should be hidden and they don't appear. I also did the reverse.

## Manual Coverage

- [X] Tested on Niri
- [ ] Tested on Hyprland
- [ ] Tested on Sway

## Screenshots / Videos

Message when more than 50 results:

Screenshot_on_06-27-2026_at_16-07-56

## Checklist

### DIFF
--- a/src/shell/settings/settings_content.cpp
+++ b/src/shell/settings/settings_content.cpp
@@ -28,9 +28,11 @@
 #include <functional>
 #include <memory>
 #include <optional>
+#include <ranges>
 #include <string>
 #include <string_view>
 #include <type_traits>
+#include <unordered_map>
 #include <utility>
 #include <variant>
 #include <vector>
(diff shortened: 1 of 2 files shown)


src/shell/settings/settings_content.cpp
```diff
@@ -1278,40 +1280,41 @@ namespace settings {
...
       if (!e.visibleWhen.has_value()) {
         return true;
       }
-      for (const auto& cond : e.visibleWhen->all) {
-        if (!visibilityConditionMatches(cond)) {
-          return false;
-        }
-      }
-      return true;
+      return std::ranges::all_of(e.visibleWhen->all, visibilityConditionMatches);
```
mattvilim, line 1319: Logically equivalent...just cleanup with `std::ranges`.

src/shell/settings/settings_content.cpp
```diff
@@ -1278,40 +1280,41 @@ namespace settings {
...
+    constexpr std::size_t kMaxSearchResults = 50;
+    bool truncated = false;
     const std::string normalizedSearchQuery = normalizedSettingQuery(ctx.searchQuery);
 
     BarWidgetEditorContext barWidgetEditorCtx = makeBarWidgetEditorContext(factory);
 
+    // Visibility conditions reference other settings by path. Build an index once so each
+    // visibility check is an O(1) lookup instead of full registry scan.
+    const auto joinPath = [](const std::vector<std::string>& path) {
+      return path | std::views::join_with('.') | std::ranges::to<std::string>();
```
ItsLemmy: Should not join with '.'  as it would break if someone has a bar or monitor override containing a dot.

return path | std::views::join_with('\x1f') | std::ranges::to<std::string>();

mattvilim: Fixed and rebased :+1: (lint failure is on main unrelated to this PR)

ItsLemmy:
Cool, thanks!

PR #3197, opened 2026-06-28
fix(tray): dedupe equivalent status notifier items

### LINKED ISSUE #3194: Duplicate NordVPN tray icon from equivalent StatusNotifier registrations
Noctalia shows two NordVPN tray icons, but there is only one `norduserd` process.

Image

Versions:

```text
noctalia v5.0.0
commit 3b93fcff890da32b485e82ad72244b56d31cb326
NordVPN 5.1.0
Wayland/Niri
```

Process list:

```text
1380 root     .../nordvpnd
1750 bhunter  .../norduserd
1897 bhunter  noctalia
```

Watcher items:

```console
$ busctl --user get-property org.kde.StatusNotifierWatcher /StatusNotifierWatcher org.kde.StatusNotifierWatcher RegisteredStatusNotifierItems
as 3 ":1.3/StatusNotifierItem" ":1.42/org/blueman/sni" "org.kde.StatusNotifierItem-1750-1/StatusNotifierItem"
```

The two NordVPN entries seem to be the same item registered under both the unique bus name and the well-known `org.kde.StatusNotifierItem-*` name.

I think the issue is in `src/dbus/tray/tray_service.cpp`: `canonicalItemId()` uses `busName + objectPath`, so these become two separate tray items even though they point to the same object/process.

Expected: Noctalia should show one NordVPN icon.

Actual: Noctalia shows two NordVPN icons.

Possible fix: when registering an item, dedupe equivalent registrations that have the same object path and same DBus owner/process.

divyam234:
Fixes #3194.

NordVPN can register the same StatusNotifier item under both its unique DBus name and a well-known `org.kde.StatusNotifierItem-*` name. Noctalia currently keys tray items by `busName + objectPath`, so both registrations are rendered as separate icons.

This dedupes new tray registrations when an existing item has the same object path and the same DBus connection process ID.

Tested locally with NordVPN 5.1.0 on Wayland/Niri. The duplicate NordVPN icon is gone, and unrelated tray items still show.

### DIFF
--- a/src/dbus/tray/tray_service.cpp
+++ b/src/dbus/tray/tray_service.cpp
@@ -1775,6 +1775,24 @@ void TrayService::requestProcessNameForItem(const std::string& itemId, const std
   }
 }
 
+std::uint32_t TrayService::connectionPidForBusName(const std::string& busName) const {
+  if (busName.empty() || m_dbusProxy == nullptr || !looks_like_dbus_name(busName)) {
+    return 0;
+  }
+
+  try {
+    std::uint32_t pid = 0;
+    m_dbusProxy->callMethod("GetConnectionUnixProcessID")
+        .onInterface(kDbusInterface)
+        .withArguments(busName)
+        .storeResultsTo(pid);
+    return pid;
+  } catch (const sdbus::Error& e) {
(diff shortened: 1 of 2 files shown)


Ly-sec:
Hey @divyam234 , looks good to me thank you very much.

PR #3202, opened 2026-06-28
fix(core): reject directories in commandExists

### LINKED ISSUE #3167: [BUG] `process::commandExists` incorrectly returns true for directories
### Submission checklist

### Bug description

The core validation function [`process::commandExists`](https://github.com/noctalia-dev/noctalia/blob/d956d3ed46eacabe0196b050d5476f193d75061c/src/core/process.cpp#L710) incorrectly identifies existing directories as valid executable commands. The current implementation uses `::access(path, X_OK)`. `X_OK` returns a "success" status for directories if the user has search/enter permissions, causing the function to falsely validate a directory as an executable file.

### Steps to reproduce

There is no way of reproducing it easily on a freshly installed noctalia-shell. Personally, I discovered it when I was refining a plugin and it validated an incorrect but existent directory path as true, instead of returning false and giving an error. But it can be validated through the plugin API:

```lua
-- Correctly returns false
print(luau_commandExists("/not/a/real/path"))
```

```lua
-- Incorrectly returns true
print(luau_commandExists("/usr/bin"))
```

### Expected behavior

`process::commandExists` (and dependencies such as `luau_host::luau_commandExists`) should strictly verify if the path points to a regular executable file. If a directory path is given, it should evaluate to false.

### Actual behavior

The function evaluates any accessible directory as a valid executable command and returns true.

### Logs / error output

```text

```

### Noctalia version

v5.0.0

### Compositor

Niri

### Distribution

NixOS

### Environment information

```text
Not relevant.
```

### Additional context

_No response_

BUSTheid:
## Summary

`process::commandExists` uses `::access(path, X_OK)` which returns true for directories with search permission. Added `!std::filesystem::is_directory()` checks.

## Motivation

Fixes a validation bug where directory paths like `/usr/bin` or `/` are incorrectly reported as valid executables, affecting all callers, including the plugin API's `luau_commandExists`.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Refactoring
- [ ] Build / packaging

## Related Issue

Closes #3167

## Testing

`just test` with added unit tests
`just build` gives no errors

## Manual Coverage

- [ ] Tested on Niri
- [ ] Tested on Hyprland
- [ ] Tested on Sway
- [x] Not UI/visual change

## Checklist

### DIFF
--- a/src/core/process.cpp
+++ b/src/core/process.cpp
@@ -713,7 +713,7 @@ namespace process {
     }
 
     if (std::strchr(name, '/') != nullptr) {
-      return ::access(name, X_OK) == 0;
+      return ::access(name, X_OK) == 0 && std::filesystem::is_regular_file(name);
     }
 
     const char* pathEnv = std::getenv("PATH");
@@ -728,7 +728,7 @@ namespace process {
       const std::string_view dir = end == std::string_view::npos ? path.substr(start) : path.substr(start, end - start);
       const std::filesystem::path candidate =
           dir.empty() ? std::filesystem::path(name) : (std::filesystem::path(dir) / name);
-      if (::access(candidate.c_str(), X_OK) == 0) {
+      if (::access(candidate.c_str(), X_OK) == 0 && std::filesystem::is_regular_file(candidate)) {
         return true;
       }
       if (end == std::string_view::npos) {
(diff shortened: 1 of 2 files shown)


NousadS:
Wouldn't it make more sense to use `std::filesystem::is_regular_file` instead of `!std::filesystem::is_directory` here? The function follows symlinks so the symlink case is also covered.

PR #3219, opened 2026-06-29
Add clipboard color previews

JohnnyP0T:
## Summary

This PR adds color-aware previews to the clipboard panel.

Text clipboard entries that look like color values now show a swatch in the list and in the preview area instead of only a generic text icon. It also makes text preview truncation safer by avoiding cutting UTF-8 sequences in the middle.

## Included changes

- detect color-like clipboard text values
- support hex, `rgb(...)`, `rgba(...)`, `hsl(...)`, and `hsla(...)` parsing
- show a color swatch in clipboard list rows for parsed colors
- show a larger color preview in the clipboard preview pane
- keep normal image thumbnail behavior unchanged
- make `ClipboardService::buildTextPreview()` stop at a complete UTF-8 sequence within the preview byte limit

## Validation

- ran `clang-format` on the touched files
- built successfully with `ninja -C build-debug`

### DIFF
--- a/src/render/core/color.cpp
+++ b/src/render/core/color.cpp
@@ -2,9 +2,12 @@
 
 #include <algorithm>
 #include <cctype>
+#include <charconv>
 #include <cmath>
 #include <cstdio>
+#include <numbers>
 #include <string>
+#include <system_error>
 
 namespace {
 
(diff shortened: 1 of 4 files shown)


ItsLemmy:
Thanks for the PR. Totally forgot that we had something similar in v4.

I refactored the color parsing in the appropriate class and added support "hsl(74 64.3% 59.4%)" and "rgb(255 0 0)" (no comma)

PR #3222, opened 2026-06-29
feat(build): Add system_tomlplusplus option

carlwgeorge:
## Summary

Add system_tomlplusplus option to allow building against the system tomlplusplus library instead of the vendored one.  The option is disabled by default to retain the current vendored behavior.

## Motivation

* [Fedora packaging guidelines](https://docs.fedoraproject.org/en-US/packaging-guidelines/#bundling) 
* [Debian policy manual](https://www.debian.org/doc/debian-policy/ch-source.html#embedded-code-copies)
* [openSUSE bundled software policy](https://en.opensuse.org/openSUSE:Bundled_software_policy)

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [x] Build / packaging

## Testing

I'm using a local RPM package built with this change, linked against `libtomlplusplus.so.3` (`tomlplusplus-3.4.0-7.fc44`) in Fedora.  `meson test` passed during the build, and as far as I can tell everything works the same as it does without the patch.

## Manual Coverage

- [x] Tested on Niri
- [ ] Tested on Hyprland
- [ ] Tested on Sway

## Checklist


ItsLemmy:
Thanks!

PR #3227, opened 2026-06-30
feat(wifi): Add IWD support 

rXelelo:


ItsLemmy:
1. src/dbus/network/iwd_service.cpp (strengthFromValue / signalToPercent, ~line 57) — Wi-Fi signal strength is computed from the wrong unit, so every network reports 0%.
Station.GetOrderedNetworks() returns the signal strength in 100·dBm (e.g. −54 dBm arrives as -5400), not raw dBm. strengthFromValue falls through to signalToPercent(value), which treats -5400 as dBm: value <= -100 → returns 0. Result: a RelativeStrength of 0% for all visible APs and 0% for the connected network's signalStrength. The fix is to scale: signalToPercent(value / 100). (The value >= 0 && value <= 100 branch is also dead — iwd never returns a 0–100 value here — and masks the unit confusion.)

2. src/dbus/network/iwd_service.cpp (activateAccessPoint(ap, psk), ~line 314) — the Wi-Fi passphrase is passed as an iwctl --passphrase <psk> command-line argument, leaking it via /proc.
For the lifetime of the connect attempt (up to the 20s timeout), the cleartext PSK is visible in /proc/<pid>/cmdline to any local user. The wpa backend avoids this by passing the secret over D-Bus (AddNetwork args). The clean iwd equivalent is registering an iwd agent (net.connman.iwd.Agent / AgentManager) so Network.Connect can request the passphrase over the bus, rather than shelling out to iwctl.

3. (minor, cleanup) signalToPercent is duplicated verbatim from wpa_supplicant_service.cpp. Both files define the identical dBm→percent helper in an anonymous namespace. Once finding 1. is fixed it diverges (iwd needs the /100 scaling), so the cleanest form is a shared dBmToPercent helper in a network util header that iwd calls after dividing by 100 — keeps the conversion in one place.

ItsLemmy:
1. src/dbus/network/iwd_service.cpp:329 — connecting to a new secured network is now non-functional on iwd.
By removing the iwctl path, activateAccessPoint(ap, psk) discards the passphrase and issues Network.Connect. For a network iwd doesn't already know, Connect needs a registered agent to obtain the passphrase; with none registered it returns an error → caught → activateAccessPoint returns false. So a user on iwd who selects a secured SSID and types the password just gets a failed connection (only already-saved/known networks connect). This is a deliberate trade-off (correctly chosen over the /proc leak), but it leaves the headline Wi-Fi use case broken on iwd. The proper fix is registering a net.connman.iwd.Agent via AgentManager.RegisterAgent — analogous to the existing NetworkSecretAgent for NetworkManager — so the passphrase travels over D-Bus. Worth at least a tracking note; ideally a follow-up before this is the only backend a user has.

ItsLemmy:
We should not use "callMethod" (synchronous) for long running dbus calls, or it will freeze the entire shell.

Untested code below...

Header, add a member to hold the in-flight connect:
```
std::unique_ptr<sdbus::IProxy> m_connectProxy; // keeps an in-flight async Connect alive
```
activateAccessPoint(ap) must  go async:
```
bool IwdService::activateAccessPoint(const AccessPointInfo& ap) {
  try {
    // Starting a new connect supersedes any previous in-flight one.
    m_connectProxy = sdbus::createProxy(m_bus.connection(), kIwdBusName, sdbus::ObjectPath{ap.path});
    m_connectProxy->callMethodAsync("Connect")
        .onInterface(kNetworkInterface)
        .uponReplyInvoke([this, ssid = ap.ssid](std::optional<sdbus::Error> error) {
          m_pendingPsk.clear();            // consumed by the agent, or abandoned, never keep it around
          if (error) {
            kLog.warn("Connect failed for {}: {}", ssid, error->what());
          }
          refresh();
        });
    return true;                            // dispatched; real result arrives via the change callback
  } catch (const sdbus::Error& e) {
    m_pendingPsk.clear();
    kLog.warn("Connect dispatch failed for {}: {}", ap.ssid, e.what());
    return false;
  }
}
```

ItsLemmy:
almost there tho, dont give up!

ItsLemmy:
- iwd_secret_agent.cpp:47  request.ssid is set to the network object path, and the comment says it "Extract[s] SSID from the network path" but doesn't. Cosmetic now (callback ignores it). Either resolve it to the real SSID or rename to networkPath.

- application_services.cpp:935  the dynamic_cast<IwdService*> + setSecretAgent special-case bypasses the generic supportsSecretAgent() mechanism. Works, but a backend-specific RTTI branch in shared init. Cleaner long-term to let IwdService own its agent.

- Minor, signalToPercent still duplicated with wpa_supplicant_service.cpp.

thoastbrot:
It wasn't complete yet - I used claude to fix the remaining problems for me. It was a very small patch, attached. Not sure if this is valid for everyone - it fixed it for me.
[iwd-ordered-networks.patch](https://github.com/user-attachments/files/29659545/iwd-ordered-networks.patch)

PR #3229, opened 2026-07-01
feat(nix): add CUDA package to Nix flake for Cachix

LucasOe:
## Summary

Adds a `cuda` package to the flake so it can be built and cached by Cachix.

## Motivation

GPU monitoring on NVIDIA GPUs requires enabling CUDA by overriding `cudaSupport = true`. This changes the package derivation, resulting in a cache miss and forcing users to build the package locally.

This PR exposes a dedicated `cuda` package and builds it in the `cachix/cachix-action` workflow, allowing users with `cudaSupport = true` to benefit from the Cachix cache.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [x] Build / packaging

## Testing

I successfully built both the `default` and `cuda` packages locally. I was not able to test the GitHub Actions workflow.

## Manual Coverage

- [x] Tested on Niri
- [ ] Tested on Hyprland
- [ ] Tested on Sway

## Checklist

### DIFF
--- a/.github/workflows/cachix.yml
+++ b/.github/workflows/cachix.yml
@@ -26,7 +26,7 @@ jobs:
         with:
           name: "${{ secrets.CACHIX_CACHE_NAME }}"
           authToken: "${{ secrets.CACHIX_AUTH_TOKEN }}"
-      - run: nix build . --fallback
+      - run: nix build . .#cuda --fallback
 
   push:
     needs: cache
--- a/.github/workflows/cachix.yml
+++ b/.github/workflows/cachix.yml
@@ -13,10 +13,11 @@ on:
   workflow_dispatch:
 
 jobs:
-  cache-default:
+  cache:
     strategy:
       matrix:
         runner: [ubuntu-latest, ubuntu-24.04-arm]
+        target: [default, cuda]
     runs-on: ${{ matrix.runner }}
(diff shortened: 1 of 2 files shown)


LucasOe:
@linusammon

linusammon (changes requested):
Wouldn't this double the time it takes for a workflow to finish/push stuff to Cachix? Perhaps these could be separated into their own runs with another matrix?

linusammon:
Another potential issue is Cachix storage. Right now, cached packages are garbage collected after about a week if they aren't accessed regularly, and this would pretty much cut that time in half.

That said, I'm happy to give it a try and see if it actually becomes a problem.

LucasOe:
I split the package builds into two separate jobs so they can run in parallel.

One thing I'm unsure about is whether it makes sense to keep the `ubuntu-24.04-arm` runner for the CUDA package. My understanding is that CUDA support in Nixpkgs is primarily targeted at `x86_64-linux`, so the ARM build may not provide much value. I couldn't find anything in the NixOS wiki confirming or refuting that, though, so I left it as-is for now. If the ARM job isn't useful here, we could remove it.

linusammon (approved):
LGTM, thanks :)

PR #3233, opened 2026-07-01
pipewire: use sink/source node descriptions for audio device labels

Ape:
The current version was just overly complex and didn't use the correct names. Pipewire already provides human readable names.

### DIFF
--- a/src/pipewire/pipewire_service.cpp
+++ b/src/pipewire/pipewire_service.cpp
@@ -1235,8 +1235,8 @@ void PipeWireService::onNodeParam(
 
   auto& nd = *it->second;
   // Device nodes get their volume/mute authoritatively from mixer-api (onMixerVolumeChanged); their
-  // SPA_PARAM_Props volume/mute echoes are ignored. Route metadata (availability, description, route
-  // mute) is still tracked for device selection and effective mute.
+  // SPA_PARAM_Props volume/mute echoes are ignored. Route availability and mute are still tracked
+  // for device selection and effective mute.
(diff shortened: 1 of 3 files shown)


ItsLemmy:
I'm still working in this area so I cannot merge. Also this brings back a major pain point for me on my laptop where I have 4  Sinks named  very similarily.

```
 │      51. Alder Lake PCH-P High Definition Audio Controller HDMI / DisplayPort 3 Output [vol: 0.39 MUTED]
 │      52. Alder Lake PCH-P High Definition Audio Controller HDMI / DisplayPort 2 Output [vol: 0.39]
 │      53. Alder Lake PCH-P High Definition Audio Controller HDMI / DisplayPort 1 Output [vol: 0.39]
 │  *   54. Alder Lake PCH-P High Definition Audio Controller Speaker [vol: 0.27]
 ```
 
 So in noctalia all I see is: "Alder lake PCH-P..."   And I'm unable to distinguish what is the speaker vs HDMI.
 I'll see if left eliding is a good option

Ape:
> I have 4 Sinks named very similarily.

Is this also how they are shown in other apps like pwvucontrol? Perhaps you might want a local wireplumber rename rule.

Also, don't worry about merging this if it's not useful right now.

ItsLemmy:
went with middle elidding. which is not pretty but helps.
image

On the laptop I renamed the devices, so it's cool now.

Debating if we really need Output:   Input:  since we already have a Icon that convey what they are pretty well.

PR #3249, opened 2026-07-03
feat(theming): add custom time for day/night scheduling

SlowNicoFish:
## Summary

Adds enableCustomScheduling to [location], letting users explicitly opt into manual sunrise/sunset scheduling for auto theme mode and night light. Fixed times are now an intentional opt-in rather than a silent fallback when coordinates are unavailable.

Sunset/sunrise settings are moved from the Location section to the Appearance section of the settings UI, where they're visible when theme mode is auto and enableCustomScheduling is on.

## Motivation

Fixed times was previously a fallback for only when location resolution failed, making intentional use unpredictable. This makes manual scheduling easier for the user.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Refactoring
- [ ] Build / packaging

## Testing

Built with just build
Tested explicitly running just test 
Verified enable_custom_scheduling = true switches theme/night light at configured times
Verified default (false) still uses solar calculation
Verified settings entries show/hide correctly in the UI
Verified warn log fires when times are set but use_fixed_times is off

## Manual Coverage

- [x] Tested on Niri
- [ ] Tested on Hyprland
- [ ] Tested on Sway

## Screenshots / Videos

Screenshot from 2026-07-06 23-58-27

## Checklist

## Additional Notes

isManualMode no longer takes resolved coordinates. Manual mode is now purely config-driven. Existing configs without use_fixed_times default to false, so behavior is backward-compatible.


ItsLemmy:
Thanks for the PR.

I like that we move from a silent fallback to an explicit opt-in.
That being said :

- We don't want to move sunset /sunrise in Appearance, this should stay in Location which is the central place for all stuff location based.  This will also fix the regression you introduced where the  manual sunset/sunrise time wont be accessible (for gamma)  if you did not pick "Auto"  which most people don't.

- I would go with "enable_custom_scheduling" rather than "use_fixed_times"  for consistency with the code base.

The rest looks okay.

Stricken1670:
Might I suggest using the mechanism of https://sr.ht/~kennylevinsen/wlsunset/ ? This causes the transition between day and night to be gradual, instead of instant.

ItsLemmy:
Thanks

PR #3253, opened 2026-07-03
fix(toplevels): prevent null toplevel output when spanning monitors

turtlestarship:
## Summary

Track all outputs a toplevel is visible on instead of only the most recently entered one. When a leave event fires, fall back to the latest visible output rather than unconditionally setting `output` to `nullptr`. This prevents `output` becoming null whilst the toplevel is still visible on previously entered outputs.

## Motivation
When using a bar on each monitor, both with a taskbar, and with "Show All Monitors" disabled. When I drag a window to span two monitors, and then drag it back to to its original monitor without having fully left the original monitor. The app icon disappears from both taskbars. 

1. Window opens in Monitor A
    - `output_enter` event and `output = A`
2. Window is dragged so that it partially enters monitor B 
    - `output_enter` event and `output = B` (but no `output_leave` has fired for A)
3. Window is moved back to Monitor A without having fully left it
    - `output_leave` event for monitor B `output = nullptr` 

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Build / packaging

## Testing

- `just format`
- `just test`
- `just build`
- Dragged a window to span two monitors, dragged it back, an icon remained visible in a taskbar throughout. Before this fix the icon disappeared on drag back. 

## Manual Coverage

- [ ] Tested on Niri
- [ ] Tested on Hyprland
- [ ] Tested on Sway
- [x] Tested on another compositor: labwc

## Screenshots / Videos

N/A

## Checklist

## Additional Notes

The fix was intentionally narrow in scope. No attempt was made to expose multi-output toplevel state more broadly to the rest of the codebase. e.g. through modifying `WlrToplevelSnapshot`, `allAppIds` or other places output is used.

The behaviour of updating `output` on `output_enter` rather than on `output_leave` was also preserved. An alternative specifically related to the taskbar would be to only update `output` when the toplevel leaves its oldest visible output rather than update `output` to its newest visible output.

### DIFF
--- a/src/wayland/wayland_toplevels.cpp
+++ b/src/wayland/wayland_toplevels.cpp
@@ -361,6 +361,9 @@ void WaylandToplevels::closeHandle(zwlr_foreign_toplevel_handle_v1* handle) {
 void WaylandToplevels::onHandleOutputEnter(zwlr_foreign_toplevel_handle_v1* handle, wl_output* output) {
   auto it = m_handles.find(handle);
   if (it != m_handles.end()) {
+    if (!std::ranges::contains(it->second.activeOutputs, output)) {
+      it->second.activeOutputs.push_back(output);
+    }
     if (it->second.output != output) {
       it->second.output = output;
       it->second.dirty = true;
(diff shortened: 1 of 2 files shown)


Ly-sec:
Hey @turtlestarship , looks good to me. I'll merge it.

PR #3255, opened 2026-07-03
fix(lock-surface): do not clip the status line

CLEMENTINATOR:


enklht:
Thanks for the fix!

I noticed one small follow-up issue.
The panel height doesn't seem to account for the hint band, so the vertical margins become quite small.
Would it make sense to increase `panelY` so that the panel gets `smallSm` padding as intended?

The only solution I came up with was adding another argument to `panelContentLayout` to account for the hint band's height, but that feels like a wrong absrtaction.

PR #3259, opened 2026-07-03
fix(calendar): expand CalDAV recurring events client-side when server ignores <C:expand>

floydya:
## Summary

Recurring events only showed up on their start date instead of repeating.

Noctalia's calendar service never expanded recurrence -  it relies on the server (Google's `singleEvents=true`, CalDAV's `<C:expand>`). If a CalDAV server ignores `<C:expand>`, it hands back the master `VEVENT` with its `RRULE` and `ical_parser` turned that into a single instance on `DTSTART`.

The fix: if a `VEVENT` still has a `RRULE` after fetch - expand it, bounded to the fetch window. Server that already expand are unaffected (no `RRULE` left to act on), and Google doesn't touch this path. The fix handles `FREQ`, `INTERVAL`, `COUNT`, `UNTIL`, weekly `BYDAY` and `EXDATE`.

## Motivation

On self-hosted CalDAV servers without `<C:expand>` support, a daily/weekly event shows up exactly once, which makes recurring events basically useless.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Build / packaging

## Related Issue

No related issue was created before.

## Testing

Added `tests/ical_recurrence_test.cpp` (wired into `meson.build`) with cases for daily/weekly/monthly recurrence, `INTERVAL`, `COUNT`, `UNTIL`, `EXDATE`, a series that starts before the window, and an old unbounded daily one that still needs to reach today. All pass.

Manually switched noctalia flake from github source to local, built and tested the results - recurring events are shown correctly(added before/after screenshots below).

## Manual Coverage

- [x] Tested on Niri
- [ ] Tested on Hyprland
- [ ] Tested on Sway

## Screenshots / Videos

Before:
image

After:
image

Using the same calendar config, "after" events are all valid, so there are no "ghost" events.

## Checklist

## Additional Notes

Occurrences repeat at the same UTC instant, so times can be off by an hour across a DST change. Doesn't handle nth-of-month `BYDAY` (`2MO`), `BYMONTHDAY`/`BYMONTH`/`BYSETPOS`, or `WKST` - none of my calendars use them, easy to add later if needed(decided to skip things I can't easily verify to be sure I deliver a completely working commit).

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1108,6 +1108,16 @@ string_utils_test = executable('string_utils_test',
 
 test('string_utils', string_utils_test)
 
+ical_recurrence_test = executable('ical_recurrence_test',
+  sources: files(
+    'tests/ical_recurrence_test.cpp',
+    'src/calendar/ical_parser.cpp',
+  ),
+  include_directories: _noctalia_inc,
+)
+
+test('ical_recurrence', ical_recurrence_test)
+
 time_format_test = executable('time_format_test',
   sources: files(
     'tests/time_format_test.cpp',
--- a/src/calendar/caldav_client.cpp
+++ b/src/calendar/caldav_client.cpp
@@ -73,7 +73,7 @@ namespace calendar {
 
     const std::string calendarName = account.calendarName;
     const std::string color = account.color;
-    http.request(std::move(req), [cb = std::move(cb), calendarName, color](HttpResponse resp) {
+    http.request(std::move(req), [cb = std::move(cb), calendarName, color, start, end](HttpResponse resp) {
       if (!resp.transportOk || (resp.status != 207 && resp.status != 200)) {
         kLog.warn("caldav REPORT failed http={}", resp.status);
         cb(false, {});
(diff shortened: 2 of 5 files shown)


ItsLemmy:
Hey thanks for the PR.

Here is the roadblock :

src/calendar/ical_parser.cpp:~257,320,333
all-day (VALUE=DATE) recurrences land on the wrong day for timezones ahead of UTC.
parseDateTime stores an all-day event's local midnight converted to UTC, then expandRecurrence derives startDay/day-of-month from floor(base.start) (a UTC day). For a UTC+ zone (e.g. Tokyo), local 2024-01-01 00:00 = 2023-12-31 15:00Z, so startDay is Dec 31. A VALUE=DATE:20240101 + FREQ=MONTHLY then recurs on the 31st of each month (skipping Feb/Apr/Jun/Sep/Nov) instead of the 1st; FREQ=YEARLY recurs on Dec 31 2023; weekly BYDAY=MO shifts to the wrong weekday. Wrong dates and wrong counts for any east-of-UTC user with an all-day recurring event on a non-expanding server. (Weekly without BYDAY self-corrects.) Fix: derive the civil date for all-day events from the local date, not floor of the UTC instant.

floydya:
Hey, thanks for the review.

Fixed the bug you mentioned and added a regression test.

ItsLemmy:
Thanks! I'll fix the test that fails here.

PR #3264, opened 2026-07-04
fix(clipboard): fix insufficient height for list items

enklht:
## Summary

Increase the height of clipboard history items to provide balanced vertical padding around the contents.

## Motivation

The clipboard history items have noticeably less bottom padding than top padding, making them look unbalanced (see the screenshots below).
The cause is that the current height is too small.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Build / packaging

## Related Issue

None.

## Testing

Ran `just test` and `just build`.
Confirmed there were no errors.

## Manual Coverage

- [x] Tested on Niri
- [ ] Tested on Hyprland
- [ ] Tested on Sway

## Screenshots / Videos

Before:
Screenshot from 2026-07-04 16-26-19

After:
image

## Checklist

## Additional Notes

I think the ideal solution might be to fit the item height to its contents rather than using a fixed value, as that would keep the layout consistent if the contents change in the future. 
However, since the height is currently hardcoded, this PR simply increases it.

The new height (56px) was chosen because it is the smallest multiple of 8 that fully accommodates the current contents. While that value is somewhat arbitrary, it follows the existing spacing convention.

### DIFF
--- a/src/shell/clipboard/clipboard_panel.cpp
+++ b/src/shell/clipboard/clipboard_panel.cpp
@@ -32,7 +32,7 @@
 
 namespace {
 
-  constexpr float kRowHeight = 46.0f;
+  constexpr float kRowHeight = 56.0f;
   constexpr float kPreviewImageHeight = 280.0f;
   constexpr float kListGlyphSize = 24.0f;
   constexpr float kListThumbSize = 40.0f;


ItsLemmy:
Thanks for the PR, Unfortunately this is not a reliable fix, since this issue only happens with fonts that have a weird line height.  So we can't rely on a hardcoded row height, it would fix it for you and break it for everyone else.

I came up with a dynamic solution that should solve it in 4f9e8343

enklht:
Thanks for the explanation and the fix. That makes sense. I suspect I'm seeing this issue because of my CJK font.

~I also noticed similar confliction between hardcoded sizes and font sizes on the lock screen, where the hint message overflows.
From the quick look at the code, the lock screen panel, passwordinput, confirmation button, hint message, and the surrounding frame are all siblings rather than having a natural hierarchy and the sizes are computed manually, so the fix might be a bit more involved. Do you mind if I open an issue for that?~
(edit: The overflow issue was already fixed by a pr merged yesterday. Sorry for the oversight.)

Also, thanks for this amazing project! I'm really impressed.

PR #3272, opened 2026-07-04
feat(build): Add system_nlohmann option

NickHastings:
## Summary
This allows building against the system nlohmann json library instead of the vendored one. The option is disabled by default to retain the current vendored behavior.

## Motivation

Debian has a very strong stance on this. From the [Debian policy manual](https://www.debian.org/doc/debian-policy/ch-source.html#embedded-code-copies):

> Some software packages include in their distribution convenience copies of code from other software packages, generally so that users compiling from source don’t have to download multiple packages. Debian packages should not make use of these convenience copies unless the included package is explicitly intended to be used in this way.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [x] Build / packaging

## Testing

On Debian 13 I tested this branch using builds with the system `nlohmann-json3-dev` (`meson setup .. -Dsystem_nlohmann=true`) and also with the vendored `third_party/nlohmann`.

## Manual Coverage

- [x] Tested on Niri
- [ ] Tested on Hyprland
- [ ] Tested on Sway

## Checklist


meson_options.txt
```diff
@@ -2,3 +2,4 @@ option('jemalloc', type: 'feature', value: 'auto', description: 'Use jemalloc on
 option('tests', type: 'feature', value: 'auto', description: 'Build unit tests (auto: on for unsanitized debug builds)')
 option('system_md4c', type: 'boolean', value: false, description: 'Use system md4c instead of vendored')
 option('system_tomlplusplus', type: 'boolean', value: false, description: 'Use system tomlplusplus instead of vendored')
+option('system_nlohmann', type: 'boolean', value: false, description: 'Use system nlohmann instead of vendored')
```
carlwgeorge: I think this flag should be `system_nlohmann_json`.  json is just one library from that developer, and things could get confusing if this project ever starts using another one of their libraries.

meson.build
```diff
@@ -187,6 +187,15 @@ else
   )
 endif
 
+# ── Vendored: nlohmann (header-only) ─────────────────────────────────
+if get_option('system_nlohmann')
+  nlohmann_dep = dependency('nlohmann_json')
+else
+  nlohmann_dep = declare_dependency(
+    include_directories: include_directories('third_party/nlohmann', is_system: true),
```
carlwgeorge, line 195: Similar to the flag name, it's probably best if this directory structure allows for the possible inclusion of other nlohmann libraries in the future, something like:

```
third_party/
    nlohmann_json/
        nlohmann/
            json.hpp
```

carlwgeorge:
Aside from the nits on the flag name and directory structure, I can confirm that with this patch I can build against Fedora's json-devel package.

meson_options.txt
NickHastings: Done.

meson.build
NickHastings: I agree, but think it may be the something for a separate PR.

NickHastings: Will be happy to add this either here, or in a separate PR after this is merged. Will wait for feedback from core devs.

PR #3275, opened 2026-07-05
feat(build): Add system_stb option

NickHastings:
## Summary
This allows building against the system stb library instead of the vendored one. The option is disabled by default to retain the existing behavior.

## Motivation

Debian has a very strong stance on this. From the [Debian policy manual](https://www.debian.org/doc/debian-policy/ch-source.html#embedded-code-copies):

> Some software packages include in their distribution convenience copies of code from other software packages, generally so that users compiling from source dont have to download multiple packages. Debian packages should not make use of these convenience copies unless the included package is explicitly intended to be used in this way.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [x] Build / packaging

## Testing
On Debian 13 I tested this branch using builds with the system std (`meson setup .. -Dsystem_std=true`) and also with the vendored `third_party/std`.

Note that `include_type: 'system'` was used to suppress some warnings from implicit conversions between integer types in both `stb_image_resize2.h` and `stb_image_write.h`.

## Manual Coverage

- [x] Tested on Niri
- [ ] Tested on Hyprland
- [ ] Tested on Sway

## Checklist


meson.build
```diff
@@ -174,9 +174,13 @@ mcu_dep = declare_dependency(
 )
 
 # ── Vendored: stb_image_resize2 (header-only) ─────────────────────────────────
-stb_dep = declare_dependency(
-  include_directories: include_directories('third_party/stb', is_system: true),
-)
+if get_option('system_stb')
+  stb_dep = dependency('stb', include_type: 'system')
```
carlwgeorge: `dependency` relies on either pkgconfig or cmake files for detection.  Debian has a pkgconfig file, but it's the result of a downstream patch.  Fedora and Arch at least don't have that pkgconfig file, and probably other distros don't either.  Those distros can still keep using the bundled one, but ideally the `system_stb` flag shouldn't be only useful on Debian.

carlwgeorge: Here's an alternate approach that should be more cross-distro compatible.

```suggestion
  stb_dep = declare_dependency(
    include_directories: include_directories('/usr/include/stb', is_system: true),
  )
```

carlwgeorge (changes requested):

meson_options.txt
```diff
@@ -2,3 +2,4 @@ option('jemalloc', type: 'feature', value: 'auto', description: 'Use jemalloc on
 option('tests', type: 'feature', value: 'auto', description: 'Build unit tests (auto: on for unsanitized debug builds)')
 option('system_md4c', type: 'boolean', value: false, description: 'Use system md4c instead of vendored')
 option('system_tomlplusplus', type: 'boolean', value: false, description: 'Use system tomlplusplus instead of vendored')
+option('system_stb', type: 'boolean', value: false, description: 'Use system tomlplusplus instead of vendored')
```
carlwgeorge:
```suggestion
option('system_stb', type: 'boolean', value: false, description: 'Use system stb instead of vendored')
```

carlwgeorge (changes requested):

meson.build
NickHastings: Yeah, since only the headers are needed, that would work.

I wonder if in that case, the actual location headers should be the variable, instead of a bool. By default it could be set to `third_party/stb`, then it could be overridden to be with something like:
```
meson setup .. -Dstb_headers=/usr/include/stb
```

meson_options.txt
```diff
@@ -2,3 +2,4 @@ option('jemalloc', type: 'feature', value: 'auto', description: 'Use jemalloc on
 option('tests', type: 'feature', value: 'auto', description: 'Build unit tests (auto: on for unsanitized debug builds)')
 option('system_md4c', type: 'boolean', value: false, description: 'Use system md4c instead of vendored')
 option('system_tomlplusplus', type: 'boolean', value: false, description: 'Use system tomlplusplus instead of vendored')
+option('system_stb', type: 'boolean', value: false, description: 'Use system stb instead of vendored')
```
NickHastings:
```suggestion
option('stb_headers', type: 'string', value: 'third_party/stb', description: 'Path to stb headers')
```

NickHastings: Then the entry for stb in `meson.build` would be
```
stb_dep = declare_dependency(
  include_directories: include_directories(get_option('stb_headers'), is_system: true),
)
```

meson_options.txt
NickHastings: Done

ItsLemmy:
Thanks, will clean up the README a bit to make it more obvious that stb_headers expect a path and is different from the system_xxx ones

PR #3279, opened 2026-07-05
fix(pw): simplify relative volume adjustment to use flat steps

### LINKED ISSUE #3265: [BUG] volume control is jumpy and uneven
### Submission checklist

### Bug description

instead of the volume going by steps of two, which was what i had set, it gets jumpy, sometimes going from 4-15, then 16, then 18, then 22, then 30, and so on so forth

### Steps to reproduce

use media controls to adjust volume and it should happen

### Expected behavior

it should increase by steps of 2 and be even and smooth

### Actual behavior

it gets jumpy and increases at odd increments, sometimes skipping large sets of volume

### Logs / error output

```text

```

### Noctalia version

5.0.0.r4171.g4aa9f436d-1

### Compositor

Hyprland

### Distribution

Arch-based

### Environment information

```text
installation method: AUR
GPU: intel integrated gpu
```

### Additional context

https://github.com/user-attachments/assets/af682a53-b5aa-475d-8d19-d7ad3f4cb1b1

TotalyEnglizLitrate:
## Summary
Simplifies the relative volume adjustment logic by restoring flat step increments (e.g., exactly 5% per event tick) for keybind-driven adjustments. This resolves the issue where volume steps are uneven when media keys are held down (due to attempted time interval based smoothing).

## Motivation

The previous time-based acceleration logic integrated over the wall-clock interval (`dt`) between consecutive IPC calls to calculate the step size. However, since compositors (usually) handle key repeats by spawning asynchronous client processes (`noctalia msg volume-up`), scheduling overhead and process-start latency introduced variable arrival jitter (`dt` fluctuations) at the daemon, making the volume step sizes feel highly uneven and non-granular.

Looking at the history of the method, flat steps were originally dropped due to stale echoes in the read-back volume (`sink->volume`) causing lost updates and rubber-banding when commands were sent in rapid succession. Because of the introduction of the gesture-local accumulator (`m_relativeAdjust.target`) which tracks the target volume independently of the lagging read-back volume, we can safely move back to a flat step without re-introducing the lost update and rubber-banding issues.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Build / packaging

## Related Issue

Closes: #3265 

## Testing

verified behaviour against original faulty flat step impl, current impl, and proposed fix with the following (fish) shell script
```sh
for i in (seq 1 30) # default vol step 5, with overdrive 150/5
    noctalia msg volume-up &
end
```

## Manual Coverage

- [x] Tested on Niri
- [ ] Tested on Hyprland
- [ ] Tested on Sway

## Screenshots / Videos

https://github.com/user-attachments/assets/fea4e534-9ce9-43cd-9f48-61c67a11561e

## Checklist

## Additional Notes

I do not have a knob to control volume with, and I believe this fix carries the intended behaviour w.r.t knob based ipc invocations but I'm not 100% sure.

### DIFF
--- a/src/pipewire/pipewire_service.cpp
+++ b/src/pipewire/pipewire_service.cpp
@@ -1816,29 +1816,18 @@ float PipeWireService::relativeAdjustTarget(
 ) {
   const auto now = std::chrono::steady_clock::now();
   const bool held = m_relativeAdjust.gesture == gesture && (now - m_relativeAdjust.lastAt) <= kVolumeHoldWindow;
-  const float dt =
-      held ? std::min(std::chrono::duration<float>(now - m_relativeAdjust.lastAt).count(), kVolumeHoldMaxDt) : 0.0f;
   m_relativeAdjust.gesture = gesture;
   m_relativeAdjust.lastAt = now;
 
   if (!held) {
     // Isolated tap or new gesture: a fixed, granular step from the live volume.
-    m_relativeAdjust.heldSeconds = 0.0f;
     m_relativeAdjust.target = std::clamp(current + direction * baseStep, 0.0f, maxVolume);
     return m_relativeAdjust.target;
   }
 
-  // Held: advance the gesture-local target by velocity * capped event-time credit. Accumulating the
-  // target across the gesture keeps stale daemon echoes in the read-back volume from rubber-banding
-  // the ramp, and capping dt makes an arrival gap (keyboard repeat delay, IPC spawn jitter) worth
-  // about one base step instead of a jump. The velocity ramp advances by the same capped credit, so
-  // acceleration follows the event flow rather than the wall clock. The per-event cap of one base
-  // step keeps burst sources (rotary knobs) proportional to their tick count.
-  const float baseVel = baseStep / kVolumeHoldMaxDt;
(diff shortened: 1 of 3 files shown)


TotalyEnglizLitrate:
I guess clang-format modified the settings file? (didn't notice until now) lmk if I should drop it or let it be

ItsLemmy:
Unfortunately this PR brings back the main pain point for me:
On my thinkpad when starting at low volume (ex: 10%) and holding the media/volume up key (until I reach 80% on the OSD) the volume you hear is like 2.5s late...  so it keeps ramping up 2.5s after you release the key.

TotalyEnglizLitrate:
leading theory is exec (compositor side) and ipc call overhead, I presume this doesn't happen with scrolling over the bar/desktop widget?

TotalyEnglizLitrate:
bit of a shot in the dark, added an 80ms rate limiter to mirror the current implementation as closely as possible while still sticking to the step value, lmk if this makes the lag any better. it does make the volume moving up noticeably slower on my laptop but it's not _that_ much slower, nothing I can't live with; and I only noticed it because I was A/B testing, so

ItsLemmy:
Thanks, works well for me.
I'll go ahead and merge and then lower the rate limiter to 50ms as it still works perfect for me and make things less sluggish.

PR #3288, opened 2026-07-05
feat(theme): add `color_dark` and `color_light` options for custom colors

### LINKED ISSUE #3262: [FEATURE] Allow manually specified values for custom color light/dark modes
### Submission checklist

### Feature type

Configuration / customization

### Feature summary

For each custom color, instead of only being able to specify `color` under `[config.custom_colors]` or `[theme.templates.custom_colors]`, allow for specification of `color-light` and `color-dark` for light and dark modes respectively. If these are set, the color tokens (e.g. `on_{name}`, `{name}_container`) should be generated for light and dark modes separately; in particular, `colors.{name}_source.light` and `colors.{name}_source.dark` should return exactly the user-specified `color_light` and `color_dark`.

### Motivation / use case

Noctalia's app theming gives users global scope (similar to tools like Stylix) while providing far more granular control over app theming as well as automatic dark/light mode switching. However, maybe the only control it fails to give to users is how custom colors are changed between light and dark modes: currently, users are forced to accept what the automatic palette generation yields. This is in contrast to user palettes, where users are able to separately specify dark and light variants for the Material 3 colors.

Personally, I rely on custom colors a fair amount as I like having more granularity than the Material 3 roles, and I would prefer being able to specify what a custom color looks like in both light and dark mode instead of using the autogenerated ones. This feature is the last barrier to making Noctalia app theming the perfect ricing tool for me (replacing tools like Stylix or nix-colors, both of which are essentially limited to base16). Moreover, adding this feature looks to be a relatively small change in the source code (see below), and it should have minimal impact on Noctalia's performance/footprint.

### Proposed solution

If both `color-light` and `color-dark` are set, then perform most of the config in `TemplateEngine::processConfigTable` ([link](https://github.com/noctalia-dev/noctalia/blob/18b50824c375f825f4484b29c3ab9f8a0c6c2412/src/theme/template_engine.cpp#L1352)) for each of light and dark mode. In particular, create one `material_color_utilities::DynamicScheme`/palette per mode, and extrapolate the color tokens for light and dark mode from their respective, separate palettes.

### Alternatives considered

As far as I know, there isn't really any other way to automatically switch between light and dark themes across multiple apps which is nearly as convenient as Noctalia's app theming. The manual option is to write my own scripts to change and reload all my dotfiles at specific times (or for Nix, having to constantly change and rebuild my config).

This feature would also be effectively the same as if user palettes supported custom colors outside of the specified Material 3 roles in custom palettes, as then users would be able to specify separate light and dark versions of the colors manually. This would probably be nicer but will almost certainly take more work to implement.

_No response_

### Additional context

_No response_

AnonymousRand:
## Summary

Allows custom colors for user templates to specify `color_dark` and `color_light` separately as an alternative to just `color`, with dark and light mode palettes generated separately in that case. In particular, `<name>_source.dark` and `<name>_source.light` return the original `color_dark` and `color_light` values respectively. Moreover, the `color` value takes the value of `color_dark` if the former is not provided (e.g. for comparing colors).

## Motivation

Noctalia's custom colors gives users very granular control over implementing a consistent color scheme across apps, allowing customization to go beyond the Material 3 palette. However, while palette colors allow for specification of both light and dark mode variants, custom colors do not currently have this feature, instead forcing users to rely on an autogenerated value for the non-default mode.

This pull request adds this feature, giving users better control over their custom colors across both dark and light modes and making Noctalia an even better tool for customization.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Build / packaging

## Related Issue

Closes #3262.

## Testing

`just test` run, all tests passed (except for `noctalia:process`, but this appears to just be my terminal appending a BEL character to the end of a result string, as verified by printing out the result string). New functionality verified on my own NixOS setup.

## Manual Coverage

- [x] Tested on Niri
- [ ] Tested on Hyprland
- [ ] Tested on Sway

## Screenshots / Videos

N/A

## Checklist

## Additional Notes

I'm not the most knowledgeable about Material 3, so feel free to point out any issues like pulling the wrong color tones from palettes for color tokens or whatever. I also hope this doesn't "violate" the design philosophy of Material 3 and Noctalia; after all, users already have the ability to specify dark and light variants separately in user palettes.

### DIFF
--- a/src/config/config_types.h
+++ b/src/config/config_types.h
@@ -1265,8 +1265,8 @@ struct ThemeConfig {
   struct TemplateColorConfig {
     std::string name;
     std::string color;
-    std::string color_dark;
-    std::string color_light;
+    std::string color_dark = "";
+    std::string color_light = "";
     bool blend = true;
 
     bool operator==(const TemplateColorConfig&) const = default;
--- a/src/config/config_types.h
+++ b/src/config/config_types.h
@@ -1265,6 +1265,8 @@ struct ThemeConfig {
   struct TemplateColorConfig {
     std::string name;
     std::string color;
+    std::string color_dark;
+    std::string color_light;
     bool blend = true;
 
     bool operator==(const TemplateColorConfig&) const = default;
--- a/src/config/schema/config_schema.cpp
+++ b/src/config/schema/config_schema.cpp
@@ -945,7 +945,7 @@ namespace noctalia::config::schema {
 
     // [theme.templates.custom_colors]: a name-keyed map whose value is either a
     // bare color string or a { color, color_dark, color_light, blend } table.
-    // Kept only when name+color or name+color_dark+color_light are non-empty;
+    // Kept only when name+color are non-empty (color is set to color_dark if not provided);
     // emitted only when the list is non-empty.
     Field<ThemeConfig::TemplatesConfig> customColorsField() {
       return custom<ThemeConfig::TemplatesConfig>(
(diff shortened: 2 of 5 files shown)


ItsLemmy:
Thanks, made it  so there is no regression for existing configs.

PR #3337, opened 2026-07-09
fix(bluetooth): don't rebuild the whole device list on RSSI updates

FormalSnake:
## Summary

The Bluetooth tab in the control center rebuilds its entire device list
whenever `listKey()` changes, and that key includes every device's raw RSSI.
While a scan is running RSSI updates constantly, so the key changes on almost
every D-Bus signal and the whole list gets torn down and rebuilt, including the
paired and connected rows that never show RSSI at all.

This keys on the value that actually gets rendered instead: the signal percent,
and only for Available devices (the one bucket that draws the signal pill).
Connected and paired rows stop churning, and Available rows still refresh when
their shown percent changes.

## Motivation

With a scan active, the row you are aiming at (like Connect on an already
paired device) keeps getting recreated under the cursor as other nearby devices
advertise. It makes the panel hard to use for the common case of reconnecting a
known device.

## Type of Change

- [x] Bug fix

## Testing

Built from source (meson release, ninja), compiles clean with no new warnings.
Ran clang-format against the repo config with no changes to the hunk. The RSSI
pill and `signalPercentFromRssi()` are already gated to Available devices, so
removing raw RSSI from the key for the other buckets changes nothing that is
displayed.

## Manual Coverage

- [x] Tested on Hyprland

## Checklist

- [ ] I ran `just format` with clang-format v22+ installed, or this PR has no code changes.

### DIFF
--- a/src/shell/control_center/tabs/bluetooth_tab.cpp
+++ b/src/shell/control_center/tabs/bluetooth_tab.cpp
@@ -582,7 +582,11 @@ std::string BluetoothTab::listKey() const {
     key.push_back(':');
     key += std::to_string(static_cast<int>(d.batteryPercent));
     key.push_back(':');
-    key += std::to_string(static_cast<int>(d.rssi));
+    // Only the rendered signal percent (Available rows) belongs in the key; raw
+    // RSSI churn would otherwise rebuild the whole list, including connected rows.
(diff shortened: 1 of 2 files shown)


ItsLemmy:
THanks

PR #3343, opened 2026-07-09
feat: add M3-Oled generation scheme

danielepintore:
## Summary
Added a OLED version of the M3 Content color palette generation scheme. This is useful in certain models of OLED monitor to avoid the banding issued that is present with a dark gray background (probably it is also useful to implement OLED panel background, but that will be another pull request in case).
While there are community palette sources that implement a OLED color scheme they do not adapt the colors to the current wallpaper. This pull request tries to fix that.
For the name of the color scheme I've chose M3 OLED, but since that version doesn't actually exists maybe it should be named differently.

Here it is an example of the banding issue for reference (photo taken from a reddit post):
banding

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Build / packaging

## Manual Coverage

- [ ] Tested on Niri
- [x] Tested on Hyprland
- [ ] Tested on Sway

## Screenshots / Videos
image

## Checklist

### DIFF
--- a/assets/translations/en.json
+++ b/assets/translations/en.json
@@ -3413,6 +3413,7 @@
       "m3-content": "M3 Content",
       "m3-fruit-salad": "M3 Fruit Salad",
       "m3-monochrome": "M3 Monochrome",
+      "m3-oled": "M3 OLED",
       "m3-rainbow": "M3 Rainbow",
       "m3-tonal-spot": "M3 Tonal Spot",
       "muted": "Muted",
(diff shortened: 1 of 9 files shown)


ItsLemmy:
Not feeling this approach,  this addition is not a generation scheme but rather a modifier/transform. Also this would open the door to M3 Fruit Salad Oled, Vibran Oled, Faithful Oled, etc...   which sounds terrible on the long run.

I went ahead a rewrote your idea but in a transformative way, it's less intrusive and works with all kinds of palettes: wallpaper generated and predefined ones.  I don't have an OLED to test on, but it's looking I think.

See commit b721a345

[theme]
pure_black_dark = true

or 

image

danielepintore:
Tested right now, it works perfectly thanks!

PR #3352, opened 2026-07-10
feat(plugins): expose barWidget.outputName to bar widgets

gegnep:
## Summary

Adds a `barWidget.outputName()` binding for bar-widget plugins, returning the connector name of the output the widget instance is placed on, mirroring the existing `barWidget.isVertical()` binding.

## Motivation

A bar-widget plugin instance currently has no way to know which monitor
it is rendered on. `noctalia.focusedOutputName()` exists but returns the
globally focused output — the same value for every bar on every monitor
— so it can't be used to scope per-monitor content. Native widgets get
their output at construction (e.g. the workspaces widget filters its
workspaces by it); plugins can't replicate that, which forces manual
per-placement settings like "type your connector name here".

Concrete case: a niri workspace/taskbar plugin that should show each
bar's own monitor's workspaces, the way the built-in workspaces widget
does. With this binding it Just Works on multi-monitor setups with zero
configuration.

## Type of Change

- [x] New feature

## Related Issue

—

## Testing

- Built from source on NixOS (patch applied on `main` @ 3137323) and ran the full shell.
- Exercised live by a niri workspace/taskbar bar-widget plugin on a two output setup: each bar's widget instance reports its own connector name and filters its content accordingly; the same plugin placed on both bars returns different values.
- Verified `nil` handling: the binding returns nil (not an empty string) when the snapshot carries no output name, and plugins without the call are unaffected.
- `just format` (clang-format 22) — no diff.

## Manual Coverage

- [x] Tested on Niri
- [ ] Tested on Hyprland
- [ ] Tested on Sway

## Screenshots / Videos

image

workspaces on mon 1

image

workspaces on mon 2 (only 1)

## Checklist

- [x] I ran the relevant build or test command not run.
- [x] I will update end-user documentation aft change user-facing configuration or behavior.

## Additional Notes

Returns the raw connector name rather than a richer output object to keep the surface minimal; plugins needing geometry can cross-reference `noctalia.outputs()` by name. Docs follow-up after merge: add `outputName()` next to `isVertical()` on the bar-widget API page.

Disclaimer: AI-assisted tooling was used.

### DIFF
--- a/src/scripting/plugin_bindings.cpp
+++ b/src/scripting/plugin_bindings.cpp
@@ -207,6 +207,16 @@ namespace {
     return 1;
   }
 
+  int luau_outputName(lua_State* L) {
+    auto* context = getContext(L);
+    if (context == nullptr || context->snapshot.outputName.empty()) {
+      lua_pushnil(L);
+      return 1;
+    }
+    lua_pushlstring(L, context->snapshot.outputName.data(), context->snapshot.outputName.size());
+    return 1;
+  }
+
   int luau_setVisible(lua_State* L) {
     bool visible = lua_toboolean(L, 1) != 0;
     if (auto* context = getContext(L)) {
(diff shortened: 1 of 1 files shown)


ItsLemmy:
Thanks

gegnep:
thanks for the shell and getting the docs done before i could <3

PR #3356, opened 2026-07-10
feat(plugins): add noctalia.appIconPath icon resolution binding

gegnep:
## Summary

Adds a `noctalia.appIconPath(appId, sizePx?)` binding that resolves an
app id (or raw icon name) to an icon file path using the same components
the native taskbar uses: `app_identity::findDesktopEntry` over the
shared desktop-entry index for the icon name, then `IconResolver` with
the caller's target size. Returns nil when nothing resolves. To make the
lookup safe from plugin script worker threads, `DesktopEntryCache` gains
a mutex-guarded, non-refreshing `desktopEntriesSnapshot()` accessor
(freshness stays owned by the main thread's reload path) and the
resolver instance is `thread_local`.

## Motivation

Plugins that display windows or applications (taskbars, docks, window
switchers) currently have no icon story: `ui.image` renders by file path
only, and there is no lookup from an app id to an icon. Authors are left
re-implementing .desktop scanning and icon-theme resolution in Luau —
poorly, compared to the host's resolver (no index.theme metadata, no GTK
cache, no size selection). This binding hands plugins exactly the icons
Noctalia itself shows, with one call.

## Type of Change

- [x] New feature

## Testing

- Built from source on NixOS (patch on `main` @ 3137323), full shell run.
- Exercised live by a niri taskbar plugin resolving icons for Firefox,
  Ghostty, Kitty, and Electron apps (StartupWMClass matching) at 2×
  display size — results match the native taskbar's icons exactly;
  unresolved ids return nil and the plugin falls back gracefully.
- Verified thread-safety design: worker-thread reads copy under the new
  mutex; the main-thread reload path swaps under the same lock; no
  refresh is triggered off the main thread.
- `just format` (clang-format 22) — no diff.

## Manual Coverage

- [x] Tested on Niri

## Screenshots / Videos

image

## Checklist

## Additional Notes

The binding intentionally returns a path rather than image data so it
composes with the existing `ui.image` path prop and the reconciler's
image cache. Unmatched inputs fall through to
so plugins can also resolve themed icons (e.g. `"folder-music"`)
directly. Docs follow-up after merge: add to the plugin API reference

Disclaimer: AI-assisted tooling was used.

### DIFF
--- a/src/scripting/luau_host.cpp
+++ b/src/scripting/luau_host.cpp
@@ -332,18 +332,19 @@ namespace {
   int luau_appIconPath(lua_State* L) {
     size_t len = 0;
     const char* appId = luaL_checklstring(L, 1, &len);
-    const int targetSize = static_cast<int>(luaL_optinteger(L, 2, 0));
+    const int targetSize = luaL_optinteger(L, 2, 0);
 
     std::string iconName;
-    if (const auto entry = app_identity::findDesktopEntry(std::string_view(appId, len), desktopEntriesSnapshot());
+    const auto entries = desktopEntriesSnapshot();
+    if (const auto entry = app_identity::findDesktopEntry(std::string_view(appId, len), *entries);
         entry.has_value() && !entry->icon.empty()) {
       iconName = entry->icon;
     } else {
       iconName.assign(appId, len);
     }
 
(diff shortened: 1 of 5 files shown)


ItsLemmy:
Thanks

PR #3383, opened 2026-07-11
feat(calendar): Add option to show week numbers

borgsten:
## Summary
This adds a column displaying the ISO-8601 week numbers to the left of the date grid in the calendar tab of the control center. 

It is an opt-in feature which can be enabled in settings under the services tab with the other calendar settings.

## Motivation

Week numbers are heavily used in certain locales, it is very useful to get the current(and future) week numbers at a glance.

Furthermore this seems to have been supported before the rewrite.

## Type of Change

- [x] New feature

## Testing
`just test && just test release`
`python3 tools/i18n-check.py`

Tested with `LC_TIME=en_DK.UTF-8` and the numbers match other calendars.

Also tested setting the new config value directly in both `~/.config/` and `~/.local/state` toml file.

## Manual Coverage

- [x] Tested on Hyprland

## Screenshots / Videos
Setting enabled:

image

Setting disabled:
image

Settings:
image

## Checklist

## Additional Notes
Let me know if I should add some header for the week number column instead of keeping it blank or if the colors should be changed.

### DIFF
--- a/assets/translations/en.json
+++ b/assets/translations/en.json
@@ -2445,6 +2445,10 @@
           "description-unavailable": "No Secret Service provider is available. Calendar events remain available for this session only.",
           "label": "Calendar Event Storage"
         },
+        "calendar-week-numbers": {
+          "description": "Show ISO 8601 week numbers alongside the calendar grid",
+          "label": "Week Numbers"
+        },
         "custom-schedule": {
           "description": "Use manual sunrise/sunset times instead of computing from coordinates",
           "label": "Custom Schedule"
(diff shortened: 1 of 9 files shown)


borgsten:
Only tested this on hyprland as I do not think it should make any difference for the calendar widget in the control center. But if it does, then I can perform further testing in other compositors.

borgsten:
I making the week numbers underlined which I think I like more than my initial implementation as it might remove some confusion if week number and dates line up.

image


If you prefer that look I can expose it through `LabelProps`, if there are any other UI directions you want to go then let me know.

ItsLemmy:
Reworked it to make it a bit simpler. %V gives us the week number directly.

image

PR #3387, opened 2026-07-12
fix(pw): show all input devices in audio panel, not just the default source

### LINKED ISSUE #3386: [BUG] Audio panel lists only the default input device; all other microphones hidden (SPA_DIRECTION_INPUT == 0 fails the wantDir != 0 guard)
### Submission checklist

### Bug description

The audio panel's **input (microphone) list only ever shows a single device (the current default source).** All other input devices (built-in laptop mic, Bluetooth headset mic, etc.) are hidden, even though they are fully functional and visible in `wpctl` / `pavucontrol`. The **output** list is unaffected and correctly shows all sinks.

I traced this to a direction-guard bug in `src/pipewire/pipewire_service.cpp`: `SPA_DIRECTION_INPUT == 0`, but the availability logic is guarded with `wantDir != 0`, so the guard is never true for input devices. Every `Audio/Source` node therefore gets `available = false` and is filtered out of the panel by `availableDevices()`, which keeps only `device.available || device.id == defaultId`. The default source is the sole survivor.

### Root cause

In `src/pipewire/pipewire_service.cpp`:

```cpp
[[nodiscard]] std::uint32_t routeDirectionForMediaClass(std::string_view mediaClass) {
  if (mediaClass == "Audio/Source") { return SPA_DIRECTION_INPUT; }   // == 0
  if (mediaClass == "Audio/Sink")   { return SPA_DIRECTION_OUTPUT; }  // == 1
  return 0;                                                            // also 0 -> "no direction" sentinel
}
```

`SPA_DIRECTION_INPUT` is `0` (see `spa/utils/defs.h`), which collides with the `return 0` "not a device / no direction" sentinel. The availability computation then guards on `wantDir != 0`:

```cpp
const std::uint32_t wantDir = routeDirectionForMediaClass(nd->mediaClass);
const DeviceRouteData* activeRoute = wantDir != 0 ? activeRouteForDirection(nd->routes, wantDir) : nullptr;
...
    if (activeRoute == nullptr && wantDir != 0) {
      activeRoute = activeRouteForDirection(device->routes, wantDir);
    }
...
const bool hasDirRoutes = std::ranges::any_of(nd->routes, matchesDir)          // matchesDir uses wantDir, NOT guarded
    || (device != nullptr && std::ranges::any_of(device->routes, matchesDir));
node.available = activeRoute != nullptr || !hasDirRoutes;
```

For any `Audio/Source`, `wantDir == 0`, so `activeRoute` is **always** `nullptr`, while `hasDirRoutes` is `true` (input routes exist). Result: `node.available = false || !true = false`. Then in `audio_tab.cpp`:

```cpp
// Hide devices whose active route is unavailable ... but always keep the current default.
for (const auto& device : devices) {
  if (device.available || device.id == defaultId) { out.push_back(device); }
}
```

…drops every input except the default. Outputs are fine because `SPA_DIRECTION_OUTPUT == 1` passes the `!= 0` guard.

The same guard also breaks `recomputeEffectiveMute()` for inputs (`nodeRoute` / `deviceRoute` are always `nullptr` for sources), so per-input route mute state is not resolved either.

### Evidence

`pw-dump` route availability on the affected machine — the built-in mic's active input route reports `unknown` (which the code comment explicitly says should stay visible), yet it is hidden:

```
DEVICE alsa_card.pci-0000_c2_00.6
  [Route] dir=Input  avail=unknown  name=analog-input-mic          <- hidden (bug)
  [Route] dir=Output avail=unknown  name=analog-output-speaker     <- shown (OK)
```

The only visible input was whatever happened to be the default source at the time.

### Proposed fix

Guard on the media class being a device node rather than on `wantDir != 0`. `isDeviceNode` is already used this way elsewhere in the same file (the `SPA_PARAM_Route` parsing block), so this is consistent with existing style.

```diff
@@ PipeWireService::(node availability computation)
     const std::uint32_t wantDir = routeDirectionForMediaClass(nd->mediaClass);
-    const DeviceRouteData* activeRoute = wantDir != 0 ? activeRouteForDirection(nd->routes, wantDir) : nullptr;
+    const bool isDeviceNode = nd->mediaClass == "Audio/Sink" || nd->mediaClass == "Audio/Source";
+    const DeviceRouteData* activeRoute = isDeviceNode ? activeRouteForDirection(nd->routes, wantDir) : nullptr;
     const DeviceData* device = nullptr;
     if (nd->deviceId != 0) {
       if (const auto devIt = m_devices.find(nd->deviceId); devIt != m_devices.end()) {
         device = &devIt->second;
-        if (activeRoute == nullptr && wantDir != 0) {
+        if (activeRoute == nullptr && isDeviceNode) {
           activeRoute = activeRouteForDirection(device->routes, wantDir);
         }
       }
     }
@@ PipeWireService::recomputeEffectiveMute
   const std::uint32_t wantDir = routeDirectionForMediaClass(nd.mediaClass);
-  const DeviceRouteData* nodeRoute = wantDir != 0 ? activeRouteForDirection(nd.routes, wantDir) : nullptr;
+  const bool isDeviceNode = nd.mediaClass == "Audio/Sink" || nd.mediaClass == "Audio/Source";
+  const DeviceRouteData* nodeRoute = isDeviceNode ? activeRouteForDirection(nd.routes, wantDir) : nullptr;
   const DeviceRouteData* deviceRoute = nullptr;
-  if (nd.deviceId != 0 && wantDir != 0) {
+  if (nd.deviceId != 0 && isDeviceNode) {
     const auto it = m_devices.find(nd.deviceId);
     if (it != m_devices.end()) {
       deviceRoute = activeRouteForDirection(it->second.routes, wantDir);
     }
   }
```

Alternatively, change `routeDirectionForMediaClass` to return `std::optional<std::uint32_t>` (`std::nullopt` for non-device classes) and check `.has_value()`, which removes the sentinel collision entirely.

### Workaround

Until fixed, set the desired microphone as the default source (`pavucontrol`, `wpctl set-default <id>`, or `pactl set-default-source <name>`); it will then appear and be used, though you cannot switch inputs from within the panel.

### Steps to reproduce

1. Have two or more input devices available (e.g. a built-in analog mic **and** a USB/Bluetooth mic).
2. Open the audio panel in the control center.
3. Look at the **Input** device list.

### Expected behavior

All input devices whose active route is available (or reports `unknown`, as internal mics do) should be listed, exactly like output devices so you can switch between microphones.

### Actual behavior

Only one input is listed: whichever is the current default source. Every other microphone is hidden. Changing the default source (e.g. `pactl set-default-source …` / `wpctl set-default …`) swaps which single mic is shown, confirming that only the default ever appears.

### Logs / error output

```text
None relevant (functional/logic bug).
```

### Noctalia version

v5.0.0 (`main` at time of report also affected). Observed on the CachyOS package `noctalia 5.0.0_beta2`.

### Compositor

Hyprland

### Distribution

Arch-based

### Environment information

```text
Installation method: pacman (CachyOS repo), package noctalia 5.0.0_beta2
GPU: AMD (amdgpu)
Audio: PipeWire 1.6.7 / WirePlumber 0.5.15
```

### Additional context

The `availableDevices()` "hide unavailable routes" feature and the `Cards that report 'unknown' ... stay visible` comment both work correctly for outputs; only inputs are affected, purely because `SPA_DIRECTION_INPUT` is `0`.

TheBinaryLoop:
## Summary

The audio panel's input list only ever showed a single microphone — the current default source — hiding every other input device (built-in mic, USB, Bluetooth headset mic, etc.). Outputs were unaffected.

Root cause: `routeDirectionForMediaClass()` returns `SPA_DIRECTION_INPUT` for `Audio/Source`, and `SPA_DIRECTION_INPUT == 0`, which collides with the `0` value used as the "no direction" sentinel. The availability logic in `rebuildState()` guarded on `wantDir != 0`, so for every input device `activeRoute` was forced to `nullptr` while `hasDirRoutes` stayed `true`, giving `node.available = false`. `availableDevices()` then keeps only `device.available || device.id == defaultId`, so all non-default inputs were dropped. The same `wantDir != 0` guard also disabled route resolution in `recomputeEffectiveMute()` for inputs.

The fix guards on the media class being a device node (`Audio/Sink || Audio/Source`) instead of `wantDir != 0`, matching the existing `isDeviceNode` check used during `SPA_PARAM_Route` parsing. Outputs were never affected because `SPA_DIRECTION_OUTPUT == 1`.

## Motivation

Users with more than one microphone could not see or switch between them from the audio panel — only the default input was listed. Reported in #3386.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Build / packaging

## Related Issue

Closes #3386

## Testing

- `just build` (debug) — succeeds, no new warnings.
- `just test debug --print-errorlogs` — 38 passed, 0 failed.
- `just format` with clang-format 22.1.8 — no changes beyond this diff.
- `clang-tidy -p build-debug -warnings-as-errors='*' src/pipewire/pipewire_service.cpp` — no reported diagnostics.
- Manual: on the affected machine (PipeWire 1.6.7 / WirePlumber 0.5.15) with a built-in analog mic, a USB dock input, and a Bluetooth headset, the audio panel previously listed only the default source. After the fix all available inputs appear simultaneously and can be selected. Verified via `pw-dump` that the built-in mic's active input route reports `available = unknown` (should stay visible), confirming availability was never the discriminator — only the direction guard.

## Manual Coverage

- [ ] Tested on Niri
- [x] Tested on Hyprland
- [ ] Tested on Sway

## Screenshots / Videos

N/A — device-list logic fix, no visual/layout change.

## Checklist

## Additional Notes

The change is intentionally minimal and mirrors the `isDeviceNode` idiom already present in `pipewire_service.cpp`. An alternative would be to make `routeDirectionForMediaClass()` return `std::optional<std::uint32_t>` (`std::nullopt` for non-device classes) to remove the sentinel collision entirely — happy to take that route instead if preferred.

### DIFF
--- a/src/pipewire/pipewire_service.cpp
+++ b/src/pipewire/pipewire_service.cpp
@@ -1604,12 +1604,15 @@ void PipeWireService::rebuildState() {
     // explicitly unavailable and no available alternative is hidden. Cards that report "unknown"
     // (many HDA/HiFi setups) stay visible.
     const std::uint32_t wantDir = routeDirectionForMediaClass(nd->mediaClass);
-    const DeviceRouteData* activeRoute = wantDir != 0 ? activeRouteForDirection(nd->routes, wantDir) : nullptr;
+    // SPA_DIRECTION_INPUT == 0, so `wantDir != 0` would wrongly exclude every Audio/Source; guard on the
+    // media class being a device node instead (matches the isDeviceNode check used during route parsing).
+    const bool isDeviceNode = nd->mediaClass == "Audio/Sink" || nd->mediaClass == "Audio/Source";
+    const DeviceRouteData* activeRoute = isDeviceNode ? activeRouteForDirection(nd->routes, wantDir) : nullptr;
     const DeviceData* device = nullptr;
     if (nd->deviceId != 0) {
       if (const auto devIt = m_devices.find(nd->deviceId); devIt != m_devices.end()) {
         device = &devIt->second;
-        if (activeRoute == nullptr && wantDir != 0) {
+        if (activeRoute == nullptr && isDeviceNode) {
           activeRoute = activeRouteForDirection(device->routes, wantDir);
         }
       }
@@ -1674,9 +1677,11 @@ void PipeWireService::rebuildState() {
 
 void PipeWireService::recomputeEffectiveMute(NodeData& nd) {
   const std::uint32_t wantDir = routeDirectionForMediaClass(nd.mediaClass);
-  const DeviceRouteData* nodeRoute = wantDir != 0 ? activeRouteForDirection(nd.routes, wantDir) : nullptr;
+  // SPA_DIRECTION_INPUT == 0, so guard on the media class rather than `wantDir != 0` (which would skip sources).
+  const bool isDeviceNode = nd.mediaClass == "Audio/Sink" || nd.mediaClass == "Audio/Source";
+  const DeviceRouteData* nodeRoute = isDeviceNode ? activeRouteForDirection(nd.routes, wantDir) : nullptr;
   const DeviceRouteData* deviceRoute = nullptr;
-  if (nd.deviceId != 0 && wantDir != 0) {
+  if (nd.deviceId != 0 && isDeviceNode) {
     const auto it = m_devices.find(nd.deviceId);
     if (it != m_devices.end()) {
       deviceRoute = activeRouteForDirection(it->second.routes, wantDir);


ItsLemmy:
Thanks

PR #3388, opened 2026-07-12
feat(idle): add per-behavior locked screen timeout

TheBinaryLoop:
## Summary

Idle behaviors get an optional `locked_timeout` (seconds). While the session is locked, a behavior with `locked_timeout > 0` arms at that value instead of its normal `timeout`, and goes back to the normal value on unlock. Setting `timeout = 0` with `locked_timeout > 0` makes a behavior run only while locked.

The main use is `screen_off`: turn the display off quickly on the lock screen (and again after each wake, since the behavior re-arms) without making the display sleep aggressively while you're actually using the machine.

image

## Motivation

A single `timeout` can't express "turn the screen off 10s after locking, but leave it alone for 10 minutes while I'm working." You had to pick one value for both states. This gives the locked state its own, shorter timeout.

## Type of Change

- [x] New feature

## Related Issue

None.

## Testing

- `just build debug`
- `just test debug`, 38/38 pass (config_schema_roundtrip covers the new field)
- `clang-format --dry-run -Werror` clean on the changed files
- `clang-tidy -p build-debug` clean on the changed files
- `noctalia config validate` on a config using `locked_timeout` reports valid

Ran the dev build on my setup and went through the whole cycle: idle while unlocked uses the normal timeout; after locking, `screen_off` fires at the locked timeout; waking brings the display back; leaving it idle turns it off again; unlocking leaves the display on.

While testing I hit a bug this change introduced: on unlock the notifications re-arm, which tore down the `screen_off` behavior while it was already off without running its resume, leaving the display black. Fixed by running the resume action when re-arming a behavior that is still in the idled state.

## Manual Coverage

- [x] Tested on Hyprland

## Checklist

## Additional Notes

Off by default (`locked_timeout = 0` for every behavior), so existing configs behave exactly as before.

### DIFF
--- a/assets/translations/en.json
+++ b/assets/translations/en.json
@@ -1099,6 +1099,7 @@
           "suspend": "Suspend"
         },
         "kind-section-label": "Action",
+        "locked-timeout-label": "Timeout while locked (seconds, 0 = off)",
         "name-label": "Name",
         "name-placeholder": "idle-behavior",
         "resume-command-label": "Resume command",
--- a/src/app/application_ui.cpp
+++ b/src/app/application_ui.cpp
@@ -324,11 +324,13 @@ void Application::initLockScreenAndSession() {
       [this]() {
         m_idleGraceOverlay.hide();
         m_lockscreenWidgetsController.onLockStateChanged();
+        m_idleManager.setSessionLocked(true);
         m_hookManager.fire(HookKind::SessionLocked);
       },
       [this]() {
         m_idleGraceOverlay.hide();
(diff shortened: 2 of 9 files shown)


TheBinaryLoop:
The CI failure doesn't look like it's caused by my change.

Anachron:
I would find this very useful!

Ly-sec:
Hey @TheBinaryLoop , this looks good to me. I'll merge this one :) sorry for the super late reply!

PR #3412, opened 2026-07-13
rfkill: apply setRfkillSoftBlocked to all matching devices

notiant:
Currently `setRfkillSoftBlocked()` is only applied to the first matching rfkill entry that is found. Most laptops have multiple rfkill switches per radio type though (both hardware and platform-level from the vendor's kernel module). On my laptop, enabling Bluetooth with Noctalia only unblocks one of the 2 Bluetooth switches when they are soft-blocked. I've added a loop to unblock all matching entries to fix this issue.

### DIFF
--- a/src/system/rfkill_helper.cpp
+++ b/src/system/rfkill_helper.cpp
@@ -2,6 +2,7 @@
 
 #include "core/log.h"
 
+#include <algorithm>
 #include <cerrno>
 #include <cstdio>
 #include <cstring>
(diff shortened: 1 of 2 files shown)


ItsLemmy:
I changed it to use RFKILL_OP_CHANGE_ALL   instead of a loop which would not have been accurate since result is overwritten every iteration.

Can you test it, and let me know? Thanks

notiant:
> I changed it to use RFKILL_OP_CHANGE_ALL instead of a loop which would not have been accurate since result is overwritten every iteration.
> 
> Can you test it, and let me know? Thanks

I tested it and works. Yes, the result value would have been inaccurate with my changes, this is the better way.

PR #3413, opened 2026-07-13
fix(greeter-sync): use wallpaper for greeter pinned output

luxus:
## Summary

Greeter appearance sync can copy the **wrong monitor’s wallpaper** on multi-monitor setups.

### Root cause

`resolveSyncWallpaperPath()` only preferred `greeter.toml` `[output].name` when theme source was **not** wallpaper. With wallpaper-based theming (common), it fell through to `getGreeterSyncWallpaperPath()`, which uses the **last applied** wallpaper path — often a secondary / portrait panel.

Example (lea):

| Connector | Wallpaper |
|-----------|-----------|
| DP-2 (greeter pin) | Widescreen |
| HDMI-A-1 | Portrait |

Sync staged the portrait image while the greeter is pinned to DP-2.

### Fix

Always try wallpaper for `[output].name` first; only then fall back to the previous selection logic.

## Test plan

- [ ] Multi-monitor with different wallpapers; greeter.toml `name` set to one connector
- [ ] Sync Now → `/var/lib/noctalia-greeter/wallpaper.webp` matches that connector’s image
- [ ] No `[output].name` → previous fallback behavior still works
- [ ] Palette / theme colors still sync as before

### DIFF
--- a/src/shell/greeter/greeter_appearance_sync.cpp
+++ b/src/shell/greeter/greeter_appearance_sync.cpp
@@ -96,14 +96,19 @@ namespace {
   }
 
   [[nodiscard]] std::string resolveSyncWallpaperPath(const ConfigService& configService) {
-    const Config& config = configService.config();
-    if (config.theme.source != PaletteSource::Wallpaper) {
-      if (const auto output = readGreeterConfiguredOutput(); output.has_value() && !output->empty()) {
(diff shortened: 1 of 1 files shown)


luxus:
Superseded by per-output wallpaper staging PR (stages all monitors + pin-aware legacy fallback).

PR #3414, opened 2026-07-13
feat(greeter-sync): stage per-output wallpapers for multi-monitor

luxus:
## Summary

Greeter Sync Now only staged a **single** wallpaper image. On multi-monitor setups with different images per connector (ultrawide vs portrait), the greeter could only show one of them—and often the wrong one when theming used the last-applied wallpaper.

This stages **every** monitor wallpaper and writes a per-connector map for greeters that support it, while keeping the legacy single `wallpaper` field for older greeters.

### Changes

- Stage `wallpaper-<connector>.*` for each monitor wallpaper from Noctalia settings
- Write `appearance.json` `wallpapers` map: `{ "DP-2": { path, fill_mode, fill_color }, ... }`
- Keep legacy `wallpaper` field as fallback
- Resolve the single fallback image by preferring `greeter.toml` `[output].name`, then default path, then previous palette/last-applied logic (supersedes the pin-only approach in #3413)

### Compatibility

| Greeter | Behavior |
|---------|----------|
| New (reads `wallpapers` map) | Each greeter view uses the image for its bound connector |
| Old (single `wallpaper` only) | Still gets one staged image; pin-aware fallback when `[output].name` is set |

No user-facing settings: uses existing per-monitor wallpaper paths + Sync Now / greeter apply helper.

## Test plan

- [ ] Multi-monitor with different wallpapers → Sync Now → `/var/lib/noctalia-greeter/wallpaper-DP-2.*` and `wallpaper-HDMI-A-1.*` (or your connectors) exist
- [ ] `appearance.json` has both `wallpaper` and `wallpapers`
- [ ] With greeter that supports the map + pin `name = "DP-2"` → ultrawide image
- [ ] Pin `name = "HDMI-A-1"` → portrait image
- [ ] Old greeter binary (no map) still loads legacy `wallpaper.path`
- [ ] Theme colors / scheme still sync

Supersedes #3413.

### DIFF
--- a/src/config/config_service.h
+++ b/src/config/config_service.h
@@ -80,6 +80,9 @@ public:
 
   // Persisted wallpaper paths (written to settings.toml, app-managed).
   [[nodiscard]] std::string getWallpaperPath(const std::string& connectorName) const;
+  [[nodiscard]] const std::unordered_map<std::string, std::string>& monitorWallpaperPaths() const {
+    return m_monitorWallpaperPaths;
+  }
   [[nodiscard]] std::string getDefaultWallpaperPath() const;
   // Last applied wallpaper, else default. Drives palette generation and template previews.
   [[nodiscard]] std::string getPaletteWallpaperPath() const;
--- a/src/shell/greeter/greeter_appearance_sync.cpp
+++ b/src/shell/greeter/greeter_appearance_sync.cpp
@@ -131,11 +131,18 @@ namespace {
   };
 
   // Stage every per-monitor wallpaper so the greeter can pick by connector name.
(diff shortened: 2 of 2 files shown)


luxus:
Addressed Copilot review:

1. **`color:` wallpapers** — included in the per-output map with `path` kept as the `color:#…` spec (no invalid `/var/lib/…/` path).
2. **Map paths** — staged file path only when `installedName` is set; otherwise use `sourcePath`.
3. **Stable order** — sort staged entries by connector name before writing the map / choosing legacy fallback.

Ly-sec:
Hey @luxus , sorry for the late reply, I'll merge this one and the one from the greeter repo. Thank you very much.

Issue #3421, opened 2026-07-13
[BUG] Collapsibles only open sometimes wth UI scale > 1.0


Kruemmelspalter:
### Submission checklist

- [x] I have searched existing issues and confirmed this is not a duplicate.
- [x] I am using the latest available version of Noctalia.

### Bug description

When UI scale is set to any value > 1.0, collapsibles, for example in the bluetooth tab of the control center, don't always open when clicked.
I haven't been able to determine what decides if they open or not, it seems random.
Sometimes, they flicker a bit, as if there were two clicks right after each other opening it and then closing it again.
Closing it didn't seem to have this problem.

### Steps to reproduce

1. set UI scale to 1.10 in settings
2. open bluetooth tab of control center
3. click on any collapsible / device

### Expected behavior

The collapsible opens when clicked and closes again when clicked.

### Actual behavior

The collapsible only opens sometimes.

### Logs / error output

```text

```

### Noctalia version

e0880261136557b9dee29427e606692bfb412de3

### Compositor

Niri

### Distribution

NixOS

### Environment information

```text
Installation method: source build (`nix develop` => `just run m=release`)
GPU: Intel Corporation Raptor Lake-P [UHD Graphics] (CPU: 13th Gen Intel(R) Core(TM) i5-1334U)
```

PR #3422, opened 2026-07-13
feat(control-center/bluetooth_tab): show MAC address in collapsible body

Kruemmelspalter:
## Summary

The bluetooth tab in the control center now shows the MAC address of the connected device in the device dropdown, where there was only the Auto-Reconnect toggle before.

## Motivation

Two bluetooth devices can have the same name / alias and be indistinguishable in the device list except for remembering that one comes first and the other comes after.

This change shouldn't clutter the UI, because most often ppl only have one collapsible open; so the additional line is mostly visible only once on screen.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Build / packaging

## Testing
both on `m=debug` and `m=release`:
- `just build` / `just run` runs successfully
- `just test` has the same two tests failing as on main (`process` and `plugin_git_export`)

## Manual Coverage

- [x] Tested on Niri
- [ ] Tested on Hyprland
- [ ] Tested on Sway
    collapsible only opens sometimes at UI scales > 1.0; same behaviour on main; see #3421

## Screenshots / Videos

image

## Checklist

## Additional Notes

I'm very open to making this opt-in with a config option, but imo I don't really see much reason not to show it.

### DIFF
--- a/assets/translations/en.json
+++ b/assets/translations/en.json
@@ -182,6 +182,7 @@
     },
     "bluetooth": {
       "accept": "Accept",
+      "address": "Address",
       "auto-reconnect": "Auto-reconnect",
       "bluetooth": "Bluetooth",
       "enter-code": "Enter code",
--- a/src/shell/control_center/tabs/bluetooth_tab.cpp
+++ b/src/shell/control_center/tabs/bluetooth_tab.cpp
@@ -252,8 +252,38 @@ public:
 
     setHeader(std::move(header));
 
+    auto bodyColumn = ui::column(
+        {
+            .align = FlexAlign::Stretch,
+            .gap = Style::spaceSm * scale,
+        },
+        ui::row(
+            {
+                .align = FlexAlign::Center,
+                .gap = Style::spaceSm * scale,
+                .configure =
+                    [scale](Flex& body) {
+                      body.setPadding(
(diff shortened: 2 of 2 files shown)


bhuba1:
Cool feature! 
One thought: would it make sense to mask the address by default with a small eye icon next to it, that reveals it on click? Keeps it available without exposing it in screenshots/streams, and avoids needing a separate config option.

Kruemmelspalter:
> Cool feature! 
> One thought: would it make sense to mask the address by default with a small eye icon next to it, that reveals it on click? Keeps it available without exposing it in screenshots/streams, and avoids needing a separate config option.

it's already in the collapsible, so it only shows when the device options (currently only containing auto-reconnect) are opened

but i can also add a toggle, that wouldn't be a problem
it would make it two clicks instead of one to show it but i think that's okay

bhuba1:
> > Cool feature!
> > One thought: would it make sense to mask the address by default with a small eye icon next to it, that reveals it on click? Keeps it available without exposing it in screenshots/streams, and avoids needing a separate config option.
> 
> it's already in the collapsible, so it only shows when the device options (currently only containing auto-reconnect) are opened
> 
> but i can also add a toggle, that wouldn't be a problem it would make it two clicks instead of one to show it but i think that's okay

Yeah right, the masking might be an overkill then

ItsLemmy:
Thanks!

PR #3427, opened 2026-07-14
test: cover Triad workspace backend state

greenm01:
The Triad workspace backend had no direct tests, although it supplies the workspace and window state used by the taskbar and dock.

This adds focused coverage for full snapshots and incremental window updates. It checks synthetic-workspace filtering, output assignments, overview state, focused-window tracking, 64-bit window IDs, title cleanup, malformed records, and removal of windows omitted from a replacement snapshot. The test uses a private accessor, so the backend’s public interface stays unchanged.

Verification:

- Built the full Noctalia application with tests enabled.
- Ran all 39 Meson tests successfully.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1542,6 +1542,18 @@ dock_pinned_apps_test = executable('dock_pinned_apps_test',
 
 test('dock_pinned_apps', dock_pinned_apps_test)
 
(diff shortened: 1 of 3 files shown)


Ly-sec:
Thank you very much @greenm01 , I'll merge it.

greenm01:
Thanks!

PR #3441, opened 2026-07-14
Refactor widget definition

Ape:
I had this refactoring idea for more declarative widget definitions. I've migrated the battery widget as a representative for review. If you like the direction, we can go ahead and migrate the rest of the widgets.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -690,6 +690,7 @@ _noctalia_sources = files(
   'src/shell/bar/widgets/active_window_widget.cpp',
   'src/shell/bar/widgets/audio_visualizer_widget.cpp',
   'src/shell/bar/widgets/battery_widget.cpp',
+  'src/shell/bar/widgets/battery_widget_definition.cpp',
   'src/shell/bar/widgets/bluetooth_widget.cpp',
   'src/shell/bar/widgets/control_center_widget.cpp',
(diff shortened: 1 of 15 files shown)


ItsLemmy:
Very nice.

Suggestions (non-blocking)

- Lost diagnostic: the old kLog.warn("invalid widget.{}.display_mode …") at widget-creation is gone; an invalid enum now silently resolves to the default. Config-load validation still diag.warns on the enum, so coverage isn't zero, but it's a minor reduction. Fine to leave.

- Test coverage is the main gap. widget_definition.h is 536 lines with substantial subtle logic — int64 range-edge math in effectiveNumericRange, numericSettingValueAs clamping, validateStep, choice dedup/type-mixing validation — but battery only exercises bool/enum/color/string fields. None of the numeric-range or step paths are covered by any test, and there's no direct test of field()/resolve()/schemaFields(). Before migrating numeric widgets (scale, sizes, thresholds), I'd add a focused unit test that locks these invariants — that's where a future migration is most likely to silently drift. (One nuance to guard: std::clamp(concrete, range.lower, range.upper) is safe only because effectiveNumericRange throws on lower > upper; a test would pin that, and it's exactly the std::clamp hi<lo hazard your clampOrdered memo flagged.)

PR #3452, opened 2026-07-14
bug(input): prevent info leak from password input

borgsten:
## Summary
While in password mode, Ctrl+Left/Right/Backspace/Delete will ignore word boundaries and find start/end of input. Double clicking the input will also mark the whole input. Copy and cut will not set the selected input to the clipboard.

Undo history is also disabled in password mode but this can be removed if not a wanted change.

## Motivation

When Input was in password mode it still allowed some keybindings to operate on the password based on word boundaries which leaks information of the password if observed. The password could also be copied and cut and leaked to the clipboard manager.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Build / packaging

## Testing

Added a new test suite `input_password_mode_test` which passes. 

Manual testing on the lock screen.

## Manual Coverage

- [ ] Tested on Niri
- [x] Tested on Hyprland
- [ ] Tested on Sway

## Checklist

## Additional Notes
Let me know if the undo history change should be kept, feels a bit out of place in a password input but I can omit the change if not wanted.


meson.build
```diff
@@ -1038,6 +1038,42 @@ input_area_test = executable('input_area_test',
...
+    'src/core/files/resource_paths.cpp',
+    'src/core/input/key_chord.cpp',
+    'src/core/log.cpp',
+    'src/core/timer_manager.cpp',
+    'src/core/ui_phase.cpp',
+    'src/render/animation/animation.cpp',
+    'src/render/animation/animation_manager.cpp',
+    'src/render/animation/motion_service.cpp',
+    'src/render/scene/input_area.cpp',
+    'src/render/scene/input_area.cpp',
```
Ly-sec: Remove either of the two duplicate `'src/render/scene/input_area.cpp'`. Afterwards I gladly merge it.

borgsten: Didn't look after I sorted them. Done!

PR #3457, opened 2026-07-15
feat(bluetooth): also read battery percent from UPower

jinliu:
Some bluetooth devices (e.g. PlayStation 5 Controller) do not report battery stats over Bluetooth, but from UPower. So if the serial of a UPower device matches the address of a Bluetooth device, we use the former's battery percent.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Build / packaging

## Manual Coverage

- [x] Tested on Niri
- [ ] Tested on Hyprland
- [ ] Tested on Sway

## Checklist

### DIFF
--- a/src/app/application.h
+++ b/src/app/application.h
@@ -258,11 +258,12 @@ private:
   std::unique_ptr<NetworkSecretAgent> m_networkSecretAgent;
   ExternalIpService m_externalIpService{&m_httpClient, &m_configService};
   std::unique_ptr<IwdSecretAgent> m_iwdSecretAgent;
+  // Declared before m_bluetoothService so it outlives the raw pointer in that service.
+  std::unique_ptr<UPowerService> m_upowerService;
   std::unique_ptr<BluetoothService> m_bluetoothService;
   std::unique_ptr<BluetoothAgent> m_bluetoothAgent;
   Timer m_bluetoothResumeTimer;
   std::unique_ptr<PolkitAgent> m_polkitAgent;
-  std::unique_ptr<UPowerService> m_upowerService;
   std::optional<bool> m_notificationDaemonEnabled;
   bool m_notificationDaemonInitFailed = false;
   bool m_notificationShellRefreshScheduled = false;
--- a/src/app/application_services.cpp
+++ b/src/app/application_services.cpp
@@ -945,6 +945,9 @@ void Application::initSystemBusServices() {
       m_upowerService->setChangeCallback([this, shouldRefreshControlCenter]() {
         onUpowerStateChangedForHooks();
         m_batteryWarningMonitor.evaluate(m_configService.config().battery, *m_upowerService, m_notificationManager);
+        if (m_bluetoothService != nullptr) {
+          m_bluetoothService->refreshBatteryFromUPower();
+        }
         m_bar.refresh();
         m_settingsWindow.onExternalOptionsChanged();
         if (shouldRefreshControlCenter()) {
@@ -1069,7 +1072,7 @@ void Application::initSystemBusServices() {
     }
 
     try {
-      m_bluetoothService = std::make_unique<BluetoothService>(*m_systemBus);
+      m_bluetoothService = std::make_unique<BluetoothService>(*m_systemBus, m_upowerService.get());
       auto refreshBluetoothUi = [this, shouldRefreshControlCenter]() {
         m_bar.refresh();
         if (shouldRefreshControlCenter()) {
--- a/src/dbus/bluetooth/bluetooth_service.cpp
+++ b/src/dbus/bluetooth/bluetooth_service.cpp
@@ -904,6 +904,9 @@ bool BluetoothService::applyUPowerBattery() {
     }
 
     if (!bluetoothDevice.hasBattery) {
+      if (bluetoothDevice.address.empty()) {
+        continue;
+      }
       auto serial = bluetoothDevice.address;
       auto* deviceInfo = m_upowerService->deviceForSerial(serial);
       if (!deviceInfo) {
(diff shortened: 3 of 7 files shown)


ItsLemmy:
## 1. UPower-derived battery is never cleared → stale percentage persists forever

updateBatteryStatsFromUPower() only ever sets hasBattery = true. There is no path that clears it:

- Non-connected devices continue before the lookup, so a device that disconnects keeps its last UPower reading.
- A device that stays connected but whose UPower device disappears also keeps the stale value.
- BlueZ's own clearing path (onInterfacesRemoved → kBatteryInterface → hasBattery = false) never fires for these devices, because they never had Battery1 in the first place.

This is user-visible: bluetooth_tab.cpp:169 renders the battery pill on m_device.hasBattery alone, with no connected guard, so a paired-but-disconnected DS5 will show a frozen "72%" pill indefinitely.

The fix needs source tracking, because you can't blindly clear hasBattery — it may have come from BlueZ. Suggest adding a field to BluetoothDeviceInfo (e.g. bool batteryFromUPower = false) and restructuring:

```
for (auto& dev : m_devices) {
  const auto it = (dev.connected || dev.connecting)
      ? batteryDevices.find(StringUtils::toLower(dev.address))
      : batteryDevices.end();
  if (it != batteryDevices.end()) {
    const auto percent = static_cast<std::uint8_t>(std::lround(it->second.state.percentage));
    if (!dev.hasBattery || dev.batteryPercent != percent || !dev.batteryFromUPower) { ...set, hasChange = true; }
  } else if (dev.batteryFromUPower) {
    dev.hasBattery = false;
    dev.batteryPercent = 0;
    dev.batteryFromUPower = false;
    hasChange = true;
  }
}
```
Note BluetoothDeviceInfo uses = default equality, so a new field automatically participates in the updated != *dev comparisons in onPropertiesChanged — no other edits needed.

Non-blocking

## 2. emitDevices() is a hot path; the join allocates a lot per call

emitDevices() fires on every Device1 PropertiesChanged — during discovery that's an RSSI update per advertisement, many per second. Each one now:
- calls batteryDevices(), which returns a std::vector<UPowerDeviceInfo> by value (7 std::strings per element),
- copies each element again into an unordered_map,
- allocates a lowercased key string per UPower device and per BT device.

Cheap improvements, in order of value:
- Early-out before any work: if (m_upowerService == nullptr || m_devices.empty()) return; (the m_upowerService check already exists but sits after a debug log — see #4).
- Don't build a map. N is a handful on both sides; a nested loop with StringUtils::equalsIgnoreCase-style comparison avoids every allocation. If you keep the map, key it to double percentage, not a full UPowerDeviceInfo copy.
- Better still: add UPowerService::deviceForSerial(std::string_view) alongside the existing deviceForSelector() and skip batteryDevices()'s copy entirely.

## 3. hasChange is computed but unused on the emitDevicesDirty == false path

When called from emitDevices(), hasChange only gates a debug log — the caller emits unconditionally right after. The bool parameter is also a boolean trap at the call site (updateBatteryStatsFromUPower(true)). Cleaner shape:

// public — apply and notify if anything moved
void refreshBatteryFromUPower() { if (applyUPowerBattery() &&Callback(m_devices); }
// private — mutate m_devices, return whether anything changed
bool applyUPowerBattery();

emitDevices() then calls applyUPowerBattery() and ignores thenning via emitDevices()" comment becomes unnecessary becausethe structure makes it obvious.

## 4. Debug logging: too chatty, fires before the guard, wrong style

- kLog.debug("Updating battery stats from UPower service") runs before the !m_upowerService check, so it logs on every emitDevices() even when UPowedisabled.
- Two more lines log per device, per call — on the discovery path that's a debug flood.                                                             - Message style doesn't match the file: neighbours are lowerced -> {}", "GetManagedObjects failed: {}"). These arecapitalized sentences.                                                                                                                              
I'd cut all of these to a single line inside the hasChange branch, e.g. kLog.debug("upower battery for {} -> {}%", addr, percent). 

## 5. Callback ordering in application_services.cpp costs an extra refresh pass                                                                        
The new updateBatteryStatsFromUPower(true) is appended at the end of the UPower change callback, after m_bar.refresh() and m_panelManager.refresh().that refresh pass paints the old battery value, and the freshes-callback cascade a moment later — two full refreshes perUPower tick. Moving the bluetooth update to the top of the callback (before m_bar.refresh()) lets the existing refresh pick up the new value in one 

## 6. Smaller notes                                                                                                                                    
- static_cast<uint8_t>(std::round(...)) — prefer std::lround (or clamp to [0, 100]) so an out-of-range or NaN percentage can't produce UB in the    narrowing cast. The file already uses std::uint8_t in the hea is a minor inconsistency.
- batteryDevices() filters on info.isPresent. Worth double-checking that UPower reports IsPresent=true for the DS5 (some peripheral device types leafalse) — you tested it works, so presumably yes, but it's a sipheral doesn't set it.
- The filter isBatteryCapableDeviceType includes laptop batteries, so a laptop battery whose Serial happened to equal a MAC-looking string would be misattributed. Vanishingly unlikely; you could tighten it by () if you want belt-and-braces.
- Comment nit: // Must be before m_bluetoothService, as it references me — "me" is odd for a member comment. Suggest: // Declared before m_bluetoothService so it outlives the raw pointer that servic
- setUPowerService() is called after BluetoothService construction, which kicks off async GetManagedObjects. That reply lands on a later loop iteration so ordering is fine, but passing UPowerService* through the cendency structural rather than a timing assumption.

ItsLemmy:
Thanks for the PR, please confirm it still work with your dual-sense.

jinliu:
> Thanks for the PR, please confirm it still work with your dual-sense.

Can confirm. Thanks!

Issue #3186, opened 2026-06-27
[BUG] Audio volume slider missing in "Application Volumes" section during active playback in Telegram


blackcityskyline:
### Submission checklist

- [x] I have searched existing issues and confirmed this is not a duplicate.
- [x] I am using the latest available version of Noctalia.

### Bug description

When playing audio through a Telegram desktop client, the "Application Volumes" section in the audio control panel fails to display the corresponding volume slider for the application. The UI shows a "No application audio" message, indicating that no playback streams are available, despite audio playing correctly.

### Steps to reproduce

1. Launch the Telegram desktop app.
2. Play any audio message or audio file.
3. Open the Noctalia audio control panel.
4. Observe the "Application Volumes" section.

### Expected behavior

The "Application Volumes" section should display active audio streams from the Telegram client with a volume slider that allows adjusting the application's volume.

### Actual behavior

The section shows the message "No application audio" and does not list any active streams or a volume slider for the Telegram app

### Logs / error output

PipeWire logs
```text
id 64, type PipeWire:Interface:Client/3
    application.name = "Telegram"

id 66, type PipeWire:Interface:Node/3
    node.description = "Telegram"
    node.name = "Telegram"
    media.class = "Stream/Output/Audio"

id 68, type PipeWire:Interface:Link/3
    link.output.node = "66"
    link.input.node = "95"

id 71, type PipeWire:Interface:Link/3
    link.output.node = "66"
    link.input.node = "95"

id 95, type PipeWire:Interface:Node/3
    node.description = "Q5"
    node.name = "bluez_output.77_BC_9C_80_00_2C.1"
    media.class = "Audio/Sink"
```

### Noctalia version

v5.0.0 (362980a7f21a)

### Compositor

Hyprland

### Distribution

Arch-based

### Environment information

```text
Installation Method: Source build
GPU: IPGU Intel HD4000
Audio Backend: PipeWire
```

### Additional context

<img width="1366" height="768" alt="Image" src="https://github.com/user-attachments/assets/0539246b-4db2-4e67-be7a-5e6e843bf69c" />

https://github.com/user-attachments/assets/fc905580-376b-432e-a283-3634d3d9b540

Ly-sec:
Closing this one as solved, if you still encounter weird issues with it, please create a new issue.

PR #3473, opened 2026-07-16
fix(pipewire): show app streams with node.target in Application Volumes

ArturoAtomplay:
## Summary

`isProgramOutputNode` excluded every Stream/Output/Audio node that had
`target.object` set, except QEMU. apps like Sober (Roblox via SDL)
set `node.target` to route audio to a specific sink, so they never
appeared in the Application Volumes section.

The fix only skips nodes with `target.object` when `applicationName` is
also empty. apps carry a name, while loopback/filter endpoints
have no identity and stay filtered.

## Motivation

Sober's audio stream was invisible in the audio panel. The same pattern
(node.target + Stream/Output/Audio) likely affects other apps.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Build / packaging

## Related Issue

see #3186 for Telegram

## Testing

- `just build` (debug): succeeds, no new warnings.
- Manual: on Niri with PipeWire 1.6.7 / WirePlumber 0.5.15, Sober's
  audio stream previously did not appear in Application Volumes. After
  the fix it shows up.

## Manual Coverage

- [x] Tested on Niri
- [ ] Tested on Hyprland
- [ ] Tested on Sway

## Screenshots / Videos

image

## Checklist

## Additional Notes

Tested on Sober only. Telegram #3186 has the same root cause
but was not tested.

### DIFF
--- a/src/pipewire/pipewire_service.cpp
+++ b/src/pipewire/pipewire_service.cpp
@@ -696,7 +696,7 @@ namespace {
     if (!nd.linkGroup.empty() || nd.nodePassive) {
       return false;
     }
-    if (!nd.targetObject.empty() && !isQemuStreamNode(nd)) {
+    if (!nd.targetObject.empty() && nd.applicationName.empty() && !isQemuStreamNode(nd)) {
       return false;
     }
     return true;


Ly-sec:
Thanks for this, same root cause we hit for Telegram (#3186) / target.object.

We already pushed a fix on our side that does the same thing, with a slightly broader identity check (application.name or application.id or application.binary, plus the existing QEMU case) so streams that only get an id/binary still show up.

Closing this as superseded, but the diagnosis and approach were spot on. Appreciate the PR.

PR #3482, opened 2026-07-17
fix(input): don't drop wheel detent events; keep one wheel notch at one step

AlRowne:
## Summary

Two commits fixing wheel scrolling on discrete steppers (volume, brightness,
workspaces, ...), which currently apply 2–3 steps per wheel notch on
compositors that scale scroll deltas:

1. **`fix(input): don't drop wheel detent events in the seat handlers`** — the
   bug. `handlePointerAxisDiscrete()`/`handlePointerAxisValue120()` search the
   pending event queue *backwards* for the axis event to annotate, but the
   protocol guarantees the detent event is **followed** by its coupled axis
   event, so the search never matches and the detent is always discarded. The
   detent is now stashed per axis and claimed by the next axis event of that
   axis; leftovers are cleared at the frame boundary. Also fixes the legacy
   wheel fallback unit: `kLegacyWheelAxisUnitsPerStep` was 10.0, but libinput
   reports one detent as 15 degrees.

2. **`fix(input): cap wheel scrollSteps() at one step per frame`** — a design
   call, separable from 1. With the seat fixed, `scrollSteps()` sees the
   compositor's real detent count — which is not 1 per notch under a scroll
   factor (niri at 1.6 honestly reports alternating 1/2 detents per notch,
   giving an alternating 5%/10% volume pattern). A wheel notch is a hardware
   detent the user feels exactly once, so wheel sources are now capped at one
   step per frame. Free-spinning hi-res wheels emit sub-detent frames and
   still accrue to a full detent first; touchpads keep accumulating as before;
   scrolling content is untouched (it uses `scrollDelta()`, so a compositor
   scroll-factor still speeds up scroll views and scrollbars).

If you'd rather have `scrollSteps()` report the compositor's scaled detent
count faithfully, commit 1 on its own is still a strict improvement — happy to
drop commit 2.

## Motivation

Since 53c98d38b (`fix(input): quantize touch pad scroll into wheel-detent
steps`), one wheel notch on the volume widget changes the volume by 10–15%
with a configured 5% scroll step. The regression is not in the widgets — the
new quantizer is the first consumer that multiplies by the reported magnitude,
which exposed the two seat-level flaws above (the pre-53c98d38b widget code
only ever looked at the scroll direction, masking both).

Raw events captured with a minimal client binding `wl_seat` v5 (as Noctalia
does), niri 26.04 with `scroll-factor 1.6`, one physical notch per frame:

```
  axis_source src=0                  (WL_POINTER_AXIS_SOURCE_WHEEL)
  axis_DISCRETE axis=0 discrete=1    <-- arrives BEFORE the axis event
  axis        axis=0 value=24.000    <-- 15.0 (libinput detent) * 1.6
FRAME ---
```

The `axis_discrete` line always precedes its `axis` line — per spec: *"each
axis_discrete event is always followed by exactly one axis event with the same
axis number within the same wl_pointer.frame"* — so the backwards search can
never find its target. With the detent dropped, the legacy fallback divides
24.0 by 10.0 → 2.4 detents per notch → the accumulator emits 2–3 steps:

| Notch | accum before | steps | volume delta (5% step) |
|-------|--------------|-------|------------------------|
| 1     | 0.0          | 2     | 10%                    |
| 2     | 0.4          | 2     | 10%                    |
| 3     | 0.8          | 3     | 15%                    |

Note: `handlePointerAxisValue120()` has the same ordering flaw. It is
unreachable today (`kSeatVersion = 5`; `axis_value120` is v8+) but becomes
live the moment the seat version is bumped, so it is fixed alongside.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Build / packaging

## Related Issue

None — analysis and fix in one go; the full event captures are above.

## Testing

- Verified the event ordering with a minimal `wl_seat` v5 client against
  niri 26.04 (dump excerpt above).
- Replayed the measured event sequences through the exact final
  `dispatchAxis()` logic in a standalone harness: ratcheted wheel at
  scroll-factor 1.6 and 1.0 → exactly 1 step per notch; hi-res free-spin
  (0.25 lines/frame) → 1 step per 4 frames; direction reversal resets the
  accumulator; touchpad path byte-identical to before.
- Built and ran the shell on CachyOS / niri 26.04 (AMD RX 9070 XT), daily use:
  volume, brightness, workspaces, taskbar, calendar, settings sliders and
  number inputs step once per notch; settings/control-center/notification
  scroll views still follow the compositor scroll-factor; horizontal tilt
  unchanged.
- Both commits build independently with zero warnings; `meson test`: 40/40 OK
  at each commit. clang-format clean.

## Manual Coverage

- [x] Tested on Niri
- [ ] Tested on Hyprland
- [ ] Tested on Sway

Only tested on Niri 26.04 — that is the compositor I run. The seat-level fix
is compositor-independent protocol handling, but a confirmation on Hyprland or
Sway (both send `axis_discrete` on `wl_seat` v5) would be welcome.

## Screenshots / Videos

Input-behavior change only, nothing visual to show. The raw event captures in
Motivation stand in for a recording.

## Checklist

## Additional Notes

- The two commits are deliberately separable — see Summary. Dropping commit 2
  keeps commit 1 a strict improvement.
- Reviewer heads-up: with the seat fix, scroll views and scrollbars receive
  the compositor's *true* detent count. Users running a compositor
  scroll-factor were getting an inflated 2.4x line rate from the broken legacy
  fallback, so panel scrolling will feel somewhat slower for them after this —
  that is the compositor setting being honored, not a regression.

### DIFF
--- a/src/render/scene/input_area.cpp
+++ b/src/render/scene/input_area.cpp
@@ -14,6 +14,10 @@ namespace {
   // step deliberately rather than racing the finger.
   constexpr float kScrollUnitsPerStep = 20.0f;
 
+  bool isWheelSource(std::uint32_t axisSource) noexcept {
+    return axisSource == WL_POINTER_AXIS_SOURCE_WHEEL || axisSource == WL_POINTER_AXIS_SOURCE_WHEEL_TILT;
+  }
+
 } // namespace
 
 InputArea::InputArea() : Node(NodeType::Base) {}
@@ -213,9 +217,13 @@ bool InputArea::dispatchAxis(
     return false;
   }
 
-  // Quantize scroll into whole detent steps. Wheel events carry detents in
-  // axisLines and cross the threshold immediately; continuous sources
+  // Quantize scroll into whole detent steps. Wheel sources are capped at one
+  // step per frame: a ratcheted wheel emits one frame per notch, so the notch
+  // the user feels stays one step even when the compositor scales the delta
+  // (niri's scroll-factor), while free-spinning hi-res wheels emit sub-detent
+  // frames that must first accrue to a full detent. Continuous sources
   // (touchpads) accrue axisValue until a detent-equivalent is reached.
+  // Scrolling content stays on scrollDelta() and keeps the scaling.
   float axisSteps = 0.0f;
   if (axis < m_scrollStepAccum.size()) {
     float& accum = m_scrollStepAccum[axis];
@@ -226,6 +234,9 @@ bool InputArea::dispatchAxis(
     accum += detentDelta;
     axisSteps = std::trunc(accum);
     accum -= axisSteps;
+    if (isWheelSource(axisSource)) {
+      axisSteps = std::clamp(axisSteps, -1.0f, 1.0f);
+    }
   }
 
   return m_onAxis(
--- a/src/render/scene/input_area.h
+++ b/src/render/scene/input_area.h
@@ -42,9 +42,12 @@ public:
     }
 
     // Whole wheel-detent steps accumulated by the InputArea (positive = scroll
-    // down). Continuous sources (touchpads, hi-res wheels) emit a step only
-    // once a full detent-equivalent has accrued — use this instead of
-    // scrollDelta() for discrete stepping (volume, workspace cycling, ...).
+    // down). Wheel sources yield at most one step per frame, so a ratcheted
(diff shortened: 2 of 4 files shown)


Ly-sec:
Hey @AlRowne this looks good to me, thank you very much!

PR #3504, opened 2026-07-18
brightness: avoid blocking shutdown on ddcutil

Ape:
When my display port monitor was turned off, ddcutil could block and delay noctalia shutdown until its timeout. Cancel active DDC work when the brightness service is destroyed.


ItsLemmy:
Cool, thanks

PR #3515, opened 2026-07-18
feat(plugins): add noctalia.systemStats and per-core CPU sampling

ioandev:
## Summary

Exposes the system monitor to plugin scripts, and teaches `SystemMonitorService` to sample per-core CPU usage.

`SystemMonitorService::readCpuTotals()` only ever read the first line of `/proc/stat` — the aggregate `cpu` row — and none of its stats were reachable from Luau (`grep cpu src/scripting/` returned nothing). A plugin therefore could not draw anything per-core.

Sampling:

- Parse the per-core `cpuN` rows into `SystemStats::cpuCoreUsagePercent`.
- Share the idle/total jiffy maths between the aggregate and per-core paths via `parseCpuStatLine()` / `cpuUsageBetween()`, rather than duplicating it.
- Run per-core on its own fixed 1s deadline, refcounted behind `retainCpuCores()` / `releaseCpuCores()`, mirroring the existing CPU-temp gate. It is deliberately independent of `cpu_poll_seconds` (default 2s): consumers want per-second resolution, and pinning it leaves aggregate CPU behaviour untouched whatever the user configures. Costs one extra `/proc/stat` read per second, and only while a consumer holds a reference.
- Reseed rather than index across mismatched vectors when the core count changes (hotplug / offlining).
- Keep per-core out of the 120-deep history ring: the graphs plot the aggregate, and `history()` consumers would otherwise see a series populated only while someone holds a reference.

API:

- `noctalia.systemStats()` returns `cpu` (total + cores), `ram`, `swap`, `cpuTempC`, `gpu`, `net` and `loadAvg`. Absent sensors are `nil` rather than `0`, so a plugin can tell "no probe" from "idle". The first call opts the host into per-core sampling; the reference is released in `~LuauHost`.
- `noctalia.nowMs()` returns wall-clock milliseconds. `os.time()` and `noctalia.formatTime()` are both whole-second, so this is the only way a plugin can see sub-second time — needed to phase updates onto a second boundary.
- `ScriptApiContext` holds the `SystemMonitorService`. Unlike the hooks beside it, this is read straight from the script worker thread: `latest()` is mutex-guarded and returns a copy, so no main-thread marshalling is needed.

Bumps `kCurrentPluginApiVersion` to 5. `kOldestSupportedPluginApiVersion` stays at 3, so existing plugins are unaffected.

8 files, +271 / −18.

## Motivation

The bar can already show CPU usage, but only as a single averaged number. That hides exactly what people want a CPU meter for: one pegged thread, or an unbalanced load across cores, look identical to mild uniform load once averaged.

Per-core data did not exist anywhere in the codebase, and the plugin API had no access to system stats at all, so this could not be solved in a plugin. This adds the smallest host-side capability that makes it possible, in the shape the rest of the service already uses (refcounted opt-in sampling, snapshot reads).

The concrete consumer is a `cpu-bars` community plugin that draws one vertical bar per core; `systemStats()` was deliberately made general rather than CPU-only so it is not single-use.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Refactoring
- [ ] Build / packaging

Refactoring is ticked because the shared `parseCpuStatLine()` / `cpuUsageBetween()` helpers replace the inline maths on the **existing** aggregate path — see Additional Notes for the one behavioural difference that introduces.

## Related Issue

None.

## Testing

Commands run:

- `just build` — clean, no new warnings.
- `meson test -C build-debug --print-errorlogs` — **40/40 pass**, including the updated `plugin_manifest` test (5 supported, 6 too new).
- `just format` — clang-format 22.1.6.
- `noctalia config validate` — clean.

Correctness of the per-core values was checked against ground truth rather than eyeballed. Load was pinned to known cores with `taskset` and the reported vector inspected:

- Cores 3 and 17 pinned → exactly `100` at 1-based positions 4 and 18, everything else near zero. Confirms both the values and the `/proc/stat` index ordering.
- Repeated for cores 0-3 + 12-13, and 6/7/15.
- Varied duty cycles via `stress-ng --cpu-load` (100% / 75% / 55% / 25%) → bar heights tracked the requested load.

Driven end-to-end two ways: a throwaway Luau `[[service]]` plugin logging `systemStats()` once a second, then the real `cpu-bars` bar widget. Also exercised:

- `nowMs()` used for tick alignment — measured landing `.004`–`.016` past the second.
- Theme changes mid-run (the palette is wallpaper-derived here) — role tokens tracked.
- Plugin reload / disable, to confirm the per-core reference is released.

Not covered: `cpu_poll_seconds` values other than the default; CPU hotplug (the mismatched core-count path is reasoned-through, not exercised); the GPU/VRAM fields of `systemStats()` were never observed non-`nil` on this machine.

## Manual Coverage

- [x] Tested on Niri
- [ ] Tested on Hyprland
- [ ] Tested on Sway

Only Niri, on a single 3440x1440 output at scale 1, with a bottom bar. This PR adds no UI of its own, so bar position / density / scaling / multi-monitor were not exercised — the consuming plugin renders through the existing declarative `ui.*` path, which those settings already cover.

## Screenshots / Videos

No user-visible change in the shell itself; this is a plugin-facing API.

## Checklist

Notes on the above: no new settings, config keys or translation strings are added — the two new bindings are plugin-facing only. Plugin API docs live in the separate docs.noctalia.dev repo and still need writing for `noctalia.systemStats` and `noctalia.nowMs`.

## Additional Notes

**One behavioural difference on the existing aggregate path.** Routing it through `cpuUsageBetween()` adds two things it did not have: the result is clamped to `[0, 100]`, and the idle delta is guarded against underflow (`current.idle >= prev.idle ? ... : 0`) instead of subtracting raw. Both are no-ops in normal operation — `idleDelta <= totalDelta` holds — but they are a real diff on a hot path, so worth a reviewer's eye rather than being buried in a refactor.

**Why per-core does not follow `cpu_poll_seconds`.** Tying it to the existing CPU deadline would mean either changing the 2s default for everyone (a battery regression for a feature most users will not enable) or shipping a per-core meter that updates every 2s. The fixed 1s deadline is only live while something holds a reference, so the cost is opt-in.

**Why `ScriptApiContext` holds a raw service pointer.** Every other member there is a main-thread-only hook, so this deviates. The alternative — pushing a snapshot from the main thread like `setOutputs()` — would add a copy and a publish cadence for data that is already mutex-guarded and cheap to read off-thread. The member is commented to say so, since the surrounding convention implies otherwise.

**Known limitation.** `systemStats()` returns a snapshot at whatever cadence each metric is configured for; the per-core vector is empty for the first tick after a consumer attaches, because there is no previous sample to diff against. Plugins should treat an empty `cpu.cores` as "not ready yet" rather than "no cores".

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1221,6 +1221,21 @@ cpu_temp_sensor_test = executable('cpu_temp_sensor_test',
 
 test('cpu_temp_sensor', cpu_temp_sensor_test)
 
+cpu_stat_parse_test = executable('cpu_stat_parse_test',
+  sources: files(
+    'tests/cpu_stat_parse_test.cpp',
+    'src/system/system_monitor_service.cpp',
+    'src/system/cpu_temp_sensor.cpp',
+    'src/system/intel_gpu.cpp',
+    'src/system/format_units.cpp',
+    'src/core/log.cpp',
+  ),
+  include_directories: _noctalia_inc,
+  dependencies: [dl_dep],
+)
+
+test('cpu_stat_parse', cpu_stat_parse_test)
+
 i18n_language_tag_test = executable('i18n_language_tag_test',
   sources: files(
     'tests/i18n_language_tag_test.cpp',
--- a/src/app/application_services.cpp
+++ b/src/app/application_services.cpp
@@ -770,6 +770,7 @@ void Application::initAuxServicesAndHooks() {
 
   try {
     m_systemMonitor = std::make_unique<SystemMonitorService>(m_configService.config().system.monitor);
+    m_scriptApi.setSystemMonitor(m_systemMonitor.get());
     if (m_systemMonitor->isRunning()) {
       kLog.info("system monitor service active");
     } else {
--- a/src/scripting/luau_host.cpp
+++ b/src/scripting/luau_host.cpp
@@ -17,6 +17,7 @@
 #include "system/app_identity.h"
 #include "system/desktop_entry.h"
 #include "system/icon_resolver.h"
+#include "system/system_monitor_service.h"
 #include "system/terminal_launch.h"
 #include "time/time_format.h"
 #include "ui/dialogs/color_picker_dialog.h"
@@ -175,6 +176,20 @@ namespace {
     lua_setfield(L, -2, key);
   }
 
+  void setTableNumber(lua_State* L, const char* key, double value) {
+    lua_pushnumber(L, value);
+    lua_setfield(L, -2, key);
+  }
+
+  // Pushes an optional as a number, or leaves the key absent (nil) when it has no value —
+  // the Luau-idiomatic way to say "this machine has no such sensor".
+  template <typename T> void setTableOptionalNumber(lua_State* L, const char* key, const std::optional<T>& value) {
+    if (!value.has_value()) {
+      return;
+    }
+    setTableNumber(L, key, static_cast<double>(*value));
+  }
+
   void setTableString(lua_State* L, const char* key, const std::string& value) {
     lua_pushlstring(L, value.data(), value.size());
     lua_setfield(L, -2, key);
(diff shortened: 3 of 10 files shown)


ItsLemmy:
Branch have drifted to much for a proper merge.
Re-implemented properly in b99f269a13fb13440cd6e15292e5f8e4d45eb5a8

PR #1548, opened 2026-01-25
Framed Bar

tibssy:
# Pull Request

This PR introduces a new "Framed Bar" style to Noctalia.

## Motivation

This style renders a border frame around the screen edges, offering a distinct aesthetic alternative to the existing "Simple" and "Floating" bars. It is fully integrated into the system, including:

- Settings Panel: Option to switch to "Framed" in the Bar Appearance tab.
- Layout Logic: Correctly handles exclusion zones and window placement.
- Setup Wizard: Added as a choice during the initial setup customization step.

## Type of Change
Mark the relevant option with an "x".
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring

## Related Issue
- Closes #(issue number) (if any)

## Testing
Describe how you tested your changes and mark the relevant items.

- [x] Tested on niri
- [x] Tested on Hyprland
- [ ] Tested on sway
- [x] Tested with different bar positions and density settings
- [ ] Tested at different interface scaling values
- [x] Tested with multiple monitors (if applicable)

## Screenshots / Videos
Image

https://github.com/user-attachments/assets/51756811-c6ab-45f7-9eb4-379602816e73

## Checklist
- [x] Code follows project style guidelines
- [x] Self-reviewed my code
- [x] No new warnings or errors
- [x] Documentation or comments updated (if relevant)

## Additional Notes
Add any additional context or follow-up notes for reviewers.

PR #3524, opened 2026-07-19
feat(frame): solid inset border around the output with a rounded desktop hole

ioandev:
## Summary

Adds a "screen frame": a solid border drawn around each output with a rounded rectangular hole in the middle that the desktop shows through. The edge that already carries a bar is inset by the bar's own thickness rather than the frame's, and the hole supplies the rounded inner corners for all four sides, so the frame and the bar read as one continuous shape rather than two strips meeting at a corner.

New module `src/shell/frame/`, a new `[shell.frame]` config section, and three settings entries under Desktop. No changes were needed under `render/` — the frame is built from the existing `Box` primitive plus the existing `ScreenCorner` node, whose shader already fills *outside* a superellipse, which is exactly one inner corner of a frame-with-hole.

Also changes `screen_corners` from the `Top` layer to `Overlay`. Those masks model the physical shape of the display, so nothing the shell draws should sit on top of them. On `Top` they competed with the bar, dock and frame purely by creation order, and any partial rebuild silently reordered them.

## Motivation

This existed in v4 as the **"Framed" bar type** (`barType = "framed"`, `frameThickness`, `frameRadius`) and was dropped in the C++ rewrite. It landed in v4 as PR #1548 (`680dfccff`, `dba523592`); none of those commits are ancestors of v5 `main`, so the feature was never deleted — it simply was not ported.

The port is deliberately not a transcription. v4 drew one full-screen `ShapePath` with an even-odd fill hole under Qt's `CurveRenderer`; v5 has no such primitive, so this uses the scene graph instead. It also does not reproduce v4's habit of swallowing clicks on the frame rails, nor its per-screen mask bugs.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [x] Build / packaging

## Related Issue

None.

## Testing

```
just format                 # clean
meson compile -C build-debug  # clean, no new warnings
meson test -C build-debug   # 43/43 pass
python3 tools/i18n-check.py # OK: 788 tr keys — all present in catalog
/repos/noctalia.sh          # built, hot-swapped, 4 frame surfaces up, no errors in log
```

New unit test `tests/frame_geometry_test.cpp` (10 groups). The frame's geometry is extracted into `src/shell/frame/frame_geometry.h` as pure functions so it is testable without a compositor, renderer or live bar — the same pattern `taskbar_workspace_label.h` uses. `frame.cpp` delegates to it rather than duplicating it, so the tests exercise shipping code. Cases cover: bar-edge insets (bar thickness, not bar + frame), reserved-but-empty vs released bands, `match_bar` opacity being independent of `reserveSpace`, rails never entering the bar's band, end caps sized by the reserved inset, exhaustive disjointness across 5 edges x 3 bleed values, and degenerate output sizes. Every case corresponds to a defect that actually occurred during development; the comments name the symptom.

Extended `tests/config_schema_roundtrip_test.cpp` and `tests/config_path_resolution_test.cpp` for the six new config keys.

Manual verification was done by pixel probe against a flat wallpaper rather than by eye — several defects in this work were initially misdiagnosed from screenshots, so the numbers are the evidence:

- Rail, end cap and bar render identically at `(17, 92, 6)`; the cap read `(23, 38, 8)` before the compositor blur region was added.
- Inner corners continuous across the frame/bar junction, all four verified.
- `bleed` closes the compositor gap: frame runs to the window border with no wallpaper between.
- The opacity slider's `visibleWhen` was confirmed in the running settings window (5 controls with Match Bar on, 6 with it off), not assumed.

## Manual Coverage

- [x] Tested on Niri
- [ ] Tested on Hyprland
- [ ] Tested on Sway

Only a bottom bar was exercised live, on a single 3440x1440 output at scale 1. All four bar positions are covered by unit tests, but not on a real compositor. Scaling and multi-monitor were not tested at all — see Additional Notes.

## Screenshots / Videos

screenshot-frame

## Checklist

This should be marked **Draft**. The known gaps below are not polish items; two of them are wrong behaviour that a user with more than one bar will hit immediately.

### Config

`[shell.frame]` — `enabled` (bool, false), `thickness` (int, 8, 1-64), `radius` (int, 12, 0-64), `match_bar` (bool, true), `opacity` (float, 1.0, 0-1), `bleed` (int, 0, 0-64). Settings UI under **Desktop → Screen Frame**; the opacity slider is hidden while Match Bar is on, since it has no effect then.

### Three invariants worth preserving

Each caused a visible bug when violated, so they are stated as invariants in the code and pinned by tests:

1. **The bar owns its reserved band; the frame owns everything outside it.** A rail crossing that band hides the bar's widgets outright whenever the frame happens to stack above the bar.
2. **End caps are sized by the reserved inset, never the painted one.** Sizing them with `thickness + bleed` reaches past the bar's edge and clips its outermost widget.
3. **The frame requests the same compositor blur as the bar.** The bar calls `setBlurRegion`, so its translucent fill sits on a blurred backdrop and stays flat across a patterned wallpaper. Without the same request the frame's fill tracks the wallpaper and visibly drifts from the bar's colour where the two meet.

### Known gaps

- **Multiple bars are handled incorrectly.** `styleBar()` returns the first enabled bar, so the frame cedes a single edge and both reserves and paints over any other bar. With a left or right bar added alongside a bottom one, that bar renders behind the frame. `holeInsets`, `rails` and `endCaps` are pure functions taking a single edge; the fix is to make them take a set of edges, and the existing single-bar test cases extend to that directly. This is the most important follow-up.
- **Enabling a bar at runtime leaves it behind the frame.** Toggling a bar on sets `changed.bars`, which rebuilds the frame; wlr-layer-shell stacks same-layer surfaces by creation order with no way to restack, so the frame lands on top. A plain config reload does not rebuild the frame and the bar correctly stays above, which is what isolates this to ordering rather than geometry. Registering the frame's reload callback before `Bar::initialize` did not resolve it and the cause is not yet established.
- **Per-monitor bar overrides are ignored,** and multi-monitor is untested — only one output was available. v4 hit the same bug (`b3fe8930c`).
- **`bleed` is a single value for all four edges,** so asymmetric compositor gaps cannot be fully covered. Concretely: niri `gaps 14` with `struts { left -10 right -10 top -10 bottom 0 }` yields a 4px gap on three edges and 14px on the fourth. Per-edge bleed would fix it.
- **Auto-hide is not animated in step with the bar.** The frame switches its bottom edge on a discrete visibility transition while the bar slides over an animation, so it snaps rather than following. Painting that edge unconditionally does sync perfectly, but only if the frame is reliably stacked below the bar, which it is not — that attempt put the frame's rail on top of the bar and was reverted.
- **An IPC `bar-hide` leaves that edge unframed.** Releasing the exclusive zone hands the band to windows, so the frame correctly cedes it, but then draws no rail there because `holeInsets()` still sizes that edge from the bar's thickness. Cosmetic.
- **Outer corners are square.** Rounded display corners remain the separate `[shell.screen_corners]` feature. This matches v4, where the framed outer contour used a 0.01px radius and screen rounding was a distinct module.

### What the frame cannot fix

Compositors apply their own gap *inside* whatever exclusive zone the shell reserves (niri `layout.gaps`, Hyprland `general:gaps_out`), leaving wallpaper between the frame and the nearest window. Window placement belongs to the compositor and a shell cannot prevent it; the frame sits above windows, so `bleed` paints over it. Gaps *between* windows are not reachable at all — the frame has no surface there, and that space is meant to show the desktop.

### Stacking order

wlr-layer-shell stacks same-layer surfaces by creation order. Two places must agree and nothing enforces it: `reconcileOutputSurfaces()` in `application_services.cpp`, and reload-callback registration order in `application_ui.cpp`. Moving `screen_corners` to `Overlay` removes the corners from that race entirely, which is why it is included here. Routing reloads through the same single ordering used by `reconcileOutputSurfaces()` would remove the remaining class of bug, including the runtime bar-enable case above, and is the natural follow-up.

### DIFF
--- a/assets/translations/en.json
+++ b/assets/translations/en.json
@@ -1105,6 +1105,7 @@
         "filtering": "Filtering",
         "focus-styling": "Focus Styling",
         "formats": "Formats",
+        "frame": "Screen Frame",
         "general": "General",
         "greeter": "Greeter",
         "grouping": "Grouping",
@@ -1726,6 +1727,30 @@
         }
       },
       "desktop": {
+        "frame-bleed": {
+          "description": "Extra pixels the frame paints inward, to cover the gap your compositor leaves around windows (niri layout.gaps, Hyprland general:gaps_out)",
+          "label": "Gap Coverage"
+        },
+        "frame-enabled": {
+          "description": "Draw a solid border around the screen with a rounded hole for the desktop; the edge holding a bar uses the bar's own thickness",
+          "label": "Screen Frame"
+        },
+        "frame-match-bar": {
+          "description": "Use the bar's background opacity for the frame so the two read as one surface; turn off to set the frame's opacity separately",
+          "label": "Match Bar"
+        },
+        "frame-opacity": {
+          "description": "Opacity of the frame border",
+          "label": "Frame Opacity"
+        },
+        "frame-radius": {
+          "description": "Corner radius of the frame's inner opening in pixels",
+          "label": "Inner Radius"
+        },
+        "frame-thickness": {
+          "description": "Width of the frame border in pixels, on the edges that do not hold a bar",
+          "label": "Frame Thickness"
+        },
         "hot-corners-bottom-left": {
           "description": "Action when the pointer hits the bottom left corner",
           "label": "Bottom Left Corner"
--- a/meson.build
+++ b/meson.build
@@ -641,6 +641,7 @@ _noctalia_sources = files(
   'src/shell/tooltip/tooltip_manager.cpp',
   'src/shell/hot_corners/hot_corners.cpp',
   'src/shell/screen_corners/screen_corners.cpp',
+  'src/shell/frame/frame.cpp',
   'src/shell/panel/panel_click_shield.cpp',
   'src/wayland/hyprland/focus_grab_service.cpp',
   'src/shell/panel/panel_manager.cpp',
@@ -1527,6 +1528,27 @@ config_path_resolution_test = executable('config_path_resolution_test',
 )
 
 test('config_path_resolution', config_path_resolution_test)
+
+frame_geometry_test = executable('frame_geometry_test',
+  sources: files(
+    'tests/frame_geometry_test.cpp',
+    'src/config/config_types.cpp',
+    'src/render/core/color.cpp',
+    'src/ui/palette.cpp',
+    'src/theme/builtin_palettes.cpp',
+    'src/theme/fixed_palette.cpp',
+    'src/theme/color.cpp',
+    'src/theme/contrast.cpp',
+    'src/core/input/key_chord.cpp',
+  ),
+  include_directories: _noctalia_inc,
+  dependencies: [
+    xkbcommon_dep,
+    tomlplusplus_dep,
+  ],
+)
+
+test('frame_geometry', frame_geometry_test)
 
 config_validate_env = environment()
 config_validate_env.set('NOCTALIA_CONFIG_HOME', meson.current_source_dir() / 'tests/config_validate/config-home')
--- a/src/app/application.h
+++ b/src/app/application.h
@@ -35,6 +35,7 @@
 #include "shell/bar/bar.h"
 #include "shell/desktop/desktop_widgets_controller.h"
 #include "shell/dock/dock.h"
+#include "shell/frame/frame.h"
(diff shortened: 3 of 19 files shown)


ItsLemmy:
No thanks.

PR #3532, opened 2026-07-19
feat(lockscreen,polkit): face authentication via Gaze

### LINKED ISSUE #3000: [Feature Request] howdy face login support on lockscreen
### Submission checklist

### Bug description

Frontend UI for howdy face login on lockscreen.
Option to make a default way of unlock by password, fingerprint or face unlock.
Frontend UI like Windows to select unlock way.

### Steps to reproduce

1. Open Lockscreen
2. Click on Face Unlock
3. Starts Scanning

### Expected behavior

1. Open Lockscreen
2. Click on Face Unlock
3. Show frontend UI for Face Scanning and Unlock

### Actual behavior

No such options.

### Logs / error output

```text

```

### Noctalia version

v5.0.0.r1191.g39a4a335c-11

### Compositor

Niri

### Distribution

Arch-based

### Environment information

```text
Installation method: AUR
```

### Additional context

_No response_

bennypowers:
## Summary

- Add face unlock to the lockscreen using [GunduLabs/gaze](https://github.com/GunduLabs/gaze) facial authentication daemon via D-Bus (`com.gundulabs.Gaze`)
- Add polkit auto-response for `GAZE_CONFIRMATION_REQUEST` from `pam_gaze.so`, enabling face auth through the polkit PAM stack
- Add `face-auth` config toggle and settings UI entry under Security > Lock Screen

## Details

`FaceAuthenticator` mirrors the existing `FingerprintAuthenticator` pattern:

- Pre-checks `HasEnrolledFaces` and `IsCameraAvailable` before claiming
- Async `Claim`/`Release` session lifecycle with `VerifyStart`/`VerifyStop`
- Subscribes to `FaceStatus` (capture quality feedback) and `VerifyStatus` (match result) signals
- Retries up to 3 times on `verify-no-match` with 500ms delay
- Handles sleep/wake via `PrepareForSleep` with deferred proxy reclaim for stale claims
- Runs concurrently with fingerprint auth; whichever succeeds first unlocks

When Gaze is not installed or no face is enrolled, face auth silently bails with no errors.

For polkit, when `pam_gaze.so` is in the polkit PAM stack and sends a `GAZE_CONFIRMATION_REQUEST` prompt on successful face match, noctalia auto-responds with `CONFIRM`.

## Test plan

- [ ] Build and install with Gaze not installed: lockscreen and polkit work normally, no errors in log
- [ ] Build with Gaze installed, face enrolled: lockscreen shows "Look at the camera", unlocks on face match
- [ ] Verify face auth retries on no-match, stops after 3 failures
- [ ] Verify face auth stops when user starts typing password, restarts after failed password attempt
- [ ] Verify sleep/wake: face auth restarts after resume
- [ ] Verify settings toggle disables face auth
- [ ] Verify polkit face auth with `pam_gaze.so` in `/etc/pam.d/polkit-1`
- [ ] Verify concurrent fingerprint + face auth both work, first to succeed unlocks

Closes #3000


bennypowers:
Currently, when the lockscreen comes up, a gaze session starts immediately, runs the CPU until timeout, then cancels, and subsequently falls back to password authentication. Not ideal. I'd like to get some feedback on this for a better workflow.

TotalyEnglizLitrate:
you could mirror the hyprlock approach for fprintd (atleast from back when I used it), waiting for an empty password entry to start fprintd auth. I believe the whole drive fprintd over dbus instead of pam exists solely to avoid that semi-hacky workaround, but it seems like a reasonable UX decision in this case?

P.S. this introduced me to gaze so thanks for that!

jinliu:
> you could mirror the hyprlock approach for fprintd (atleast from back when I used it), waiting for an empty password entry to start fprintd auth. I believe the whole drive fprintd over dbus instead of pam exists solely to avoid that semi-hacky workaround, but it seems like a reasonable UX decision in this case?

Yes, for `sudo` and `polkit`, we do need a manual confirmation here. Probably don't want it to automatically succeed (possibly invoked in background by some bad script) whenever I sit in front of my PC.

But instead of starting facial verification only after the user presses Enter, maybe a better approach is to start facial verification immediately, and upon success, hide the password box, so the user can just press Enter to continue.

Lockscreen is similar. We can start Gaze on screen-on and any user interaction (keyboard/mouse), and hide the password box on verification success.

Fingerprint / SmartCard is different in that, swiping is already an explicit action from the user, so we don't need any extra confirmation.

TotalyEnglizLitrate:
Face unlock works after a recent update? just noticed now. I just have the grosshack pam from gaze. I don't use the polkit agent provided by noctalia so idk if that works as well.
I can do the following to unlock
a. use the fingerprint
b. trigger face unlock by any password entry (doesn't have to be correct)
c. fallback to fprint if no face match, then password if no fprint match (this is just the gaze -> fprint -> password pam ordering afaict)

edit: tested with polkit agent, that works too. I'll switch to the non-grosshack pam module to see if that works as well, it should.
edit 2: yep, works without the grosshack gaze pam

ItsLemmy:
Is this PR still needed then?

TotalyEnglizLitrate:
It adds the need to hit enter to avoid unintentional authentication/authorization, but apart from that it's pretty much redundant afaict.

computer-cat:
It would be helpful to add for users without a physical keyboard, I have a Flow Z13 with a detachable keyboard so it would be nice if there was an integrated flow for it. Currently I have to connect the keyboard to type something (doesn't need to be correct) into the password box, otherwise it won't trigger gaze.

PR #3538, opened 2026-07-19
feat(screenshot): add Ctrl+C and Ctrl+S shortcuts for region capture

### LINKED ISSUE #3359: [FEATURE] CTRL+C to copy screenshot and CTRL+S to save screenshot
### Submission checklist

### Feature type

New functionality

### Feature summary

If you activate the option to press a button to confirm the screenshot, it should give the option to copy to the clipboard with CTRL+C or save into a default folder with CTRL+S.

### Motivation / use case

The need to keep changing the option to save into a folder or copy to clipboard through the settings every time you need one or another is time consuming for a simple task.

### Proposed solution

Give the option to press CTRL+C to copy to clipboard and CTRL+S to save into a default folder.

### Alternatives considered

_No response_

_No response_

### Additional context

_No response_

oldirtty:
## Summary

Added `Ctrl+C` and `Ctrl+S` keybinds to the interactive screenshot region overlay. 
When `confirmRegion` is active:
- `Ctrl+C` forces the output to be copied to the clipboard (ignoring the default config).
- `Ctrl+S` forces the output to be saved to a file (ignoring the default config).
- `Enter` or `Space` keeps the default configured behavior.
Also updated the English translation string for the region hint to display these new shortcuts.

## Motivation

It solves the friction of having to constantly change the `noctalia` configuration just to switch between saving a screenshot to a file or copying it to the clipboard. Users can now decide the output behavior on the fly during the region selection.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Build / packaging

## Related Issue

Closes #3359

## Testing

Compiled locally and tested the region capture through widget and keybind (`confirmRegion = true` is required). 
- Verified that pressing `Ctrl+C` copies the image and ignores the save-to-file config.
- Verified that pressing `Ctrl+S` triggers the file save notification and ignores the clipboard.
- Verified that `Enter` properly respects the default output config.

## Manual Coverage

- [x] Tested on Niri
- [ ] Tested on Hyprland
- [ ] Tested on Sway

## Screenshots / Videos

https://github.com/user-attachments/assets/13af666c-7104-45e9-b756-e2737d65a054

## Checklist

## Additional Notes
- Had to disable freeze screen in order to the recording to show the keys being pressed, it works the same when freezen screen is enabled though
- The `Ctrl+C` and `Ctrl+S` actions perform an exclusive override (e.g., `Ctrl+C` sets `copyToClipboard = true` and forces `saveToFile = false` and `pipeToCommand = false`). This was an intentional design choice because a user explicitly using a targeted shortcut expects that specific immediate action, overriding any background defaults.


oldirtty:
Screenshot setting could be have a 3 option button like the one in the image (Save to File, Copy to Clipboard and None) instead of some toggle options since in order to use screenshots you gotta have at least one enabled. I could implement that with intended
image

ffs-akl:
If a user already mapped Ctrl+C or Ctrl+S in their compositor config (e.g. Hyprland/Sway/Niri), how can you ensure those keystrokes reach your overlay instead of being intercepted by the compositor?

oldirtty:
You got a point, maybe be able to set the keybinds for each operation, like the Keybinds section in noctalia settings.

oldirtty:
> If a user already mapped Ctrl+C or Ctrl+S in their compositor config (e.g. Hyprland/Sway/Niri), how can you ensure those keystrokes reach your overlay instead of being intercepted by the compositor?

Ctrl+C/Ctrl+S are read from keyboard events on the overlay's layer-shell surface, which uses exclusive keyboard_interactivity. That guarantees no other client steals focus, but it doesn't guarantee the compositor won't consume the same combination first if it's also registered as a global bind.

Tested on Niri: with Ctrl+C bound globally, the overlay never receives the keystroke because the compositor intercepts it before dispatch. Ctrl+S (unbound) worked as expected. This is compositor-level bind precedence, not something the client can override and it wouldn't be appropriate for a client to try (that would mean overriding the user's own compositor config).

If you use Ctrl+C or Ctrl+S as global shortcuts in your compositor, they won't reach this overlay. Workaround: avoid mapping those combinations globally, or unbind them for now. Configurable shortcuts (reusing Noctalia's Keybinds settings) are a possible follow-up but out of scope for this PR.

ffs-akl:
> You got a point, maybe be able to set the keybinds for each operation, like the Keybinds section in noctalia settings.

Fair enough. My main concern is just ensuring keybinds can be customized or entirely disabled (left blank) down the road so they don't collide with compositor configs.

ffs-akl:
Default binds are great for out-of-the-box usability, but users must always have the freedom to override or completely disable them. If every app hardcoded standard bindings like Ctrl+C or Ctrl+S, workflow conflicts would be impossible to fix.

oldirtty:
Should be good now!

image

ffs-akl:
> Should be good now!

Perfect!

ItsLemmy:
imo the shell keybinds should stay as agnostic to the features as possible, or we will end up with a 100 keybinds in the settings which is clearly not what we want.  I would just go with two new kinds:
- Copy
- Save

That could be re-used later if necessary, we already have another PR which adds "Delete" (for clipboard entries and more) that would complement it well.

oldirtty:
Fair point, agree that keeping shell-level keybinds generic is the right call here. I'll rename ScreenshotConfirmClipboard/ ScreenshotConfirmSave to just Copy/Save so it's reusable.

ItsLemmy:
Thanks!

PR #3550, opened 2026-07-20
fix: correctly parse escaped characters in desktop entries

kirusfg:
Desktop entries apply general string escaping before `Exec` argument quoting, so both layers must be decoded before arguments are passed to applications.

https://specifications.freedesktop.org/desktop-entry-spec/latest/exec-variables.html

## Summary

Introduced decoding escaped characters in desktop entry Exec commands before launching applications. Added a test case to prevent regressions.

## Motivation

Couldn't run this desktop entry after upgrading to v5:

```
[Desktop Entry]
Name=scrcpy
GenericName=Android Remote Control
Comment=Display and control your Android device
# For some users, the PATH or ADB environment variables are set from the shell
# startup file, like .bashrc or .zshrc… Run an interactive shell to get
# environment correctly initialized.
Exec=/bin/sh -c "\\$SHELL -i -c scrcpy"
Icon=scrcpy
Terminal=false
Type=Application
Categories=Utility;RemoteAccess;
Keywords=adb;android;screen;mirroring;display;control;phone;device;
StartupNotify=false
```

Turns out it was because of the escaped `$` in the `Exec` line. The launcher code was not decoding the escape sequence before passing the command to `/bin/sh`. The argument received by `/bin/sh` was `\$SHELL` instead of `$SHELL`.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Build / packaging

## Testing

- `just test release`
- Built and installed the patched release, then successfully launched the desktop entry above with the Noctalia launcher

## Manual Coverage

- [ ] Tested on Niri
- [ ] Tested on Hyprland
- [x] Tested on Sway

## Checklist

### DIFF
--- a/src/system/desktop_entry_launch.cpp
+++ b/src/system/desktop_entry_launch.cpp
@@ -13,6 +13,29 @@ namespace {
 
   constexpr Logger kLog("desktop_entry_launch");
 
+  std::string decodeDesktopStringValue(std::string_view value) {
+    std::string result;
+    result.reserve(value.size());
+    for (std::size_t i = 0; i < value.size(); ++i) {
+      if (value[i] == '\\' && i + 1 < value.size()) {
+        const char next = value[i + 1];
+        if (next == '\\' || next == 's' || next == 'n' || next == 't' || next == 'r') {
+          // https://specifications.freedesktop.org/desktop-entry/latest/value-types.html
+          // https://specifications.freedesktop.org/desktop-entry-spec/latest/exec-variables.html
+          result += (next == 's') ? ' ' : (next == 'n') ? '\n' : (next == 't') ? '\t' : (next == 'r') ? '\r' : '\\';
+          ++i;
+          continue;
+        }
+        // Any other \X: remove the backslash, keep X
+        result += next;
+        ++i;
+        continue;
+      }
+      result += value[i];
+    }
+    return result;
+  }
+
   std::string stripFieldCodes(std::string_view exec) {
     std::string result;
     result.reserve(exec.size());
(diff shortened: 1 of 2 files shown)


Ly-sec:
Looks good to me @kirusfg , thanks for the clear writeup and the test.

The scrcpy case is a real bug: desktop entries do string escapes first, then Exec quoting, and we were leaving `\$` in the argv. Your two-layer decode matches the spec and the regression test covers it.

A couple of small non-blocking notes:

- Putting the general unescape inside `stripFieldCodes` is a bit mixed. A tiny dedicated helper would read a bit clearer, but not required for merge.
- Longer term, string unescaping probably belongs at `.desktop` parse time for all string keys. Fine to keep it launch-side for now while the parser stores raw values. Just don't double-unescape if that moves later.
- Pre-existing: we still strip field codes before undoing quotes. Spec wants the opposite. Out of scope here.

Happy to merge as-is, if you do want to solve those small nits, feel free to do that. I'll wait for your response :)

kirusfg:
I looked at the freedesktop spec and the code handling desktop entry parsing more closely. What you said is right, it'd be better if the parsing code handled those escape sequences earlier. Undoing quotes is also better done before removing field codes.    

I only extracted escape sequence parsing into a helper for now. Merge it like this so that people have a bit more desktop entries working in the interim.

Will do another pull request fixing the issues above later.

Ly-sec:
Awesome @kirusfg , I'll merge this as is and I'll keep an eye out for the other PR then :). Thank you very much!

PR #3578, opened 2026-07-22
feat(plugin-store): add sort button

### LINKED ISSUE #3547: [FEATURE] Allow plugins to be sorted
### Submission checklist

### Feature type

UI / visual improvement

### Feature summary

Allow plugins to be sorted something like `Newly Added`

### Motivation / use case

Because plugins are sorted alphabetically, a new plugin will not stand out. If a few new plugins are added since a user has last checked the available plugins, they need to check each row 1 by 1 to see if a new plugin they would like to use got added.

### Proposed solution

A sort option like 'Newly Added` would fit in case there is a plan to make this page sortable. 

### Alternatives considered

If there is no plan to make this page sortable, at least a  badge on new plugins could make them more visible.

_No response_

### Additional context

_No response_

Yilmaz4:
## Summary

I've added a button to sort the plugins in the plugin store by name, last update or date added, each in both ascending or descending order. To make space for the button, I've also made the source selector and the category selector into two dropdowns.

The catalog.toml file did not have any information for this so I had to update `update_catalog.py` in both official-plugins and community-plugins repos to generate `last_modified` and `date_added` fields based on the last commit to `plugin.toml` and the commit that created `plugin.toml` of each plugin respectively.

## Motivation

Without any date sorting, it's very difficult to see which plugins are new since the last time the user checked.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Build / packaging

## Related Issue

Closes #3547 

## Manual Coverage

- [x] Tested on Niri
- [ ] Tested on Hyprland
- [ ] Tested on Sway

## Checklist

## Additional Notes

I had to change a few strings in en.json and removed two that were no longer needed, but I'm not sure if I should've just created new strings instead so that it's clear to translators that translation is required.

I used a lot of the same code from the sort button from the wallpaper selector for consistency.

UI is not final. I'm open to ideas.

~~TODO: Make it so the same plugin remains selected after the sorting is changed. Currently the selection box stays where it was and, if the positions of the plugins in the grid changes, whatever plugin is at the index now becomes selected.~~

Could also make it so the dates are displayed in the plugin details page when clicking on a plugin, but that's probably out of scope of this PR.


ItsLemmy:
The PR title says "add sort button" but you went ahead and reworked the entire header.

IMO, Select (dropdowns) are inferior to the buttons/pill we have currently for this kind of quick filtering.

There is ample space to accomodate the sort button
image

Yilmaz4:
Fair, I just thought there was too much wasted space and the sort button wouldn't have made sense anywhere. In your example too, it would look awkward for there to be a singular square button on the top right and then everything else scattered around.

What about this? @ItsLemmy 




We could also make the source selector into pills again depending on how many custom sources we can expect the average user to add I suppose.

I can rename the PR if necessary but I do think the header needs adjustment.

ItsLemmy:
Yeah the original header was implemented quickly to get the ball rolling, but it definitely needs love, as it looks disconnected and floating. But Selects everywhere is not the way.

I'd also like to keep Official and Community sources shown at all time, so it's clear what you are seeing. People have a hard time understanding that noctalia's team do not provide direct support for community and custom source plugins. So I want to make that clear.   

Ofc we cannot know the amount of sources enabled, but I'd still go with buttons to select them.

Maybe a better idea would be to have the sorting (45 plugins [ ]) closer/just above to the results. And keep all filtering on top. It would feel more natural.

For the actual filters: 
First block/row: The sources as buttons (could wrap to multiple rows if needed but I doubt it will happen often)

Second block/row would be the tags, and they could use the full available width.

Yilmaz4:
Something like this?




(don't mind the arrow in the categories button facing the wrong way, I forgot to fix that while experimenting)

It might also be a good idea to make a new widget type for this kind of thing, where "All sources", "Official" and "Community" would be connected to each other in a long rectangle and the selected one would be highlighted in a box, I don't know how to explain it better but something similar to what material design has

ItsLemmy:
We already have that control, it's called "Segmented"

Yilmaz4:
Oh perfect

image

ItsLemmy:
merged the changes to the catalog.toml.   renamed the 2 fields for consistency. (update_at, added_at)

ItsLemmy:
Let me know when you consider it ready.

Yilmaz4:
I think it's ready now

ItsLemmy:
Thanks!

PR #3616, opened 2026-07-24
feat(shell): shared liquid-glass material engine + chrome rollout

theesfeld:
## Summary
Shared surface material engine for the whole shell, with global settings, per-surface overrides, experimental refraction gating, and chrome consumers (OSD / toasts / popup menus).

### Engine (`shell::material`)
- Modes: `solid` | `soft` | `liquid_glass`
- Density + luminance-aware body alpha
- Specular rim / outer-edge catch light for liquid glass
- Pure helpers: `fromShell`, `effectiveMode`, `resolve`, `applyToStyle`, refraction gating

### Global config
```toml
[shell.material]
mode = "liquid_glass"
density = 1.0
experimental_refraction = false
```

### Consumers
| Surface | Path |
|---------|------|
| Bar / dock / panels | `fromShell` + `applyToStyle` |
| OSD | material fill on card |
| Notification toasts | material fill on card |
| Popup chrome / context menus | `addCardBackground` + `setMaterialConfig` |

### Per-surface overrides
- `bar.*.material_mode`, monitor overrides
- `dock.material_mode`
- `shell.panel.material_mode`
- Empty/unset inherits `shell.material.mode`

### Experimental refraction
- Flag off (default): unchanged liquid-glass recipe
- Flag on + liquid glass + sampling available: stronger lens-style rim/softness (client attempt)
- Flag on + sampling unavailable: clear-glass paint fallback (no crash)

## Test plan
- [x] `config_schema_roundtrip` (opacity, overrides, refraction gating)
- [x] `config_path_resolution` (material paths)
- [x] `config_migration`
- [x] Full debug build of `noctalia`
- [ ] Appearance → Liquid Glass; confirm bar/dock/panels/OSD/toasts/menus
- [ ] Per-bar/dock material override vs global
- [ ] Toggle experimental refraction


ItsLemmy:
Thanks for the work here, but I'm going to close this.

We're in beta right now and the priority is stabilising and documenting what already exists, not expanding the surface. This doesn't add a capability we're missing: background_opacity and border (colour + width) are already per-surface and per-monitor, so a translucent bar with a soft edge over compositor blur is something a user can configure today. On top of that, the PR reinterprets background_opacity depending on a mode flag (raw alpha when off, luminance-mapped density when on), which is exactly the kind of context-dependent option that's hard to document cleanly. We also already have transparency_mode = solid|soft|glass on panels, so this adds a second spelling of the same idea.

For a cosmetic preset that's reachable with existing config, that's more complexity and documentation burden than it earns right now. Appreciate the effort, it's a scope call, not a quality one.

Issue #3614, opened 2026-07-24
[BUG] Taskbar active indicator shown on the wrong workspace


spasche:
### Submission checklist

- [x] I have searched existing issues and confirmed this is not a duplicate.
- [x] I am using the latest available version of Noctalia.

### Bug description

When the same application has multiple windows with identical titles on different workspaces, the Taskbar widget may associate the active indicator with the wrong window after Noctalia is restarted.

For example, with two GNOME Calculator windows—one on workspace 1 and another on workspace 2—focusing the window on workspace 1 may display the active dot under the Calculator icon on workspace 2.

### Steps to reproduce

1. Configure Taskbar to group by workspace
2. Open two windows of the same application with identical titles, such as two GNOME Calculator windows.
3. Move one window to workspace 1 and the other to workspace 2.
4. Restart Noctalia while both windows remain open.
5. Focus the application window on workspace 2.
6. Observe the active indicator in the Taskbar widget.

### Expected behavior

The active indicator should be displayed under the taskbar entry for the focused window on its current workspace. In this example, focusing the Calculator window on workspace 2 should show the active dot only under the Calculator icon on workspace 2.

### Actual behavior

The active indicator may be displayed under the matching application icon on another workspace. In this example, focusing the Calculator window on workspace 7 may show the active dot under the Calculator icon on workspace 1 instead.

### Logs / error output

```text

```

### Noctalia version

5.0.0_beta.4-1

### Compositor

Niri

### Distribution

Arch-based

### Environment information

```text
Installation method: cachyos repo
GPU: AMD integrated (amdgpu)
```

### Additional context

<img width="633" height="171" alt="Image" src="https://github.com/user-attachments/assets/9757a979-c6f0-44ec-b8cf-146285cd1336" />

Ly-sec:
Hey @spasche , I just fixed it with said commit (or at least, I can't reproduce the issue anymore). Please do let me know if that solved it for you.

spasche:
Thanks for the quick action. That doesn't fix it in my case. I asked codex if that can help:

• No — 55ce2a0 does not reliably fix the niri active-window-on-wrong-workspace issue.

  Why:

  - Niri task handles currently receive an app_id:title identifier from wlr-foreign-toplevel (src/wayland/wayland_toplevels.cpp:329).
  - Niri workspace assignments use numeric niri window IDs (src/compositors/niri/niri_workspace_backend.cpp:289).
[...]

PR #3620, opened 2026-07-24
fix(taskbar): preserve compositor ID workspace matches (#3614)

spasche:
Follow up to #3614

Workspace assignment first binds tasks using current or focused compositor window IDs. The previous-frame fallback could then revisit those tasks and overwrite the authoritative match, putting the active task back in the wrong workspace.

Only reuse a previous window ID for tasks that remain unassigned.


Ly-sec:
Merging it, I was very much overthinking it.

Issue #3584, opened 2026-07-23
[FEATURE] Add pinned apps to taskbar widget like on dock


gmbhneo:
### Submission checklist

- [x] I have searched existing issues and confirmed this has not been requested before.
- [x] I have checked existing pull requests for similar changes.

### Feature type

Configuration / customization

### Feature summary

Add the ability to show / manage apps also on panels taskbar. Like adding the ability to pin them, start them etc. 

### Motivation / use case

To me, having the need for two bars is waisted space. Most desktops only have one taskbar. With noctalia you are not able to use only a single taskbar atm. since you are in need to use the dock and the panel. 

I am used to have my apps in the taskbar widget. So I manage them there too. Which would mean I need the ability to pin/unpin them, reorder etc. Kinda like having the dock inside of the panel taskbar.

This would lead to a much cleaner desktop experience. 

### Proposed solution

Add the ability of the dock to the taskbar widget of the panel.

### References / related projects

https://danklinux.com/ 

Dank Material Shell for example is able to do exactly this. 

Also KDE Plasma or Windows etc. is using the same kind of method.

### Additional context

Just add most of the features and settings from dock to taskbar widget.

GhostEagle68:
## Disclosure

This investigation and the attached patch was put together with Claude Code's help. I'm not
an experienced programmer. I have read through the relevant source and tested the patch myself (built it locally, swapped it into my live session, confirmed pin/unpin round-trips correctly), but the code should
be reviewed since it was written by AI mainly. Sharing it as a starting point/confirmation of root cause, not finished in any means.

---

I noticed this too. Half of it looks like it's already fixed by #3130. The taskbar widget's own right-click Pin/Unpin works fine for me on current `main` (confirmed on `7b649fc`, 2026-07-23). But the other half is
still open: the **Application Launcher's** right-click menu only offers "Pin to dock"/"Unpin from dock." If you're not using the dock (just a `taskbar` bar widget, like this issue describes), apps pinned from the launcher go to `dock.pinned` instead of the taskbar's own `pinned` list. From the UI it looks like pinning from the launcher just does nothing.

<img width="653" height="585" alt="Image" src="https://github.com/user-attachments/assets/a7ee61d1-8c7e-453c-93fb-0cef1bc07684" />

[...]

Issue #3130, opened 2026-06-24
[BUG] V5 - Pinned Apps - Taskbar not appearing - Only on Dock


gmbhneo:
### Submission checklist

- [x] I have searched existing issues and confirmed this is not a duplicate.
- [x] I am using the latest available version of Noctalia.

### Bug description

If I add an App to be pinned to the taskbar, they won't appear like in 4.7 for example. They only get pinned on the dock, which I am not using.

In 4.7 pinned apps still appear at the taskbar widget of the panel bar.

Also, you are not able to manage them like in the dock. And you are not able to pin an app from within the taskbar widget at all.

### Steps to reproduce

Pinn an App and see them not appear in the taskbar widget at the panel

### Expected behavior

Show the pinned app

### Actual behavior

They simply won't appear and you are not able to configure them like in the dock.

### Logs / error output

```text

```

### Noctalia version

5.0.0.r1191.g39a4a335c-12

### Compositor

Niri

### Distribution

Arch-based

### Environment information

```text
AUR, Noctalia-Git
```

Ly-sec:
Hey @gmbhneo , I just implemented pinning apps to the taskbar when not grouped by workspace.

PR #3635, opened 2026-07-25
feat(launcher): pin/unpin apps to taskbar widgets, not just dock

GhostEagle68:
Disclosure

This investigation and the attached patch was put together with Claude Code's help. I'm not
an experienced programmer. I have read through the relevant source and tested the patch myself (built it locally, swapped it into my live session, confirmed pin/unpin round-trips correctly), but the code should
be reviewed since it was written by AI's help. Sharing it as a starting point/confirmation of fix, not finished in any means.

## Summary

The Application Launcher's right-click "Pin"/"Unpin" only ever wrote to `dock.pinned`, even when no
dock is configured. Just a `taskbar` bar widget. Apps pinned from the launcher went to a list
nothing displays. This adds independent **"Pin to taskbar" / "Unpin from taskbar"** entries to the
launcher's context menu (alongside the existing dock ones, not replacing them) that write through
the same `{"widget", <name>, "pinned"}` override path `TaskbarWidget::setEntryPinned` already uses
for its own right-click pin, so pin state stays in sync between the launcher and the taskbar strip.

## Motivation

Half of the taskbar-pinning story (issue #3584) was already fixed by #3130; right-clicking an icon
directly in the taskbar strip works fine. But the Application Launcher's own pin action was never
updated to look at taskbar widgets at all; it's hardcoded to `ConfigService::config().dock.pinned`.
From a user's perspective (no dock enabled, just a taskbar), pinning from the launcher silently does
nothing.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Build / packaging

## Related Issue

Related to #3584 (this covers the pin/unpin half of that issue; drag-to-reorder is not addressed
here).

## Testing

- `python3 tools/i18n-check.py`: passes (830 tr keys, 7 trp base keys, all present in catalog).
- `just format`: clean, no additional diff after running.
- `meson setup build-release --buildtype=release -Db_lto=true -Dcpp_std=c++23 --prefix=/usr/local &&
  meson compile -C build-release`: builds with zero warnings.
- `meson test -C build-release` (with `-Dtests=enabled`): all 55 existing unit tests pass, including
  `dock_pinned_apps` (the helper this patch reuses).
- No automated test covers `LauncherPanel::openAppActionsMenu` directly, so this was also verified
  manually end-to-end (see Manual Coverage).

## Manual Coverage

- [x] Tested on Hyprland
- [ ] Tested on Niri
- [ ] Tested on Sway

Manual round-trip tested: pin from the Application Launcher → app appears in the taskbar strip
immediately → `widget.taskbar.pinned` updates in `settings.toml` → right-clicking the now-pinned
icon in the taskbar correctly shows "Unpin" → unpinning from either surface removes it cleanly and
stays in sync.

Only tested single-monitor / single bar with one `taskbar` widget. Multi-bar setups with more than
one taskbar widget should work per the code (it loops over every configured `taskbar`-type widget),
but I haven't verified that myself.

## Screenshots / Videos

---

image

---

Image

---

Image

---

Image

## Checklist

- [x] I will update end-user documentation after merge, or this PR does not change user-facing
      configuration or behavior.
- [x] I did not edit non-English translation files unless this PR is explicitly for translation
      tooling, an import/export sync, or a maintainer-requested locale change.

## Additional Notes

This doesn't touch drag-to-reorder, which #3584 also asks for. The dock already has that
(`dock_instance.h` `DragState`, `dock.cpp` `beginDrag`/`updateDrag`/`endDrag`) but the taskbar strip
doesn't yet. Happy to look into porting that too if it'd help, but wanted to keep this PR scoped to
the part I could verify working end-to-end.

No end-user documentation exists for the launcher's pin context menu today (checked `README.md`,
`CONTRIBUTING.md`, and elsewhere in this repo), and this doesn't add a new config key. Just a new
UI entry point onto the existing `[widget.<name>].pinned` field the taskbar already reads/writes. So there was nothing in-repo to update.

### DIFF
--- a/assets/translations/en.json
+++ b/assets/translations/en.json
@@ -627,7 +627,9 @@
     "context-menu": {
       "open": "Open",
       "pin-to-dock": "Pin to dock",
-      "unpin-from-dock": "Unpin from dock"
+      "unpin-from-dock": "Unpin from dock",
+      "pin-to-taskbar": "Pin to taskbar",
+      "unpin-from-taskbar": "Unpin from taskbar"
     },
     "empty": {
       "no-results": "No results found",
--- a/src/shell/launcher/launcher_panel.cpp
+++ b/src/shell/launcher/launcher_panel.cpp
@@ -1611,10 +1611,36 @@ void LauncherPanel::openAppActionsMenu(std::size_t index, float anchorX, float a
       m_config != nullptr && shell::dock::pinned_apps::containsEntry(m_config->config().dock.pinned, *match);
   const bool canPinToDock = m_config != nullptr && !dockPinned;
   const bool canUnpinFromDock = m_config != nullptr && dockPinned;
-
+  // Taskbar widgets keep their own `pinned` list under `[widget.<name>]` (see
+  // TaskbarWidget::pinnedConfigIds), separate from `[dock].pinned`. Mirror pin state across every
+  // configured taskbar widget so "Pin" here matches what right-clicking an icon in the taskbar does.
+  std::vector<std::string> taskbarWidgetNames;
+  if (m_config != nullptr) {
+    for (const auto& [name, widgetConfig] : m_config->config().widgets) {
+      if (widgetConfig.type == "taskbar") {
+        taskbarWidgetNames.push_back(name);
+      }
+    }
+  }
+  const bool hasTaskbarWidget = !taskbarWidgetNames.empty();
+  bool taskbarPinned = false;
+  if (hasTaskbarWidget) {
+    const auto& widgets = m_config->config().widgets;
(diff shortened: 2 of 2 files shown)


GhostEagle68:
@Ly-sec Here's the PR for #3584 :)

ItsLemmy:
I'm closing this one. We are refactoring all 3 app pinning mechanism
- Launcher
- Dock
- Taskbar

To keep things simple each component will be responsible for its own app pinning, rather than having the launcher do multiple things.

Thanks none the less.

Issue #3294, opened 2026-07-06
[FEATURE] Support subscribing to public/read-only ICS (webcal) URLs as a calendar provider


dfseifert:
### Submission checklist

- [x] I have searched existing issues and confirmed this has not been requested before.
- [x] I have checked existing pull requests for similar changes.

### Feature type

New functionality

### Feature summary

Noctalia v5's Calendar service currently only supports iCloud, CalDAV, and Google as providers, all of which need an authenticated account, with no built-in way to subscribe to a plain public/read-only `.ics` or `webcal://` URL, and I found no existing PR or issue covering this.

### Motivation / use case

The Calendar service currently supports iCloud, CalDAV, and Google as providers.
However, many calendars I would like to see in my desktop environment are published as plain, read-only `.ics` files (or `webcal://` links) with no login required, for example public holiday calendars, my university/class schedules, sports team fixtures or shared team calendars exported from tools that only offer a "public calendar link."

### Proposed solution

Add a new calendar provider type (e.g. "ICS URL" / "Web Calendar") that:

- Accepts a plain `https://` or `webcal://` URL pointing to an .ics file, no credentials required. (but maybe configurable?)
- Periodically re-fetches the file on a configurable interval
- Parses standard iCalendar (RFC 5545) events, including recurring events (RRULE) and all-day events, the same way the existing providers do.
- Shows up alongside iCloud/CalDAV/Google calendars in the calendar panel, with its own toggle/color like other calendar sources.
- Handles fetch failures gracefully (e.g. stale-data fallback + a visible error/warning rather than silently dropping events).

### Alternatives considered

I have considered setting up a local CalDAV server (e.g. [Radicale](https://github.com/Kozea/Radicale)) just to mirror a public ICS feed but it is a heavy and fragile workaround for something that should be a simple URL fetch.

### References / related projects

- #2928
- #744

### Additional context

This would primarily benefit read-only/subscribe-only use cases but it doesn't need write support (no event creation/editing back to the source), since public ICS feeds are one-way by nature.

PR #3636, opened 2026-07-25
refactor(calendar): replace custom iCalendar parser with libical

floydya:
## Summary

- Replace the custom iCalendar parser and recurrence implementation with libical.
- Preserve the existing `parseICalEvents` interface and `CalendarEvent` output.
- Use libical for RFC 5545 parsing, text escaping, timezones, recurrence rules, exclusions, and overridden occurrences.
- Add parser regression coverage for wrapped calendars, malformed events, floating and all-day events, DST boundaries, `EXDATE`, `UNTIL`, and `RECURRENCE-ID`. This caused rename of file with tests (cause it's now testing not only recurrence).
- Add libical to Meson, Nix, Guix, CI, and documented distribution dependencies.

## Motivation

- Prepare the calendar parser for future webcal support.
- Avoid maintaining a partial RFC 5545 implementation.
- Keep parsed `LOCATION` values available for a future action that opens meeting links or locations in the browser.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Refactoring
- [x] Build / packaging

## Related Issue

Preparation for #3294 

## Testing

- `just build` - passed.
- `just lint` - passed.
- `just test debug ical_parser --print-errorlogs` - passed.
- `just test debug --print-errorlogs` - 50 of 51 test targets passed. The unrelated `process_test` failed locally with `completion-only async command stdout was wrong`; this branch does not modify the process implementation or its tests.
- Tested through a local Nix flake build on Niri - calendar behavior works as expected.

No branch-specific compiler or lint warnings were found.

## Manual Coverage

- [x] Tested on Niri
- [ ] Tested on Hyprland
- [ ] Tested on Sway

## Screenshots / Videos

Not applicable. No UI changes.

## Checklist

## Additional Notes

- Webcal transport and subscription management are out of scope in this PR.
- `LOCATION` is still copied into `CalendarEvent`, including meeting URLs stored in that property.
- A standalone RFC 5545 `URL` property is not yet exposed through `CalendarEvent` and requires a follow-up change.


ItsLemmy:
### [P1] Bound COUNT-limited recurrence expansion                                                                                                         
                                                                                                                                                           
 src/calendar/ical_parser.cpp:297-300                                                                                                                      
                                                                                                                                                           
 icalcomponent_foreach_recurrence() can perform unbounded synchronous work for old recurrences with COUNT. Libical 3.0.20 only skips directly to the requested window for unbounded rules; counted rules iterate from DTSTART to preserve count semantics.                                
                                                                                                                                                           
 A probe using:                                                                                                                                            
                                                                                                                                                           
 ```ical                                                                                                                                                   
   DTSTART:20200101T000000Z                                                                                                                                
   RRULE:FREQ=SECONDLY;COUNT=10000000                                                                                                                      
 ```                                                                                                                                                       
                                                                                                                                                           
 with a 2024–2026 window returned zero events but took 4.58 seconds. Parsing happens inline in the CalDAV completion path at src/calendar/caldav_client.cpp:109. A valid remote response can therefore stall calendar processing for seconds or substantially longer. Add an explicit work budget/cancellation mechanism or move bounded expansion off the synchronous completion path.                                                         
                                                                                                                                                           
 ### [P2] Do not emit cancelled recurrence overrides                                                                                                       
                                                                                                                                                           
 src/calendar/ical_parser.cpp:280-282                                                                                                                      
                                                                                                                                                           
 A RECURRENCE-ID override with STATUS:CANCELLED and a valid DTSTART excludes the master occurrence, then gets unconditionally emitted as a normal event  because status is never inspected.                                                                                                                        
                                                                                                                                                           
 Verified with a three-day daily series: cancelling the second occurrence still returned three visible events, including the cancelled one. Keep the  recurrence ID as an exclusion, but do not append a cancelled component.                                                                                   
                                                                                                                                                           
 ### [P2] Malformed overrides must not suppress valid occurrences                                                                                          
                                                                                                                                                           
 src/calendar/ical_parser.cpp:261-266                                                                                                                      
                                                                                                                                                           
 Override IDs are collected before validating that their component has a valid DTSTART. The malformed component is skipped later at lines 271–276, but its ID still suppresses the corresponding master occurrence.                                                                                                  
                                                                                                                                                           
 Verified with a three-occurrence master plus a UID/RECURRENCE-ID-only component: only two occurrences were returned. The previous implementation only collected overrides from events with a parsed start, so this is a regression. Validate overrides before applying them, with an explicit exception for     
 properly cancelled instances that omit DTSTART.                                                                                                           
                                                                                                                                                           
 ### [P2] Expand RDATE-only recurrence sets                                                                                                                
                                                                                                                                                           
 src/calendar/ical_parser.cpp:280-282                                                                                                                      
                                                                                                                                                           
 Only components containing RRULE are passed to icalcomponent_foreach_recurrence(). A valid event containing DTSTART plus RDATE, but no RRULE, therefore emits only DTSTART; the additional recurrence dates are discarded.                                                                                        
                                                                                                                                                           
 RFC 5545 defines the recurrence set as the union of DTSTART, RRULE, and RDATE. A probe with one RDATE returned one event instead of two. Gate expansion on either RRULE or RDATE.                                                                                                                                 
                                                                                                                                                           
 ### [P2] Deduplicate overlapping RRULE and RDATE instances                                                                                                
                                                                                                                                                           
 src/calendar/ical_parser.cpp:243                                                                                                                          
                                                                                                                                                           
 Libical explicitly documents that icalcomponent_foreach_recurrence() does not remove duplicate instances produced by overlapping RRULE/RDATE values. The callback currently appends every span without deduplication.                                                                                              
                                                                                                                                                           
 Verified with RRULE:FREQ=DAILY;COUNT=2 and an RDATE equal to the second occurrence: the parser returned that occurrence twice. RFC 5545 requires duplicate recurrence instances to collapse into one.

floydya:
Thanks, all five issues are fixed with regression tests:

- Counted recurrence parsing now runs off the synchronous CalDAV completion path
- Cancelled overrides suppress the original occurrence without being emitted
- Malformed overrides no longer suppress valid occurrences; cancelled overrides without DTSTART remain supported
- RDATE-only recurrence sets are expanded
- Overlapping RRULE/RDATE instances are deduplicated

Please let me know if anything else needs to be addressed

ItsLemmy:
Can you test with my latest commit and see if all works well for you?

floydya:
> Can you test with my latest commit and see if all works well for you?

Tested this branch (including your latest commit) manually on Nix, and everything works as expected

Also:
- Found and fixed the CI failure caused by libical 3/4 API differences
- Added coverage for RRULE/RDATE expansion, including RDATE periods expressed as durations. 

The Arch CI environment now builds successfully(tested via docker), lint passes, and all 65 tests pass.
Calendar parser and CalDAV tests also pass on Nix.

PR #3637, opened 2026-07-25
feat: added support for ICS calendars

### LINKED ISSUE #3294: [FEATURE] Support subscribing to public/read-only ICS (webcal) URLs as a calendar provider

agokule:
## Summary

I added a new type of calendar (ICS) to the settings page which can take a URL to an *.ics file, and use that for the calendar.

In terms of code changes, it's mostly the same stuff that was in #2928 before it got closed. However it does fix the 2 main pain points that were mentioned because it turns out that the `parseICalEvents` function was edited to support both of those pain points, so there weren't a lot of changes needed to make it work.

## Motivation

Many apps expose their calendars in this format, so it is important to support it.

## Type of Change

- [ ] Bug fix
- [X] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Build / packaging

## Related Issue

This closes #3294
Also this goes hand in hand with PR #3636, we coincidentally happened to be working on this at the same time!

## Testing

Ran `just test`, all tests passed.

I put this content in a file `testing.ics`:

```
BEGIN:VCALENDAR
VERSION:2.0
PRODID:-//Sample Event//EN
CALSCALE:GREGORIAN
METHOD:PUBLISH
BEGIN:VEVENT
SUMMARY:Sample Team Meeting (ICS)
DTSTART:20260725T150000Z
DTEND:20260725T160000Z
LOCATION:Online Room A
DESCRIPTION:Review daily goals and projects.
STATUS:CONFIRMED
END:VEVENT
END:VCALENDAR
```

And then ran `python -m http.server` in that directory, this creates a static http server at port 8000 to that directory. Then I added `http://localhost:8000/testing.ics` using the Noctalia settings UI as an ICS calendar, and the event "Sample Team Meeting (ICS)" showed up as happening today (see screenshot below)

## Manual Coverage

- [ ] Tested on Niri
- [X] Tested on Hyprland
- [ ] Tested on Sway
- [X] Tested with multiple monitors

## Screenshots / Videos

image

## Checklist

## Additional Notes

N/A


ItsLemmy:
### [P1] Normalize webcal:// URLs before passing them to libcurl                                                                                          
                                                                                                                                                           
 fetchIcs() forwards the configured URL unchanged  /src/calendar/calendar_service.cpp#L543-L558.
 The related issue  explicitly requires webcal:// support, but the HTTP client passes the value directly to CURLOPT_URL; libcurl does not support a webcal protocol.          
                                                                                                                                                           
 Result: common calendar subscription links fail with CURLE_UNSUPPORTED_PROTOCOL, so a core requested input cannot load.                                   
                                                                                                                                                           
 Normalize webcal://… to https://… before constructing the request. Alternatively, reject it in settings with an actionable error, but that would not satisfy issue #3294.                                                                                                                                      
                                                                                                                                                           
 ### [P2] Reject successful HTTP responses that are not valid calendars                                                                                    
                                                                                                                                                           
 After any HTTP 200, fetchIcs() treats the parser output as authoritative /src/calendar/calendar_service.cpp#L560-L577. parseICalEvents() does not validate the VCALENDAR envelope; HTML login pages, expired-share pages, proxy errors, and malformed ICS therefore parse as an empty event list. accountDone(..., true, {}) then replaces previously cached events and persists the empty result.                                                          
                                                                                                                                                           
 Validate that the response is an actual calendar before calling accountDone(..., true, ...). Invalid content should use the failure path so stale events remain available. A structured parser result distinguishing “valid empty calendar” from “invalid input” would be safest.

agokule:
Thanks for the suggestions @ItsLemmy. I wanted to ask a question about the implementation of this, since you didn't mention in your original comment.

Since this PR was opened, #3636 merged and changed parseICalEvents() to take an `ICalParseControl&` and return an `ICalParseResult`. After merging main, `fetchIcs()` no longer compiles, so I have to decide what to pass for the control — and that's more of a design question than a mechanical fix.

I'm guessing then that you would want this ICS parsing to happen on a separate thread, just like the CalDav calendars are, probably why you approved `ICalParseControl` to be added in that PR.

Since Noctalia already handles CalDav in a similar way, here's what I'd like to do, pending your OK since it touches code outside this feature: extract the job queue + `jthread` + `DeferredCall` delivery out of `CalDavClient::State` into a shared `calendar::ICalParseWorker`. Basically making it generic so that both the CalDav and ICS calendar systems can use it.

I will probably start writing that code soon, just wanted to check in since you might not want that. The options as I see them:

1. Extract the shared worker as above. Bigger diff and it touches caldav_client.cpp, but there's only one implementation of the threading.
2. Parse on the main thread with a default-constructed `ICalParseControl`. Just a two line change. The `remainingRecurrenceWork` budget still bounds it, but the main loop stalls for the duration.
3. A separate small `IcsClient` with its own worker, mirroring `CalDavClient`. Leaves `caldav_client.cpp` untouched at the cost of duplicating the threading code.

(1) is my preference, but (2) is clearly the cheapest if the mainloop stall is acceptable for now and you'd rather handle the threading separately. Happy to do whichever you'd merge. Obviously if you have another idea you'd prefer that's fine with me too.

agokule:
> I will probably start writing that code soon...
> (1) is my preference...

Ok so I changed my mind on that and decided to just fix the 2 issues you described above, and just pass a blank `calendar::ICalParseControl` for now so it does run on the main thread.

I also added tests in `ical_parser_test.cpp` as well for those situations.

If you do want the multi-threading thing to be done then I would be happy to try, just know that I am new to multi-threading in C++, so it might be better for someone else to take over that portion.

PR #3642, opened 2026-07-26
widgets/tray: migrate to typed widget definition

Ape:
Follow up to #3630.

BTW, could we do the merges without squashing? The squashing makes it slightly harder to work with the commits for me.


ItsLemmy:
Yeah for yours make sense to not squash, it's just an habit I got since most of the times for other PR it's better squashed since it's many commit for a singular feature.

PR #3653, opened 2026-07-26
feat(notification): auto-clear notification history

### LINKED ISSUE #3514: [FEATURE] auto clear notifications older than x
### Submission checklist

### Feature type

New functionality

### Feature summary

A settings option which would clear all notifcations older than x (x days, minutes, hours).

### Motivation / use case

when a user doesnt want to completely disable history (#3513), but still wants to clear them after a given time

### Proposed solution

A settings option, which would automatically clear all notifications older than the threshold

### Alternatives considered

_No response_

_No response_

### Additional context

_No response_

AdityaKr015:
## Summary

- Auto-clear notification history entries older than a configurable number of hours.
- Adds a "Keep history for" stepper in Settings -> Notifications -> History.
- Value 0 disables auto-clear (keep forever).

## Motivation

Notification getting piled up doesn't look good, this allows users to auto clears them with their own choice of time.This allows automatic cleanup of stale entries without manually clearing history.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Build / packaging

## Related Issue

Closes #3514 

## Testing

- Ran `meson compile -C build-debug` && `meson test -C build-debug`  # 56/56 pass
- Then `./build-bebug/noctalia` to start debug build
- Manual:- Enabled (1h), sent test notifications via `notify-send`, notification stays for the time until the feature kicks in and clears it.

## Manual Coverage

- [x] Tested on Niri
- [ ] Tested on Hyprland
- [ ] Tested on Sway

## Screenshots / Videos

image

## Checklist

- [x] This PR is ready for review
- [x] I ran `just format` with clang-format v22+ installed
- [x] I ran the relevant build or test commands
- [x] I will update end-user documentation after merge
- [x] I added or updated `assets/translations/en.json`

### DIFF
--- a/src/notification/notification_manager.cpp
+++ b/src/notification/notification_manager.cpp
@@ -14,7 +14,6 @@
 #include <string_view>
 
 namespace {
-
   constexpr Logger kLog("notification");
   constexpr auto kImplicitDuplicateWindow = std::chrono::seconds(1);
   constexpr std::string_view urgencyStr(Urgency u) noexcept {
@@ -655,9 +654,9 @@ void NotificationManager::pauseExpiry(uint32_t id) {
 void NotificationManager::setHistoryRetentionHours(int hours) {
   m_historyRetentionHours = hours;
   if (hours > 0) {
-    m_autoClearTimer.startRepeating(std::chrono::seconds(60), [this]() { cleanupOldHistoryEntries(); });
+    m_historyRetentionTimer.startRepeating(std::chrono::seconds(60), [this]() { cleanupOldHistoryEntries(); });
   } else {
-    m_autoClearTimer.stop();
+    m_historyRetentionTimer.stop();
   }
   cleanupOldHistoryEntries();
 }
@@ -760,7 +759,14 @@ void NotificationManager::markNotificationHistorySeen() {
 }
 
 void NotificationManager::schedulePersistHistory() {
-  DeferredCall::callLater([this]() { persistHistoryToDisk(); });
+  if (m_persistScheduled) {
+    return;
+  }
+  m_persistScheduled = true;
+  DeferredCall::callLater([this]() {
+    m_persistScheduled = false;
+    persistHistoryToDisk();
+  });
 }
 
 void NotificationManager::persistHistoryToDisk() {
--- a/src/notification/notification_manager.cpp
+++ b/src/notification/notification_manager.cpp
@@ -17,7 +17,6 @@ namespace {
 
   constexpr Logger kLog("notification");
   constexpr auto kImplicitDuplicateWindow = std::chrono::seconds(1);
-
   constexpr std::string_view urgencyStr(Urgency u) noexcept {
     switch (u) {
     case Urgency::Low:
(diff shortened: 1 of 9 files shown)


ItsLemmy:
Thanks for this, the feature itself is wanted and the wiring reuses the right helpers (emitPendingDBusClose, notifyUnreadStateChangedIfNeeded, schedulePersistHistory). The processExpired() dedupe in notification_service.cpp is nice to have.   Main ask is a simplification pass, plus a few real bugs.

Drop the time unit. It's the source of most of the problems here:

- One Range{1, 525600} (minutes-in-a-year) is applied to all three units. With unit = "days" that's hours(525600 * 24) = 12.6M hours, and WallClock::duration is int64 nanoseconds (max ~292 years) — so now() - threshold overflows. No single range can be right for three units.
- The schema max (525600) and the GUI stepper max (365) disagree, so the UI can't represent a hand-edited config.
- It forces the finalize hook that silently rewrites an unknown unit to "days". That's the silent default chain CLAUDE.md forbids; enumField already exists and warns via Diagnostics.
- Plus a second GUI row and the whole settings.options.time-unit block.

Naming. "Threshold" is vague. I'd collapse the whole thing to one key, with 0 = keep forever (same pattern as the battery warning in config_types.h:1215):

[notification]
history_retention_hours = 0

One GUI row, "Keep history for", valueSuffix = "h". Stepper's value field is a text input, so a 1–8760 range is fine to type. No toggle, no unit.

Bugs independent of the above:

1. Live notifications get pruned. History holds active = true entries for on-screen toasts, and timeout = 0 (critical) ones never expire. The toast stays visible while its history row disappears; on dismissal close() re-adds it and the reschedule deletes it again. Prune closed entries only.
2. The cleanupOldHistoryEntries() call in loadPersistedHistory() is dead — history loads in the Application constructor (application.cpp:143), well before applyAutoClearConfig() runs (application_ui.cpp:703), so m_autoClearEnabled is still false.
3. Timer is steady-clock but the cutoff is wall-clock, so the prune fires late across suspend. Given the recursion between scheduleAutoClearTimer() and cleanupOldHistoryEntries() (and the double reschedule in the timer callback), a single repeating 60s timer while enabled would be simpler and suspend-proof.
4. Entries without receivedWallClock are never pruned. Handle that explicitly.

Minor:

- The "avoids iterator invalidation" comment isn't accurate (removeHistoryEntry erases mid-deque already), and it narrates rationale, please drop it.
- tests/config_schema_roundtrip_test.cpp:337 sets a non-default value for every NotificationConfig field; add the new one there.
- std::max(1, value) re-clamps what the schema already clamped, can be dropped.
- New user-facing keys need a docs follow-up in the docs repo (separate commit, not a blocker here).

AdityaKr015:
Hello, I have made the changes you asked for, please take a look

- **Dropped the time unit** , only`history_retention_hours = 0` key there. Same pattern as battery warning threshold.
- **Naming** renamed from "threshold" to `history_retention_hours`. GUI has "Keep history for" with hours option to configure.

and the other bugs & minor changes.

I will see the doc repo for the PR.

AdityaKr015:
Clipboard files failing lint...

ItsLemmy:
yeah that's on me, I fixed it earlier but forgot to push

src/notification/notification_manager.cpp
```diff
@@ -719,14 +760,7 @@ void NotificationManager::markNotificationHistorySeen() {
...
 void NotificationManager::schedulePersistHistory() {
-  if (m_persistScheduled) {
-    return;
-  }
-  m_persistScheduled = true;
-  DeferredCall::callLater([this]() {
-    m_persistScheduled = false;
-    persistHistoryToDisk();
-  });
+  DeferredCall::callLater([this]() { persistHistoryToDisk(); });
```
ItsLemmy: This should be reverted to it's original state, as it debounce the writes to disk.

src/notification/notification_manager.cpp
```diff
@@ -17,7 +17,6 @@ namespace {
 
   constexpr Logger kLog("notification");
   constexpr auto kImplicitDuplicateWindow = std::chrono::seconds(1);
-
```
ItsLemmy, line 20: unnecessary diff

src/notification/notification_manager.h
```diff
@@ -161,7 +164,8 @@ class NotificationManager {
   ) const;
   uint32_t suppressExternal(std::string_view appName, Urgency urgency);
 
-  bool m_persistScheduled = false;
+  int m_historyRetentionHours = 0;
+  Timer m_autoClearTimer;
```
ItsLemmy: m_historyRetentionTimer would match the name better

AdityaKr015:
Done

ItsLemmy:
Thanks!

AdityaKr015:
Thank you for the merge

PR #3654, opened 2026-07-26
feat(switcher): add per-workspace window switcher toggle

Aceix:
## Summary

Adds a settings-panel toggle (Keybinds > window_switcher.per_workspace) that restricts the Alt-Tab window switcher to windows on the currently active workspace, instead of always listing windows across every workspace.

## Motivation

I needed to switch between windows within the same workspace

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Build / packaging

## Manual Coverage

- [ ] Tested on Niri
- [x] Tested on Hyprland
- [ ] Tested on Sway

## Checklist

## Additional Notes

AI powered change

### DIFF
--- a/assets/translations/en.json
+++ b/assets/translations/en.json
@@ -2071,6 +2071,10 @@
         "validate": {
           "description": "Confirm selections, submit inputs, or activate the focused item",
           "label": "Confirm"
+        },
+        "window-switcher-per-workspace": {
+          "description": "Restrict Alt-Tab window switching to windows on the current workspace only",
+          "label": "Switch Windows: Current Workspace Only"
         }
       },
       "lockscreen": {
--- a/src/config/config_overrides.cpp
+++ b/src/config/config_overrides.cpp
@@ -348,6 +348,7 @@ namespace {
         && widgetMapEqual(a.widgets, b.widgets)
         && desktopWidgetsConfigEqual(a.desktopWidgets, b.desktopWidgets)
         && a.hotCorners == b.hotCorners
+        && a.windowSwitcher == b.windowSwitcher
         && lockscreenWidgetsConfigEqual(a.lockscreenWidgets, b.lockscreenWidgets)
         && a.wallpaper == b.wallpaper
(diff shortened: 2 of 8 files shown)


ItsLemmy:
## [P1] Match the backend workspace key, not only Workspace::id                                    
window_switcher.cpp:328-332 returns workspace.id, while workspaceKey is compared         
directly at lines 406-409  Those identifiers are not equivalent across backends:      
- Hyprland named workspaces expose the numeric/negative ID as Workspace::id, but assignments use 
the workspace name.                                                                            
- Sway numbered named workspaces can expose "1: web" as Workspace::id, while assignments use     
"1".                                                                                           
In either case, every window on the active workspace fails the comparison and the switcher can  
become empty. Workspace membership needs the backend’s assignment-key semantics—ID, name, or    
index as appropriate—not raw Workspace::id equality.                                            

## [P2] Filter assignments by the switcher’s output                                                
platform.workspaceWindowAssignments() at line 344 requests global assignments even though the active workspace 
was selected from m_output. Mango, for example, reuses tag IDs on each output and only filters  
by monitor when an output is supplied. Consequently, opening the switcher on DP-1/tag 1 also    
retains windows from HDMI-A-1/tag 1.                                                            
Membership must be output-aware. Take care that output-filtered assignment rows are not         
followed by re-adding global live toplevels as “unassigned.”

## [P2] Preserve multiple workspace memberships until after filtering                              
assignmentById[key] = assignment at line 352 collapses all assignment rows for a window to one            
arbitrary/last workspace before the new filter runs. The assignment API is one-to-many:         
- KDE emits one row per desktop for an “all desktops” window.                                    
- Mango emits one row per tag for a multi-tagged client.                                         
A window visible on the active workspace is therefore removed whenever a different membership   
wins the map overwrite. Test membership against all assignment rows first, then deduplicate the 
resulting windows.

ItsLemmy:
stale

Issue #3283, opened 2026-07-05
[BUG] Battery in Graphical mode with no label makes status icons not visible


msmafra:
### Submission checklist

- [x] I have searched existing issues and confirmed this is not a duplicate.
- [x] I am using the latest available version of Noctalia.

### Bug description

The bolt here, on Graphical mode with label disabled, can only be seen if the battery charge is low enough while the laptop is plugged in.
<img width="418" height="46" alt="Image" src="https://github.com/user-attachments/assets/f8931e81-b541-4876-b154-6c8afbf0c377" />

### Steps to reproduce

1. Switch the battery status showing mode to Graphical;
2. Disable the label;
3. if the battery is full enough you won't see the status icon inside it, from about 70% down, it will start to be visible.

### Expected behavior

To be able to see the battery/power state at all times when the battery icon is set to be visible all the time with disabled label and in Graphic mode instead of Glyph mode.

### Actual behavior

It has the status icon, but is in the same color of the background making it invisible when the laptop is most of the time plugged in.

### Logs / error output

```text

```

### Noctalia version

v5.0.0-beta1

### Compositor

Niri

### Distribution

Arch-based

### Environment information

```text
Manually compiled from the tar.gz available for the Beta 1 in CachyOS

CPU: 6-core Intel Core i7-10750H
Kernel: 7.1.2-3-cachyos
iGPU Intel CometLake-H GT2 [UHD Graphics] i915 driver
dDPU NVIDIA TU106M [GeForce RTX 2060 Mobile] nvidia driver 610.43.02
```

### Additional context

Maybe it can be solve by setting its color to one of the other ones in the palette since the only configurable one inside this widget panel is **Warning Color**.

msmafra:
The same in Beta 2

msmafra:
Same on Beta 3 and noctalia v5.0.0 (v5.0.0-beta.3-204-g135126eb6c33)

PR #3659, opened 2026-07-27
Battery widget: show state glyph always outside the graphical battery

notiant:
As described in #3283, there is a visibility issue with the state glyph icon inside the graphical battery widget. To fix this, I moved the glyph outside of the graphical battery for every case, not just when Show Label is enabled (giving the glyph inside the battery another color still can lead to bad visibility as I've found). I also switched the position of the glyph and the label text around.

Before:
Screenshot_2026-07-26 20-38-39

After:
Screenshot_2026-07-26 20-38-23

The glyph is also shown if Show Label is disabled:
Screenshot_2026-07-26 20-39-07

Additionally, I removed the coloring of the battery nub if the battery is in warning state:
Screenshot_2026-07-26 20-23-19

### DIFF
--- a/src/shell/bar/widgets/battery_widget.cpp
+++ b/src/shell/bar/widgets/battery_widget.cpp
@@ -111,7 +111,7 @@ void BatteryWidget::createGraphicMode() {
   container->addChild(
       ui::box({
           .out = &m_bodyBg,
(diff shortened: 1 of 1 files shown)


ItsLemmy:
Thanks!

Issue #2896, opened 2026-06-04
[BUG] Polkit panel shows no indication that fingerprint authentication is in progress


TheSimpleZ:
### Submission checklist

- [x] I have searched existing issues and confirmed this is not a duplicate.
- [x] I am using the latest available version of Noctalia.

### Bug description

When shell.polkit_agent = true and pam_fprintd is configured before pam_unix in the polkit-1 PAM service, triggering a polkit action shows no dialog during the fingerprint verification window. The password dialog correctly appears afterward if fingerprint times out, but during the fingerprint phase there is no visual feedback at all — the user has no way to know they should touch the sensor, or that anything is happening.

**Full flow:**

1. Polkit action triggered
2. `pam_fprintd` runs — sends `PAM_TEXT_INFO` ("Place your finger on the fingerprint reader") — **no panel appears**
3. If user touches sensor in time → auth completes silently, no dialog ever shown
4. If fingerprint times out (~30s) → `pam_unix` runs, sends `PAM_PROMPT_ECHO_OFF` → password dialog appears correctly

**Root cause:**

`pam_fprintd` sends `PAM_TEXT_INFO`, not `PAM_PROMPT_ECHO_OFF`. The `showInfoCallback` in the
polkit agent only calls `setSupplementary()` — it does not set `responseRequired`. The
`stateCallback` in `application.cpp` gates panel opening on `isResponseRequired()`:

```cpp
if (needsInput) {
    if (!m_panelManager.isOpenPanel("polkit")) {
        m_panelManager.openPanel("polkit", ...);  // ← never reached for fingerprint
    }
} else if (m_panelManager.isOpenPanel("polkit")) {
    m_panelManager.refresh();  // ← panel isn't open, so also a no-op
}
```
The `PAM_TEXT_INFO` message is stored in `supplementaryMessage` correctly, but the panel
never opens to display it.

### Steps to reproduce

1. Enable shell.polkit_agent = true
2. Configure polkit-1 PAM with pam_fprintd as sufficient before pam_unix
3. Trigger a polkit action (e.g. pkexec ls)
4. Observe: no dialog during fingerprint window; password dialog appears only after ~30s timeout

### Expected behavior

[...]

PR #3663, opened 2026-07-27
fix(polkit): open panel during fingerprint auth

### LINKED ISSUE #3656: [BUG] Polkit panel stays hidden for the whole fingerprint window (regression of #2896 in 1f32612)
### Submission checklist

This is a **regression** of #2896 (closed as completed on 2026-06-12 by 07acb7c). It was reintroduced by
1f32612 `fix(polkit): adjust height based on content` (2026-07-15), first shipped in v5.0.0-beta.3, and is
still present in v5.0.0-beta.5. #2896 already has confirmations from four other users
(@gabrielrinaldi, @mtriam, @gitjuice, @alexreid) and a request to reopen, so I'm filing a fresh report
with the root cause on the new commit rather than adding a fifth "still broken" comment.

### Bug description

With `shell.polkit_agent = true` and `pam_fprintd` ahead of `pam_unix` in the `polkit-1` PAM stack, any
polkit action (`pkexec`, or an app requesting a privileged action) shows **no panel at all** during the
fingerprint verification window. The reader is armed and waiting the entire time, but nothing on screen
says so. After the ~30s fprintd timeout, `pam_unix` runs and the password panel appears normally.

Symptomatically this is indistinguishable from a hang, which is what makes it worse than a cosmetic
issue: the affordance to complete auth in two seconds exists but is invisible, so the user waits out a
timeout instead.

**Root cause.** 1f32612 added an early return to the state callback in
`src/app/application_services.cpp` (line 306 as of beta.5):

```cpp
m_polkitIdleCloseTimer.stop();
// Open once the session asks for a response so preferredHeight includes the
// password field. BeginAuthentication alone still has responseRequired=false.
if (!m_polkitAgent->isResponseRequired() && !m_panelManager.isOpenPanel("polkit")) {
  return;
}
```

`PolkitAgentSession` emits three distinct signals: `request`, `show-info`, and `show-error`.
`pam_fprintd` never emits `request`. It emits only `show-info` ("Place your finger on the fingerprint
reader") and then blocks on the reader. In `src/dbus/polkit/polkit_agent.cpp`, `showInfoCallback` routes
to `setSupplementary(text, false)`, which sets `supplementaryMessage` and calls `emitStateChanged()` but
leaves `responseRequired` false.

So every `show-info` during the fingerprint window is stored correctly and then discarded, because the
gate returns before a panel exists to render it. Only when fprintd gives up does `pam_unix` emit a real
`request`, flipping `responseRequired` and letting the panel open.

Worth noting the gate also renders existing code unreachable. `PolkitPanel::preferredHeight()`
(polkit_panel.cpp:117) and the matching block in the panel's update path both contain a
`!isResponseRequired() && !supplementaryText.empty() && !supplementaryError` branch that promotes the
supplementary message into the prompt slot and hides the input field. That is exactly the
fingerprint-panel presentation, and it can never execute, because the panel cannot be open in that
state.

### Steps to reproduce

1. `shell.polkit_agent = true`
2. Put `pam_fprintd` before `pam_unix` in `/etc/pam.d/polkit-1` (on NixOS:
   `security.pam.services."polkit-1".fprintAuth = true`)
3. Enroll a fingerprint (`fprintd-enroll`)
4. Run `pkexec ls` from a terminal in the graphical session
5. Observe: no panel for ~30s while the reader waits; password panel appears only after the fprintd
   timeout

### Expected behavior

The panel opens as soon as the session has something to display, and shows the `show-info` text so the
user knows to touch the reader. The input field stays hidden while `responseRequired` is false.
The code for that presentation already exists (see above). It then transitions to the password prompt if
fingerprint auth fails or times out.

### Actual behavior

No panel during the fingerprint phase. The `show-info` text is captured into `supplementaryMessage` and
dropped. Panel appears only once `pam_unix` requests a password.

### Suggested fix

The gate exists for a real reason and shouldn't just be deleted. `PanelManager::openPanel()` evaluates
`preferredHeight()` **once** and bakes the result into the layer-shell surface geometry;
`PanelManager::refresh()` (panel_manager.cpp:1555) only calls `m_surface->requestUpdate()`, a repaint,
and there is no other resize entry point. Opening on bare `BeginAuthentication` sizes the panel against
an empty conversation, which is the sizing bug 1f32612 set out to fix.

Changing the gate's *condition* rather than removing it addresses both. Open once the session has
produced something to display, which is `show-info` for a fingerprint and `request` for a password:

```cpp
const bool hasContentToShow =
    m_polkitAgent->isResponseRequired() || !m_polkitAgent->supplementaryMessage().empty();
if (!hasContentToShow && !m_panelManager.isOpenPanel("polkit")) {
  return;
}
```

This works because `setSupplementary()` assigns `supplementaryMessage` before calling
`emitStateChanged()`, so the text is available when the state callback runs and `preferredHeight()`
measures the real string. It also makes the dead `!isResponseRequired()` branches in `preferredHeight()`
and the update path reachable, so the panel renders the info message in the prompt slot with the input
hidden.

I have been running this as a downstream patch on v5.0.0-beta.5 for a few days on a Framework laptop
(Sway, goodix reader). Behavior: panel appears immediately with "Place your finger on the fingerprint
reader" and no password field, auth completes in about two seconds, and the panel transitions to the
password prompt correctly on timeout or a failed scan. Happy to open it as a PR if the approach looks
right.

**Limitation, and why this isn't a complete fix.** It inherits the same one-shot-sizing weakness
upstream already has: the panel is sized against whichever content is present at open time, so if the
fingerprint message is shorter than a later password prompt or error, the later content is sized to the
shorter one. In practice both are one line, so it doesn't bite, and `preferredHeight()` already
unconditionally reserves prompt-plus-input chrome (see the comment at polkit_panel.cpp:127), which
absorbs most of the difference. A genuinely complete fix needs a resize path (recomputing
`preferredHeight()` and reconfiguring the surface from `refresh()`), which means reworking the ~500-line
geometry/anchoring block in `openPanel()` that is currently driven by a one-time `panelHeight`. That
felt out of scope for a regression fix, but it is the real ceiling here, and it's also what any
multi-phase PAM conversation (fingerprint then password then error) ultimately wants.

### Workarounds, for anyone landing here before this is fixed

Both are worse than the patch, in different ways.

1. **Pre-open the panel before invoking `pkexec`.** Re-read the gate: an already-open panel skips the
   early return and falls through to `refresh()`. So `noctalia msg panel-open polkit && pkexec ...`
   renders correctly with no patching. Closing it again is the awkward part: polkit caches
   authorizations (`org.freedesktop.policykit.exec` is `auth_admin_keep`), so a second invocation inside
   the cache window never starts a PAM exchange at all, and a naively pre-opened panel then sits on
   screen for the entire duration of the privileged command. Watching for `polkit-agent-helper-1` to
   appear and exit works, with a few-second bail-out if it never appears. Fundamental limitation: it
   only reaches callers you wrap, so `nh`, `systemctl`, GUI apps, and ad hoc `pkexec` all still hit the
   blank window.
2. **Disable fingerprint for polkit** (`fprintAuth = false` on NixOS). The password panel appears
   instantly and there is no dead window, at the cost of losing biometric auth for `pkexec` entirely.

The `/etc/pam.d/polkit-1` reordering posted in #2896 (`pam_unix` sufficient, then `pam_fprintd`) also
surfaces a panel immediately, but it changes the auth order for every polkit action on the system and
requires pressing Enter to reach the fingerprint stage.

### Noctalia version

v5.0.0-beta.5 (`3d7b986`), plus the one-line downstream patch above. Unpatched beta.3, beta.4, and
beta.5 all reproduce; beta.2 does not.

### Compositor

Sway

### Distribution

NixOS

### Environment information

```text
Noctalia: v5.0.0-beta.5 (3d7b9869950592ff7cf3704f6a53afb169d850db)
OS: NixOS 26.11.20260711.e7a3ca8 (Zokor), kernel 6.12.95
Compositor: sway 1.12
fprintd: 1.94.5
polkit: 127
Hardware: Framework Laptop 13 (12th Gen Intel), goodix fingerprint reader
```

### Additional context

Also affects `nh` on NixOS, which elevates through `pkexec`. A system rebuild appears to hang
indefinitely with no dialog. Given that #2896 was verified with a touch-required security key rather
than a fingerprint reader, a `pam_fprintd` case in whatever manual test pass covers polkit would catch
future recurrences of this specific gate, since the security-key path evidently does emit `request`.

AdityaKr015:
## Summary

Fixed the polkit state callback gate to open the panel when the session has content to display, not only when a password response is required. `pam_fprintd` emits show-info (stored as supplementaryMessage, responseRequired=false) but never request, so the old gate kept the panel invisible during the entire ~30s fingerprint timeout.

## Motivation

Fixes a regression introduced in 1f32612 where the polkit panel stays hidden during fingerprint verification (pam_fprintd before pam_unix). The user sees nothing for ~30s while the reader is armed, then the password panel appears after fprintd times out. The existing presentation code for the fingerprint panel (promoting supplementaryMessage to the prompt slot while hiding the input field) was correct but unreachable because the gate prevented the panel from opening.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Build / packaging

## Related Issue

Closes #3656 

## Testing

- Ran `just format` on the branch then build with `meson compile -C build-release`, installed system with `sudo meson install -C build-release` for manual testing, restarted laptop.
- Tested on mounting the partitions and it shows the panel and partiton mounts correctly (see image below).
- Ran `pkexec ls` ls with fingerprint-first PAM, panel appears immediately with "Place your finger on the fingerprint reader" text.
- Normal password-only polkit actions still work (regression test for #2896)

## Manual Coverage

- [x] Tested on Niri
- [ ] Tested on Hyprland
- [ ] Tested on Sway

## Screenshots / Videos

image

> Panel when I mount my partitions

image

## Checklist

- [x] This PR is ready for review
- [x] I ran `just format` with clang-format v22+ installed
- [x] I ran the relevant build or test commands

### DIFF
--- a/src/app/application_services.cpp
+++ b/src/app/application_services.cpp
@@ -303,9 +303,13 @@ void Application::syncPolkitAgent() {
       return;
     }
     m_polkitIdleCloseTimer.stop();
-    // Open once the session asks for a response so preferredHeight includes the
-    // password field. BeginAuthentication alone still has responseRequired=false.
-    if (!m_polkitAgent->isResponseRequired() && !m_panelManager.isOpenPanel("polkit")) {
+    // Open once the session has content to display so preferredHeight accounts for
+    // the real text. BeginAuthentication alone has neither responseRequired nor
+    // supplementaryMessage — skip it. show-info (fingerprint) sets supplementaryMessage
+    // without flipping responseRequired; request sets responseRequired without
+    // supplementaryMessage. Either is sufficient content to show.
+    const bool hasContent = m_polkitAgent->isResponseRequired() || !m_polkitAgent->supplementaryMessage().empty();
+    if (!hasContent && !m_panelManager.isOpenPanel("polkit")) {
       return;
     }
     if (!m_panelManager.isOpenPanel("polkit")) {


Ly-sec:
Hey @AdityaKr015 , I'll merge it thank you very much :).

AdityaKr015:
Thank you for the merge

PR #3673, opened 2026-07-27
widgets/sysmon: migrate to typed widget definition

Ape:
Follow up to #3642.


ItsLemmy:
Thanks

Ape:
@ItsLemmy, I'm taking a bit of a break from continuing these migration for now. I will return to them at some point, but if you were eager to pick up here feel free to do so.

Issue #1627, opened 2026-01-29
Problem with core contrast in light themes


dinhokusanagi:
<img width="260" height="434" alt="Image" src="https://github.com/user-attachments/assets/ab0adde8-bc6e-48ac-aba2-ada5b8ad2974" />

<img width="1014" height="466" alt="Image" src="https://github.com/user-attachments/assets/12157121-6a3d-4f75-b3c1-0a1c1a1f3375" />

<img width="459" height="454" alt="Image" src="https://github.com/user-attachments/assets/153d1e59-d356-4af3-8c52-309dc0edb8c7" />

<img width="454" height="434" alt="Image" src="https://github.com/user-attachments/assets/4d003070-76c5-42df-afda-f1e3466d9853" />

<img width="465" height="439" alt="Image" src="https://github.com/user-attachments/assets/681f8afa-a0db-432a-ba33-57009135a544" />

<img width="319" height="271" alt="Image" src="https://github.com/user-attachments/assets/a4228d55-f1e0-46fe-8d6b-5494ece3ac7d" />

Light themes are nice, but there are three common contrast issues:

- Light text on a light background
- Insufficient contrast between foreground surfaces and the background
- Loss of visual clarity due to excessive shadows and heavy font weights

ItsLemmy:
I fixed the shadow on the mediacard.

The rest of this bug report is bit pointless as you did not specifiy which color scheme or which app were used.

dinhokusanagi:
Apps:
- Kitty
- Zen Browser

Themes:
- Everforest
- Catppuccin Lavender

It’s clear that issues like this exist at a broader level, not only in these specific cases. A general review is needed, as the color distribution logic is inconsistent across different surfaces and components.

ItsLemmy:
It's very difficult to do a good light theme.  Potential issues are most likely not  in Noctalia code per say, but in the data : color schemes + app templates.

PRs welcomed !

bennypowers:
Which commit fixed this? The issue persists for me in the kitty template. NB that the dark theme is _fine_. If we could configure a separate light-mode theme, that would probably fix it

<img width="1299" height="2000" alt="Image" src="https://github.com/user-attachments/assets/6e6bc8b7-e272-4dfd-a6e3-53624ff3b4d0" />

Something like this would let authors tweak the light theme colours:

```patch
diff --git a/assets/templates/builtin.toml b/assets/templates/builtin.toml
[...]

PR #3687, opened 2026-07-28
feat(templates): use OKLCH for lightness transformations for improved contrast

bennypowers:
## Summary

Addresses the kitty-specific contrast issues from #1627.

The existing `kitty.conf` uses raw M3 dark-palette token values, which produce low contrast on light terminals. This adds `kitty-light.conf` selected when `defaultMode == "light"` via `input_path_modes` in `builtin.toml`.

Key changes in the light template vs the dark one:

- Normal ANSI colors (1-6): `set_lightness 40` (OkLCH L=0.40) on `terminal_normal_*` tokens - dark enough to read on light backgrounds (~8:1 contrast ratio)
- color7: `on_surface_variant` (medium gray), replacing the near-white `terminal_normal_white`
- color8: `surface_variant` (light, theme-tinted) instead of `on_surface` (near-black) - meant for code-span/badge backgrounds with dark text on top
- Bright colors (9-14): `set_lightness 65` (OkLCH L=0.65) - medium-light so they work as highlight/badge backgrounds
- color15: `terminal_background`, matching the actual terminal bg
- Selection foreground: `on_surface` for guaranteed dark text on selection

## Test plan

- [ ] Apply a light-mode noctalia theme and confirm `noctalia msg templates-apply` writes `kitty-light.conf` output
- [ ] Verify ANSI colors 1-6 are dark enough to read as text on the light background
- [ ] Verify ANSI colors 8-14 are light enough to use as badge/highlight backgrounds
- [ ] Test across multiple M3 palette hues (pink, green, blue primaries)

Relates to #1627

### DIFF
--- a/assets/templates/builtin.toml
+++ b/assets/templates/builtin.toml
@@ -130,9 +130,12 @@ output_path = "$XDG_DATA_HOME/color-schemes/noctalia.colors"
 post_action = "kde-color-scheme"
 
 [templates.kitty]
-input_path = "./kitty/kitty.conf"
 output_path = "$XDG_CONFIG_HOME/kitty/themes/noctalia.conf"
 post_hook = "bash '{{ config_dir }}/kitty/apply.sh'"
+
+[templates.kitty.input_path_modes]
+dark = "./kitty/kitty.conf"
+light = "./kitty/kitty-light.conf"
 
 [templates.labwc]
 input_path = "./labwc/labwc.conf"
--- a/assets/templates/kitty/kitty-light.conf
+++ b/assets/templates/kitty/kitty-light.conf
@@ -1,18 +1,18 @@
 color0 {{colors.on_surface.default.hex}}
-color1 {{colors.terminal_normal_red.default.hex | set_lightness 40}}
-color2 {{colors.terminal_normal_green.default.hex | set_lightness 40}}
-color3 {{colors.terminal_normal_yellow.default.hex | set_lightness 40}}
-color4 {{colors.terminal_normal_blue.default.hex | set_lightness 40}}
-color5 {{colors.terminal_normal_magenta.default.hex | set_lightness 40}}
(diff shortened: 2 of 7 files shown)


ItsLemmy:
This might work for palettes that where generated from wallpapers but it will absolutely destroy all the built-in ones where the contrast is fine and where colors are canonical.

bennypowers:
> contrast is fine

Contrast issues occur for the bespoke themes as well: All 6 FG colors fail WCAG 2 AA for Dracula, Nord, and Ayu light themes; 4 for Noctalia theme. `set_lightness 40` would help for those. I would recommend using OKLCH lightness, which appears more natural in how they preserve hue during lightness transformations. If you're interested, I can work up a commit which uses OKLCH.

If we proceed assuming that a solution for low contrast in non-automated themes is out of scope of this PR, could you accept a solution which exposes custom transformations or theme modes for automated themes?

ItsLemmy:
I'm not sure what I'm looking at in your screenshot #1627.   This seems indeed very wrong, but I use kitty daily, and with light theme it's not perfect but it's not as bad as the fully effed up screenshot I see in yours. Are you sure it's not something on your system?

- Built-In and Community palette should not be altered, we have to see them as brand color guide and respect their color code. So the fix is not in the templates.   If people have contrast/readability with a palette, they can use another one or open PRs if they see a deviation.  Most of those terminal colors come from the standard iTerm2 colors.

- Changing `set_lightness` behavior is also a no go, since we wanna stay as close as possible from matugen implementation, and this would  affect everyone templates, to fix a "you" issue.

- Fixing kitty only does not make any sense,  since other terminals will be left behind with the same issue.

A better approach would be to only target the wallpaper based palette generators and ensure better contrast there.  I may take a look at it soon.

ItsLemmy:
Fixed with my latest commits 
 e810c1d68e3a64994934589eee6af4bc41a1b9ca
 64bc063b64c6e21c40643d6a7a027b536df5ac3a

PR #3688, opened 2026-07-28
feat(plugins): add service enable and exit reasons

Gustav0ar:
## Summary

- add service `onEnable()` after an explicit successful plugin enable
- pass `shutdown`, `disable`, `uninstall`, and `reload` to every entry `onExit(signal, reason)`
- keep one-argument handlers and plugins below API 17 compatible
- bump the plugin API from 16 to 17

## Why

Plugins need to distinguish explicit lifecycle actions from reloads and shell shutdown.

## Testing

- `meson test -C build-debug --print-errorlogs` (62 passed)

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1137,6 +1137,7 @@ if build_tests
     'plugin_catalog',
     'plugin_git_export',
     'plugin_i18n',
+    'plugin_lifecycle',
     'plugin_manifest',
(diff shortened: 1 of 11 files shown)


ItsLemmy:
Thanks, I'll take care of update the official-plugins repo with noctalia.d.luau

PR #3689, opened 2026-07-28
[Bug]Fix launcher/clipboard panels covering fcitx5 preedit on Hyprland

shinnkka:
## Summary

On Hyprland, noctalia's panels and fcitx5's preedit window both use the Overlay layer-shell layer, causing noctalia's panels to render above and completely obscure the IME preedit box — users cannot see what they are typing.
This PR moves user-interactive panels from Overlay to Top layer — the semantically correct layer for taskbars, launchers, and popup panels. System overlays that do not accept text input (polkit, OSD, window switcher) remain on Overlay.

## Motivation

https://github.com/noctalia-dev/noctalia/issues/3438
https://github.com/noctalia-dev/noctalia/issues/2786

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Build / packaging

## Manual Coverage

- [ ] Tested on Niri
- [x] Tested on Hyprland
- [ ] Tested on Sway

## Screenshots / Videos

before:
图片

after:
图片

## Checklist

### DIFF
--- a/src/shell/clipboard/clipboard_panel.h
+++ b/src/shell/clipboard/clipboard_panel.h
@@ -38,7 +38,7 @@ public:
 
   [[nodiscard]] float preferredWidth() const override { return scaled(720.0f); }
   [[nodiscard]] float preferredHeight() const override { return scaled(560.0f); }
-  [[nodiscard]] LayerShellLayer layer() const override { return LayerShellLayer::Overlay; }
+  [[nodiscard]] LayerShellLayer layer() const override { return LayerShellLayer::Top; }
   [[nodiscard]] LayerShellKeyboard keyboardMode() const override { return LayerShellKeyboard::Exclusive; }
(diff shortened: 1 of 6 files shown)


ItsLemmy:
Unfortunately this is not an acceptable fix.

ItsLemmy:
This is a hyprland take where they assume IME popup should be rendered between Top and Overlay layers.
https://github.com/hyprwm/Hyprland/blob/main/src/render/Renderer.cpp#L1239-L1245

Other compositors render IME popups just after their parent surface, which is the better approach as it ensure it will be visible every time.

I'm gonna close this PR, as there is no reason that people that don't use IME or Hyprland should loose their beloved launcher on the overlay layer (ex: when watching a movie) because a compositor decided IME should be under the overlay layer.

Instead I'll add a way to switch all those panels from Overlay layer (default) to Top layer so you can mitigate the issue.

ItsLemmy:
image

PR #3695, opened 2026-07-29
fix(wayland): keep toplevels without output_enter in output-scoped lists

GhostEagle68:
## Summary

On Hyprland, toplevels that already exist when a client binds `zwlr_foreign_toplevel_manager_v1` are announced normally but never receive an `output_enter` event, so `WaylandToplevels::ToplevelState::output` stays null for them.

`allAppIds()`, `windowsForApp()`, and `windowsWithoutAppId()` all filtered with `outputFilter != nullptr && state.output != outputFilter`, which excluded those toplevels from every output-scoped query. They then reached the taskbar only via the ext-foreign-toplevel list, with `handle == nullptr` and the taskbar's non-KDE context-menu and activation paths require a wlr handle, so they silently did nothing.

This adds a `sawOutputEnter` flag on `ToplevelState` (set on the first `output_enter`, never cleared) so "never announced an output" can be distinguished from "left every output", plus a `matchesOutputFilter()` helper encoding the rule:

- no filter given → matches
- output matches the filter → matches
- never announced an output → matches
- announced an output and later left all of them → **still excluded**, as before

All three output-scoped queries use the helper.

It also corrects a comment in `compositor_platform.cpp` which claimed Hyprland never exports these windows over `zwlr_foreign_toplevel_management`. It does export them; just omits `output_enter` instead.

## Motivation

After restarting noctalia mid-session on Hyprland, taskbar tiles for windows that were already open stop responding: right-click shows no context menu, and left-click doesn't activate the window or switch workspace. Windows opened *after* the restart behave normally.

Reproduces on v5.0.0-beta.6 (AUR) and on `main`. It mainly affects anyone who restarts the shell. So, mostly developers but it's a real gap in the wlr path.

Root cause evidence, using `lswt` (an unrelated client that binds both protocols) while several windows were already open:

```
$ WAYLAND_DEBUG=1 lswt 2>&1 | grep -cE 'zwlr_foreign_toplevel_manager_v1#[0-9]+\.toplevel'
6
$ WAYLAND_DEBUG=1 lswt 2>&1 | grep -cE 'zwlr_foreign_toplevel_handle_v1#[0-9]+\.output_enter'
0
```

Six toplevels announced over wlr, zero `output_enter` events.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Build / packaging

## Related Issue

None filed. Happy to open one if you'd prefer it tracked separately though.

## Testing

```
just format
just build
just test
```

`wayland_toplevels_identity` passes with three new assertions covering:

1. never announced an output → matches an output filter
2. announced a different output → excluded
3. announced an output, then left all of them → excluded

Note: `i18n_supported_languages` currently fails on `main` independently of this PR (`ko.json` exists in `assets/translations/` but is missing from `kSupportedLanguages`).

Verification on Hyprland: opened several apps, restarted noctalia, then confirmed right-click context menus and left-click activate/workspace-switch work again on those pre-existing windows.

## Manual Coverage

- [ ] Tested on Niri
- [x] Tested on Hyprland
- [ ] Tested on Sway

## Checklist

## Additional Notes

**Multi-monitor is untested.** I only have one monitor. With this change, a pre-existing window will appear on *every* output's taskbar until it announces an output, instead of appearing on none. I think present-on-all beats absent-and-inert, and it self-corrects as soon as the window is touched. But if you'd prefer a different fallback for multi-output setups, let me know.

`sawOutputEnter` is deliberately sticky. Without it, `output == nullptr` would also cover windows that left every output (see `onHandleOutputLeave`), and those would have started showing up on all outputs for every user on every compositor. The flag keeps this change inert for compositors that send `output_enter` properly.

**Disclosure:** I used Claude Code as a pair-programming assistant on this one. It suggested where to instrument and proposed comparing the two toplevel protocols with `WAYLAND_DEBUG`, which is what surfaced the missing `output_enter`. The reproduction, all testing and verification on my machine, and the final code are mine, but the diagnostic path was a collaboration, and it seemed worth disclosing this.

### DIFF
--- a/src/compositors/compositor_platform.cpp
+++ b/src/compositors/compositor_platform.cpp
@@ -142,6 +142,42 @@ namespace {
     return false;
   }
 
+  std::unordered_set<std::string> outputScopedWindowIds(const WaylandWorkspaces* workspaces, wl_output* output) {
+    std::unordered_set<std::string> ids;
+    if (output == nullptr || workspaces == nullptr) {
+      return ids;
+    }
+    for (const auto& row : workspaces->workspaceWindows(output)) {
+      const auto normalized = compositors::hyprland::normalizeWindowId(row.windowId);
+      if (!normalized.empty()) {
+        ids.insert(normalized);
+      }
+    }
+    return ids;
+  }
+
+  // Toplevels whose output the compositor never announced stay in every output-scoped wlr list, so
+  // the IPC window list is what actually places them: drop the ones owned by another output.
+  void dropWindowsOnOtherOutputs(
+      std::vector<ToplevelInfo>& windows, const compositors::hyprland::HyprlandToplevelMapping& mapping,
+      const std::unordered_set<std::string>& outputWindowIds
+  ) {
+    if (outputWindowIds.empty()) {
+      return;
+    }
+    std::erase_if(windows, [&](const ToplevelInfo& window) {
+      if (window.outputAnnounced || window.handle == nullptr) {
+        return false;
+      }
+      const auto windowId = mapping.windowIdForWlrHandle(window.handle);
+      if (!windowId.has_value()) {
+        return false;
+      }
+      const auto normalized = compositors::hyprland::normalizeWindowId(*windowId);
+      return !normalized.empty() && !outputWindowIds.contains(normalized);
+    });
+  }
+
   void appendHyprlandExtOnlyWindows(
       std::vector<ToplevelInfo>& windows, const std::vector<ToplevelInfo>& extWindows,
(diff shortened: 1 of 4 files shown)


GhostEagle68:
Just verified again by A/B on one session: uninstalled the patched build to fall back to stock v5.0.0-beta.6 (856d99bbd199). Context menus and activation dead on pre-existing windows; reinstalled the patched build — both work again. Same windows, same config, only the binary differed.

ItsLemmy:
Thanks!

PR #3704, opened 2026-07-30
fix(control-center/calendar): use nominative case for month's name

MaksRawski:
## Summary
Use nominative instead of genitive case for month name displayed in the calendar tab of the control center.

## Motivation
Month name that isn't used as part of further date formatting with a day should be written in nominative, and `%B` uses genitive.
With English locale this is not a problem, but it's as if it was written "of July 2026" instead of just "July 2026".

## Type of Change
- [X] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Build / packaging

## Testing
`just test` passes. Manually verified.

## Manual Coverage

- [ ] Tested on Niri
- [X] Tested on Hyprland
- [ ] Tested on Sway

## Checklist

## Additional Notes

[`std::strftime` doesn't have an `OB` specifier](https://en.cppreference.com/cpp/chrono/c/strftime) even though [C23's `strftime`](https://en.cppreference.com/c/chrono/strftime) has. Since the standard doesn't mention it, it technically may not be available, however I assume that any implementation of C++'s `std::strftime` just calls C's `strftime` and since noctalia depends on GLIBC anyway, I further assume that GLIBC is going to be used for that and [`%OB` has been in GLIBC since 2.27 (released in 2018)](https://lists.gnu.org/archive/html/info-gnu/2018-02/msg00000.html).

Maybe an additional check for a GLIBC version should be done in the build system?

### DIFF
--- a/src/shell/control_center/tabs/calendar_tab.cpp
+++ b/src/shell/control_center/tabs/calendar_tab.cpp
@@ -54,7 +54,7 @@ namespace {
     std::tm tm{};
     tm.tm_mon = month;
     tm.tm_mday = 1;
-    return formatStrftime("%B", tm);
+    return formatStrftime("%OB", tm);
   }
 
   int daysInMonth(int yearValue, int monthValue) {
--- a/src/shell/control_center/tabs/calendar_tab.cpp
+++ b/src/shell/control_center/tabs/calendar_tab.cpp
@@ -54,7 +54,11 @@ namespace {
     std::tm tm{};
     tm.tm_mon = month;
     tm.tm_mday = 1;
+#ifdef __GLIBC__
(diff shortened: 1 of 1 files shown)


ItsLemmy:
Since Noctalia supports non-glibc builds, could we use %OB under 
#ifdef __GLIBC__ and keep %B otherwise?     

I don't believe we need to check for older GLIBC versions.

MaksRawski:
Oh ok, I quickly skimmed through `meson.build` yesterday late at night and thought it was a hard dependency, but that's cool to hear that it's not! Anyway, just pushed the fix you mentioned.

ItsLemmy:
THanks!

PR #3705, opened 2026-07-30
fix(taskbar): focus Niri windows by compositor ID

eaceaser:
## Summary

Use Niri’s compositor window IDs to give each taskbar entry a stable identity for ordering, focus state, and window actions.

On `main`, the taskbar builds its window list from the generic foreign-toplevel interface and reconciles those entries with compositor workspace data separately. This works when a foreign-toplevel handle is sufficient, but Niri’s workspace layout and focus operations are addressed by numeric window ID. Multiple windows from the same application can therefore be assigned by less-specific metadata such as app ID and title.

This change adds a Niri-specific taskbar path that:

- Reads exact window IDs, workspace placement, focus state, and layout coordinates from Niri IPC.
- Uses the matching `ext-foreign-toplevel` identifier as the immutable identity stored by each task.
- Filters taskbar windows to the IDs reported for the bar’s output.
- Orders tasks from Niri’s current workspace and layout positions.
- Activates and closes Niri windows by compositor ID.
- Selects the focused-window indicator using Niri’s focused window ID.
- Recomputes the model when Niri reports window layout changes.
- Keeps foreign-toplevel property updates pending until the protocol’s `done` event, so the taskbar only consumes complete records.
- Binds `ext-foreign-toplevel-list-v1` on Niri alongside the existing WLR foreign-toplevel manager, making the stable identifier available while preserving richer WLR state used elsewhere.

On Niri, Noctalia intentionally binds both foreign-toplevel protocols. The WLR manager continues providing richer state used by other shell features, while the ext-foreign-toplevel identifier provides the exact join to Niri’s numeric IPC window ID. When either side of that join is temporarily unavailable, the taskbar waits for the next protocol or IPC update instead of falling back to ambiguous app/title matching. Existing behavior remains unchanged on other compositors.

## Motivation

On Niri, clicking a taskbar entry does not always focus the window represented by that entry. The problem is most noticeable with several windows from the same application and with windows on secondary monitors or other workspaces.

For example, two Ghostty windows can appear in one left-to-right order while clicking their taskbar entries focuses them in the opposite order. The focused-window indicator can consequently appear under the wrong entry as well.

Niri exposes the information needed to avoid this ambiguity:

- A stable numeric ID for every window.
- The focused window’s ID.
- Each window’s workspace and position in the scrolling layout.
- An equivalent identifier through `ext-foreign-toplevel`.

Using that ID throughout the taskbar model ensures that displayed position, focused state, and click target all refer to the same compositor window.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Build / packaging

## Related Issue

N/A

## Testing

- `just format`
- `just lint debug` — passed with warnings treated as errors
- `just build debug`
- `just test debug --print-errorlogs` — 69/69 tests passed
- Added regression coverage for:
  - Niri window identity and workspace assignment
  - Layout-position updates
  - Exact compositor-ID preservation during reconciliation
  - Focus changes based on compositor window ID
  - Multiple windows with identical app IDs and titles
  - Atomic foreign-toplevel property publication on `done`
  - Pending and committed foreign-toplevel removal
  - Compositor-specific ext-foreign-toplevel binding policy

## Manual Coverage

- [x] Tested on Niri
- [ ] Tested on Hyprland
- [ ] Tested on Sway

## Screenshots / Videos

N/A — this changes taskbar window association and interaction behavior without changing its visual design.

## Checklist

- [x] Formatting was validated with clang-format v22.
- [x] I ran the relevant build and test commands.
- [x] This PR does not require end-user documentation changes.

## Additional Notes

gpt-5.6-sol was used extensively in debugging and implementation. I reviewed + iterated on the implementation to try to not overcomplicate this patch, but I did some digging around how other bars with Niri-specific behavior manage this, and they seem to just rely on Niri's specific IPC interface vs. using Wayland's generic interface. This avoids all ambiguity and issues with joining the Wayland vs. Niri representations together. I don't have enough context on the direction Noctalia is trying to take to know if that's a refactor worth exploring, but it does seem to be cleaner as far as the data model is concerned when using Niri.

### DIFF
--- a/src/compositors/compositor_platform.cpp
+++ b/src/compositors/compositor_platform.cpp
@@ -62,6 +62,23 @@ namespace {
 
   constexpr Logger kLog("compositor_platform");
 
+  [[nodiscard]] std::unordered_set<std::string>
+  windowIdsFromAssignments(const std::vector<WorkspaceWindowAssignment>& assignments) {
+    std::unordered_set<std::string> windowIds;
+    windowIds.reserve(assignments.size());
+    for (const auto& assignment : assignments) {
+      if (!assignment.windowId.empty()) {
+        windowIds.insert(assignment.windowId);
+      }
+    }
+    return windowIds;
+  }
+
+  void
+  retainToplevelsWithWindowIds(std::vector<ToplevelInfo>& windows, const std::unordered_set<std::string>& windowIds) {
+    std::erase_if(windows, [&](const ToplevelInfo& window) { return !windowIds.contains(window.identifier); });
+  }
+
   [[nodiscard]] const char* valueOrUnset(const char* value) {
     return value != nullptr && value[0] != '\0' ? value : "<unset>";
   }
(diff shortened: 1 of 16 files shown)


src/compositors/compositor_platform.cpp
```diff
@@ -875,11 +892,56 @@ std::vector<ToplevelInfo> CompositorPlatform::windowsWithoutAppId(wl_output* out
   return windows;
 }
 
+std::vector<ToplevelInfo> CompositorPlatform::taskbarWindowsForApp(
+    const std::string& idLower, const std::string& wmClassLower, wl_output* outputFilter,
+    const std::unordered_set<std::string>* allowedNiriWindowIds
```
eaceaser: just noticed this... adding something niri specific here doesn't seem right

ItsLemmy:
Thanks, merging this as is, but I have another commit on top that will keep things more agnostic and avoid having a bunch of if(compositor) where  they dont belong.

PR #3715, opened 2026-07-30
Feat(taskbar): drag to reorder pinned taskbar applications 

GhostEagle68:
## Summary

Adds drag-to-reorder for pinned tiles in the taskbar widget: press a pinned tile, drag it along the bar, and drop it in a new position. The new order is written to `widget.<name>.pinned`, so it persists.

How it works:

- **Arming** — a drag starts once the pointer travels 8px along the bar while pressed, or after a 300ms hold if you press without moving. Plain clicks, middle-click close, and right-click context menus are unaffected.
- **Visuals** — the dragged tile leaves the layout flow (`setParticipatesInLayout(false)`) so `Flex` stops repositioning it, and an invisible spacer of the same size is inserted at the target index to hold the drop gap open. Neighbours reflow around it, and the strip keeps its width. On drop the tile is parked in the gap so the strip looks settled while the config write lands.
- **Persistence** — on release, the pin list is reordered and written via `setOverride`, deferred through `DeferredCall::callLater` because the write rebuilds the taskbar and would otherwise destroy the `InputArea` whose handler is still running.

Reordering is enabled only when there are at least two pinned entries, the widget has a config name to persist to, and `group_by_workspace` is off (in that mode the flat pinned order isn't what's on screen).

## Motivation

Pinned tiles land in whatever order they were pinned, and the only way to rearrange them is to unpin and re-pin in sequence. Dragging is what people expect from a taskbar.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Build / packaging

## Related Issue

None

## Testing

```
just format
just build
just test
```

On Hyprland with a horizontal bottom bar and 6 pinned apps: dragging in both directions, dropping at both ends, dropping back on the origin (no write), plain/middle/right clicks on pinned and unpinned tiles, and clicks immediately following a drag. Reordered state verified in `settings.toml` and after restarting noctalia.

https://github.com/user-attachments/assets/0bfee7e3-ff3d-4667-bde4-b1ebc6c5dd4a

Notice the unpinned apps (VSC and Claude) on the right are not affected.

## Manual Coverage

- [ ] Tested on Niri
- [x] Tested on Hyprland
- [ ] Tested on Sway

## Checklist

## Additional Notes

**Untested paths.** Only tested on a horizontal bottom bar with one output. The vertical-bar path is implemented (all geometry is expressed along a "main axis" that follows `m_vertical`). Also multi-monitor is likewise untested.

**No `--insertAt` correction.** `Dock::endDrag` decrements the insert position when the target is past the source, because its `computeDragTargetIndex` returns an *insertion slot*. This one returns a *destination index* (`sourceIndex + lround(travel / pitch)`, clamped), so after erasing the source, inserting at the target is already correct. Copying the dock's correction here would shift rightward drags one slot short.

**Drop latency.** There's a ~55ms gap between release and the rebuild, which is `setOverride` writing the file and running `loadAll()` — the same cost as any other settings change. The tile is parked in its final position first so this isn't visible, but I mention it in case the config reload path is something you'd want to look at separately.

**Disclosure:** I used Claude Code as a pair-programming assistant on this feature; for reading unfamiliar parts of the scene graph and layout code, and for review. The design decisions, testing, and verification are all mine.

### DIFF
--- a/src/shell/bar/widgets/taskbar_widget.cpp
+++ b/src/shell/bar/widgets/taskbar_widget.cpp
@@ -915,6 +915,8 @@ void TaskbarWidget::buildTaskButtons(Renderer& renderer) {
       if (m_drag.active) {
         m_suppressTileClick = true;
         if (commitDragReorder()) {
+          // Mark drag as inactive so doLayout() cleans up spacer even if config write fails.
+          m_drag.active = false;
           // The deferred config write rebuilds the strip and destroys these nodes; restoring them
           // now would briefly relayout at the old order and snap the tile back. Park the tile in
           // the gap the spacer is holding so the strip looks settled while that write lands.
@@ -922,7 +924,6 @@ void TaskbarWidget::buildTaskButtons(Renderer& renderer) {
             m_drag.area->setPosition(m_drag.spacer->x(), m_drag.spacer->y());
           }
           m_drag.area = nullptr;
-          m_drag.spacer = nullptr;
         }
       } else if (m_drag.armed) {
         m_suppressTileClick = true;
--- a/src/shell/bar/widgets/taskbar_widget.cpp
+++ b/src/shell/bar/widgets/taskbar_widget.cpp
@@ -42,9 +42,6 @@
 namespace {
 
   constexpr auto kDragHoldDelay = std::chrono::milliseconds(300);
-  // Main-axis travel that starts a drag outright, so picking a tile up does not have to wait out
-  // the hold delay. Large enough that an unsteady click cannot trip it.
-  constexpr float kDragArmDistance = 8.0F;
 
   // Integer centering; optional odd spare pixel on the end side (right/bottom).
   [[nodiscard]] float centeredOffset(float extent, float content, float inset = 0.0F, bool oddSpareOnEnd = true) {
@@ -939,7 +936,9 @@ void TaskbarWidget::buildTaskButtons(Renderer& renderer) {
           return;
         }
         const float main = pointerMainOnStrip(*dragArea, data.localX, data.localY);
-        if (!m_drag.armed && !m_drag.active && std::abs(main - m_drag.startMain) >= kDragArmDistance) {
+        if (!m_drag.armed
+            && !m_drag.active
+            && std::abs(main - m_drag.startMain) >= Style::dragStartThreshold * m_contentScale) {
           m_drag.holdTimer.stop();
           m_drag.armed = true;
         }
--- a/src/shell/bar/widgets/taskbar_widget.cpp
+++ b/src/shell/bar/widgets/taskbar_widget.cpp
@@ -920,6 +920,10 @@ void TaskbarWidget::buildTaskButtons(Renderer& renderer) {
           m_drag.area = nullptr;
           m_drag.spacer = nullptr;
         }
+      } else if (m_drag.armed) {
+        m_suppressTileClick = true;
+        // Armed but not active means the user pressed and held long enough to arm, but never moved far enough to start
+        // a drag. Treat it as a click that was suppressed.
       }
       endDragVisual();
       m_drag = {};
--- a/src/shell/bar/widgets/taskbar_widget.cpp
+++ b/src/shell/bar/widgets/taskbar_widget.cpp
@@ -385,9 +385,9 @@ void TaskbarWidget::beginDragVisual() {
   m_drag.restMain = m_vertical ? m_drag.area->y() : m_drag.area->x();
   m_drag.restCross = m_vertical ? m_drag.area->x() : m_drag.area->y();
   m_drag.pinnedCount = pinnedConfigIds().size();
-  m_drag.area->setParticipatesInLayout(false);
-  m_drag.area->setZIndex(200);
-  // Don't create spacer here; doLayout() will handle it.
+  if (Node* container = root(); container != nullptr) {
+    container->markLayoutDirty();
+  }
 }
 
 void TaskbarWidget::updateDragVisual() {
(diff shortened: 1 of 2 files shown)


GhostEagle68:
I do believe that #3635 would be beneficial merged together with this, althoughbot necessary. And I have a local build with that PR and this one and no issues together

ItsLemmy:
Would need to be rebased and tested on vertical bar.

virtuallyunknown:
Hey, sorry to bother you, I have a quick question. Would you, be interested in widening the scope of this PR?

I'm basically looking to migrate to Noctalia, but the one thing I really miss is a classic taskbar where you can re-order the windows via dragging and dropping, much like browser tabs.

Your PR is already halfway there, with the exception that it only concerns pinned app buttons (tiles), and there also needs to be some mechanism to inform the underlying compositor about the new order, so that it can arrange it's own windows in the same order that they appear in the taskbar widget (this is only relevant to scrolling wms like Niri).

Here is a very quick demonstration (bottom bar). Waybar + [niri_window_buttons](https://github.com/adelmonte/niri_window_buttons)

https://github.com/user-attachments/assets/f022bf0a-9bdf-4450-a814-2b248559177c

I asked about my issue on the Discord server, where I was made aware of this PR. And yeah, it's probably a long shot, but it doesn't hurt to ask as they say.

Cheers.

GhostEagle68:
> Would need to be rebased and tested on vertical bar.

@ItsLemmy Rebased and tested both horizontal and vertical


https://github.com/user-attachments/assets/77e05c83-e16a-4345-b69d-903af3ddce5d


https://github.com/user-attachments/assets/8a498791-3c01-4fcf-86e7-5e7e94ac40f2

Edit:
src/dbus/network/network_manager_service.cpp Seems to have missed just format from another commit, nothing here touched that :)

ItsLemmy:
### 1. [P1] Drag cleanup incorrectly depends on a config reload                                                          

taskbar_widget.cpp:916-924                                                                                               


commitDragReorder() returns true after merely scheduling setOverride(). The release handler then clears m_drag.area and  m_drag.spacer, making endDragVisual() a no-op. The tile remains excluded from layout and the spacer remains attached,  under the assumption that setOverride() will rebuild the taskbar.                                                        

That assumption is false when:                                                                                           

- validation or file persistence fails;                                                                                  
- settings.toml is unwritable;                                                                                           
- the resulting override table is unchanged, where commitOverrideTable() returns without loadAll().                      

The taskbar then retains transient drag state until an unrelated rebuild; later layouts can move the spacer without      
moving the excluded tile. Restore a valid scene independently of persistence, or safely handle the deferred result and   
explicitly recover on failure/no change.                                                                                 

### 2. [P1] Spacer changes violate the UI phase and layer boundaries                                                     

taskbar_widget.cpp:430-448                                                                                               

beginDragVisual(), endDragVisual(), and syncDragSpacer() change layout participation and insert/remove children directly from pointer callbacks. Structural scene changes belong in Layout; callbacks should update drag state and call           
requestLayout().                                                                                                         

The spacer is also created with ui::node() and stored as Node*. Shell code may use controls only; InputArea is the sole  
raw-node exception. Use an existing control such as Box for the spacer and create/reposition/remove it during            
doLayout().                                                                                                              

### 3. [P2] Releasing after the hold delay still launches the application                                                

taskbar_widget.cpp:913-927                                                                                               

After 300 ms, the timer sets m_drag.armed, but release only checks m_drag.active. A hold followed by release without     
motion therefore falls through to onClick() and activates or launches the pinned app.                                    

That contradicts the described hold-to-drag behavior and differs from the existing dock implementation, which consumes  an armed gesture on release. Treat armed as consumed, set click suppression, and cancel the drag without committing.     

### 4. [P2] The movement threshold ignores interface scale                                                               

taskbar_widget.cpp:942                                                                                                                                                                                            

kDragArmDistance is a fixed 8.0F, while task geometry scales with m_contentScale. At 2× interface scale, the effective   
tolerance is half as large relative to the tile, making normal click jitter more likely to arm a drag.                   

The canonical threshold already exists as Style::dragStartThreshold in logical pixels. Compare against                   
Style::dragStartThreshold * m_contentScale.

ItsLemmy:
Mark it ready for review, when done, thx

GhostEagle68:
Address the drag-to-reorder review findings and preserve the intended drop behavior:

Replace the fixed drag-arm distance with Style::dragStartThreshold * m_contentScale, so the gesture threshold matches interface scaling.

Treat a held-but-unmoved drag as consumed on release. Once the hold timer has armed the gesture, releasing does not fall through to the tile click handler and launch or activate the application.

Move structural drag changes into doLayout(). While a drag is active, layout removes the dragged InputArea from Flex participation, raises its z-index, and creates or repositions a transparent, non-interactive Box spacer in the task strip. When the drag ends, layout removes the spacer and restores the tile to normal layout participation. The task strip owns the spacer; m_drag.spacer is only a non-owning pointer.

Pointer motion still updates the position of the existing floating tile and requests redraw. When the calculated target slot changes, it records the new index and marks layout dirty so doLayout() moves the spacer.

On a successful reorder, the release handler parks the floating tile at the spacer’s final coordinates while the deferred config write runs. This avoids restoring it into the old Flex order for a frame, which previously caused the tile to jump to an edge before the reordered strip appeared.

commitDragReorder() queues the config write first. The release handler then queues a second deferred callback. Deferred calls execute FIFO: if the write caused a rebuild, rebuild() clears m_drag before replacing task-strip children and the second callback returns. If drag state remains active, the second callback ends the drag and marks layout dirty, allowing doLayout() to restore the tile and remove the spacer. This covers a failed or no-op config write without relying on persistence to clean up the scene.

rebuild() clears transient drag state before it destroys task-strip children, preventing stale drag pointers during a config-driven rebuild.

ItsLemmy:
Thanks

PR #3753, opened 2026-08-01
feat(shell-completions): add shell completions

oldirtty:
## Summary

Adds shell completions for bash, fish, and zsh via a new `noctalia completions <shell>` command.

The scripts are maintained under `assets/completions/` but are embedded directly into the binary at build time. To achieve this, it uses Meson's native `fs.read()` wrapped in C++ raw string literals via configure_file().

## Motivation

Improve the developer and user experience by providing completion for all CLI commands, flags, and dynamic values (such as plugins, panels, color schemes, and wallpaper generators).

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Build / packaging

## Testing

```
alias noctalia="$PWD/build-debug/noctalia"

# fish
noctalia completions fish | source

# bash
source <(noctalia completions bash)

# zsh
rm ~/.zcompdump
noctalia completions zsh > _noctalia
fpath=($PWD $fpath)
autoload -Uz compinit && compinit -D

noctalia <TAB>
```
## Manual Coverage

- [x] Tested on Niri
- [ ] Tested on Hyprland
- [ ] Tested on Sway

## Checklist

## Additional Notes

Package maintainers could have their packages to auto install completions.


oldirtty:
Took a look at how firefox-theme was done and did the same, no python dependencies anymore.

ItsLemmy:
Good idea, genuinely useful. But the implementation contradicts the PR's own stated design, duplicates the repo's canonical asset lookup, and ships three hand-copied command lists that were **already stale at submission time**.

---

## Blocking

### 1. The PR description is wrong — nothing is embedded

> "Instead of relying on external files that can get out of sync, the completions are embedded directly into the executable."

`src/completions/cli.cpp` opens `assetDir / filename` with `std::ifstream` and dumps it. It's a runtime file read from `$datadir/noctalia/assets/completions/`.

Either embed them for real (an `xxd -i`-style meson `custom_target`, or a raw string literal generator) or fix the claim. Right now a new binary + old installed assets tree silently emits stale completions — exactly the failure mode the PR says it avoids.

### 2. Fourth independent asset-resolution chain (canonical violation)

`completionsAssetDir()` re-rolls env → source → install → exe-relative resolution. The canonical helper already exists:

```cpp
// src/core/files/resource_paths.h
[[nodiscard]] std::filesystem::path assetPath(std::string_view relativePath);
```

The whole 55-line function collapses to `paths::assetPath("completions/noctalia.bash")`.

This matters beyond dedup — the two implementations *disagree* on `NOCTALIA_ASSETS_DIR`:

| | `resource_paths.cpp` | PR |
|---|---|---|
| validates it's a real bundle | yes (`isAssetRoot`) | no |
| warns on bad value | `kLog.warn` | silent skip → next fallback |
| exe-relative candidates | 4 | 1 |

Same env var, two behaviors, one of which silently papers over misconfiguration. That is the "no fallbacks / no silent default chains" rule verbatim, and `CONTRIBUTING.md#runtime-assets` documents the *other* order as the contract.

(`src/theme/firefox_theme/css.cpp` is precedent, not justification — it predates nothing that makes a third copy OK.)

### 3. Command lists are already out of sync with `main`

Diffed the fish list against the live IPC registry on `feaed19d9`:

```
in binary, missing from completions:
  clipboard-copy
  clipboard-text
  osd-disable
  osd-enable
  osd-toggle
```

The same 5 are missing from `noctalia.bash` and `_noctalia`. Also `Nord` is missing from `_NOCTALIA_BUILTIN_PALETTES` in all three — `src/theme/builtin_palettes.cpp` defines 10 palettes, the scripts list 9.

Six items stale on day zero, across three files that must be edited in lockstep. This needs a mitigation, not just a resync. Options, in order of preference:

- **Complete `msg` subcommands dynamically.** The scripts already shell out to a running instance for panels and plugins; parsing `noctalia msg --help` (which prints the registry with descriptions) is the same class of call and kills all three hardcoded lists. Degrade to nothing when no instance is running — that is runtime resilience, not a config fallback.
- Failing that, a CI check that diffs the three lists against each other at minimum, so they cannot drift *from one another* too.

### 4. `source <(noctalia completions bash)` twice is broken

```
$ bash -c 'source ./noctalia.bash; source ./noctalia.bash'
./noctalia.bash: line 1: _NOCTALIA_TOP_COMMANDS: readonly variable
./noctalia.bash: line 14: _NOCTALIA_MSG_COMMANDS: readonly variable
./noctalia.bash: line 124: _NOCTALIA_BOOL_STATES: readonly variable
./noctalia.bash: line 126: _NOCTALIA_WALLPAPER_SCHEMES: readonly variable
./noctalia.bash: line 138: _NOCTALIA_THEME_OPTS: readonly variable
./noctalia.bash: line 153: _NOCTALIA_BUILTIN_PALETTES: readonly variable
```

Re-sourcing `.bashrc` is routine. Drop `readonly`, or move the data inside the completion function.

### 5. `Rosé Pine` never completes in bash

`_NOCTALIA_BUILTIN_PALETTES` is a real array, but it is flattened with `${arr[*]}` into `compgen -W` and re-split on IFS:

```
$ arr=(Ayu "Rosé Pine" Tokyo-Night); COMPREPLY=( $(compgen -W "${arr[*]}" -- "") )
[Ayu] [Rosé] [Pine] [Tokyo-Night]
```

Fish has the same problem, handled by embedding literal quotes — `"'Rosé Pine'"` — which completes the quote characters into the command line. Both are wrong.

---

## Should fix

### 6. `<iostream>` / `std::cout` is the first in `src/`

`grep -rn iostream src/` returns zero matches; the codebase is uniformly `std::print` / `std::println` / `std::fputs`. Use `std::fwrite` from a read buffer, or read to string + `std::fputs`. Also avoids mixing an unsynced `std::cout` with the `std::println(stderr, …)` calls in the same function.

### 7. Help goes to stdout on the error path

```cpp
if (argc < 3 || std::strcmp(argv[2], "--help") == 0) {
  std::println("{}", kHelpText);
  return argc < 3 ? 1 : 0;
}
```

Missing-arg is an error: message belongs on stderr, matching `ipc::runCli`'s `std::fputs("error: msg requires a command …", stderr)`. Also `kHelpText` already ends in `\n`, so `println` adds a stray blank line.

### 8. `selfExePath()` does not detect truncation

`n == sizeof(buf) - 1` is accepted as a valid path. `paths::executablePath()` rejects it. Moot if you adopt #2.

### 9. Meson installs to the wrong place for shells

`install_subdir('assets')` puts them under `share/noctalia/assets/completions/`, where no shell looks. Three lines make it work out of the box for every distro package:

```meson
install_data('assets/completions/noctalia.bash',
  install_dir: get_option('datadir') / 'bash-completion' / 'completions', rename: 'noctalia')
install_data('assets/completions/_noctalia',
  install_dir: get_option('datadir') / 'zsh' / 'site-functions')
install_data('assets/completions/noctalia.fish',
  install_dir: get_option('datadir') / 'fish' / 'vendor_completions.d')
```

`PACKAGING.md` "Installed layout" needs the corresponding entries.

---

## Notes / non-blocking

- **Panel completion scrapes an error string.** `noctalia msg panel-toggle '['` → `error: unknown panel "[" (available: clipboard, control-center, …)`. It works today, but that message text is not a contract, and the completion fires a real IPC call with a deliberately bogus argument. A `panel-list` IPC command — or adding the id list to `msg status`, which currently exposes only `activePanelId` / `barVisible` / `locked` / `panelOpen` — would make this honest.
- **`dmenu` and `firefox-theme` are absent** from top-level completions. Both are real dispatches in `main.cpp:294-303`. They are also absent from `noctalia --help`, so this may be deliberate — confirm intent either way.
- **zsh: `plugin_list` is not declared `local`** in `_noctalia_get_plugin_sources`, `_noctalia_get_enabled_plugins`, and `_noctalia_get_disabled_plugins` — leaks a global into the user's shell.
- Helper functions call bare `noctalia` from `PATH` rather than the command being completed. Minor, but it breaks completion for a non-`PATH` build like the one in the PR's own testing section.

---

## Verified accurate

- `config` subcommands and flags match `noctalia config --help` exactly.
- `theme` options and the 10 `--scheme` values match `noctalia theme --help` exactly.
- `install_subdir('assets')` does pick up the new directory, so the files do ship.
- Fish script parses clean: `fish -c 'source ./noctalia.fish'` → OK.

oldirtty:
Pushed an update addressing the review feedback, dropped runtime filesystem lookups in favor of build-time embedding via native meson. PR description fixed as well.

TotalyEnglizLitrate:
This got me thinking, and while this is fine for now, a future update could introduce a shared declarative CLI schema (possibly using a third party lib) and have `noctalia completions <shell>` generate completions at runtime.

The substantial part of this migration would be consolidating the currently independent CLI logic into a single source of truth covering commands, subcommands, options, aliases, positional arguments, defaults, and validation.
This could then be used to generate help, usage text and completions.

Grouping things into a single schema would also eliminate the current fragmentation, where each CLI area (main.cpp, theme/cli.cpp, config/cli.cpp, ipc/cli.cpp, dmenu_cli.cpp, plugin_lint.cpp, completions/cli.cpp, and assets/completions/*) maintains its own parsing, validation, defaults, and help/completion logic independently.

This would also make richer Zsh and Fish completions (with per-option descriptions like those shown in the screenshot) practical to maintain.

image

I would very much like to tackle this, but it's likely a large enough change that I might not have the time or expertise to handle it by myself. I'll happily help anyone willing to put in the legwork, in whatever way I can.

assets/completions/noctalia.fish
```diff
@@ -0,0 +1,234 @@
...
+end
+
+# `noctalia <TAB>` completions
+complete -c noctalia -n "not __fish_seen_subcommand_from $TOP_CMDS" -a "$TOP_CMDS"
+complete -c noctalia -n "not __fish_seen_subcommand_from $TOP_CMDS" -s d -l daemon
+complete -c noctalia -n "not __fish_seen_subcommand_from $TOP_CMDS" -s h -l help
+complete -c noctalia -n "not __fish_seen_subcommand_from $TOP_CMDS" -s v -l version
+
+# `noctalia completions <TAB>` completions
+complete -c noctalia -n "__fish_seen_subcommand_from completions" -a "bash fish zsh"
```
TotalyEnglizLitrate: P.S. just found this while messing around/testing completions, this needs to be changed

```suggestion
complete -c noctalia -n "__fish_seen_subcommand_from completions; and not __fish_seen_subcommand_from bash fish zsh" -a "bash fish zsh"
```

currently it just keeps giving completions instead of stopping after a single shell has been provided as an argument

oldirtty:
> The substantial part of this migration would be consolidating the currently independent CLI logic into a single source of truth covering commands, subcommands, options, aliases, positional arguments, defaults, and validation. This could then be used to generate help, usage text and completions.
> 
> Grouping things into a single schema would also eliminate the current fragmentation, where each CLI area (main.cpp, theme/cli.cpp, config/cli.cpp, ipc/cli.cpp, dmenu_cli.cpp, plugin_lint.cpp, completions/cli.cpp, and assets/completions/*) maintains its own parsing, validation, defaults, and help/completion logic independently.

I agree! Having a single source of truth for the entire CLI  would be the ultimate solution here.

As you mentioned, consolidating all CLI parsing and validation into a declarative schema would be a massive refactoring task. Honestly, considering the current scope, the huge amount of effort required right now wouldn't really justify the immediate return.

That's my first time doing completions, also don't have the expertise, but if the project decides to take on that larger CLI migration in the future, I'll be more than happy to jump in and help test or contribute wherever I can.

oldirtty:
Hey @TotalyEnglizLitrate, I've formalized this proposal in an issue: #3765. Feel free to check it out!

oldirtty:
I'm closing this PR because the discussion in #3765 is moving towards a more dynamic, agnostic solution rather than a hardcoded one. We'll build the completions on top of a new centralized CLI schema, as we think it's a better approach.

PR #3754, opened 2026-08-02
feat(wayland): include monitor serial number in output description

### LINKED ISSUE #3670: [FEATURE] bar monitor overrides should allow serial number filter
### Submission checklist

### Feature type

Configuration / customization

### Feature summary

The bar monitor override can only match by model or connector, but in the case I have multiple monitors of the same model or I connect the same monitor in different ports.

If we could decide to match the port, model and/or serial number it would give more flexibility.

I'm running niri and the niri msg outputs does show the serial number as part of description and noctalia doesn't when it identifies the monitor.

### Motivation / use case

There are cases where we're restricted by only matching by model or connector. If we could add serial number we can identify one monitor uniquely.

### Proposed solution

Add a tag to identify monitors instead of just relying on the name or adding the serial number to the string that we match on.

### Alternatives considered

_No response_

_No response_

### Additional context

_No response_

rogervn:
## Summary

We could improve the fields the user can match whilst overriding a bar per monitor on the monitor definition by serial number as well as the make and model that the current wl_output protocol descritption shows.

We introduce the zwlr_output_manager_v1 extension to also get the serial number alongside the make and model and add that to a full description string we can match substrings from. (We also fix a small bug that the connection was doubled in the string).

It will fall back to the original wl_output description if the compositor doesn't support the extension.

## Motivation

If the user has multiple monitors of the same make and brand and wants to setup both with different setups (e.g. one vertical and one horizontal) they might need monitor overrides that we can't do with the current string. This change makes it possible for the user to customise up to the serial number when the compositor allows.

## Type of Change

- [ ] Bug fix
- [X] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Build / packaging

## Related Issue

Closes https://github.com/noctalia-dev/noctalia/issues/3670

## Testing

Ran locally, checked the updated string on the logs containing the serial number:

```
09:58:10.256 [INF] [bar] creating #0 "main" on DP-2 (LG Electronics - LG TV SSCR2 - 0x01010101), thickness=34 position=top reserve_space=true exclusive_zone=34
```

Then created override monitor configs for the bar using:
1. Connector name (to verify we didn't break the original behaviour with the bug fix)
2. Serial number

Both override changed the intended monitor.

## Manual Coverage

- [X] Tested on Niri
- [ ] Tested on Hyprland
- [ ] Tested on Sway

## Checklist

## Additional Notes

I understand that the wayland extension might be undesirable. It's the best option I could find but I'm happy redesigning this with different bases if anyone would see a better way forward here. I'm not opinionated at all on this change.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -370,6 +370,8 @@ foreach _p : [
     'protocols/wlr-gamma-control-unstable-v1.xml'],
   ['wlr-screencopy-unstable-v1',
     'protocols/wlr-screencopy-unstable-v1.xml'],
+  ['wlr-output-management-unstable-v1',
+    'protocols/wlr-output-management-unstable-v1.xml'],
 ]
   _name = _p[0]
   _xml  = _p[1]
@@ -1127,6 +1129,7 @@ if build_tests
     'kde_color_scheme',
     'log',
     'math_provider',
+    'monitor_selector',
     'notification_filter',
     'notification_history_utf8',
     'path_browse',
--- a/protocols/wlr-output-management-unstable-v1.xml
+++ b/protocols/wlr-output-management-unstable-v1.xml
@@ -1,0 +1,601 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<protocol name="wlr_output_management_unstable_v1">
+  <copyright>
+    Copyright © 2019 Purism SPC
+
+    Permission to use, copy, modify, distribute, and sell this
+    software and its documentation for any purpose is hereby granted
+    without fee, provided that the above copyright notice appear in
+    all copies and that both that copyright notice and this permission
+    notice appear in supporting documentation, and that the name of
+    the copyright holders not be used in advertising or publicity
+    pertaining to distribution of the software without specific,
+    written prior permission.  The copyright holders make no
+    representations about the suitability of this software for any
+    purpose.  It is provided "as is" without express or implied
+    warranty.
+
+    THE COPYRIGHT HOLDERS DISCLAIM ALL WARRANTIES WITH REGARD TO THIS
+    SOFTWARE, INCLUDING ALL IMPLIED WARRANTIES OF MERCHANTABILITY AND
+    FITNESS, IN NO EVENT SHALL THE COPYRIGHT HOLDERS BE LIABLE FOR ANY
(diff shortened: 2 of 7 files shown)


ItsLemmy:
Thanks for the PR. I'm fine with the new protocol addition, as all modern compositors seems to support it.  But there are a couple issues to fix.

 ### [P1] Preserve the compositor description when adding serial matching                                                                                  
                                                                                                                                                           
 src/wayland/wayland_connection.h:107-110                                                                                                                  
                                                                                                                                                           
 displayDescription() returns the original wl_output.description only when all output-management fields are empty. Once any of make, model, or serialNumber is present, the original description is discarded. outputMatchesSelector() now searches only this replacement string.                        
                                                                                                                                                           
 This breaks existing description-based selectors on compositors supporting output management. It also mishandles valid partial metadata: the protocol explicitly allows make, model, and serial events to be omitted independently. For example:                                                                
                                                                                                                                                           
 ```text                                                                                                                                                   
description = "Acme UltraDisplay DP-2"
make = "Acme"
model = ""
serialNumber = ""
 ```                                                                                                                                                       
                                                                                                                                                           
 A selector for UltraDisplay previously matched and now fails because the effective description becomes only Acme.                                         
                                                                                                                                                           
 Keep the compositor description as the display value and match serial/make/model as additional fields, or append the available identity fields without  removing any existing description tokens. Add targeted coverage for unsupported, partial, and complete output-management metadata.                        
                                                                                                                                                           
 ### [P2] Do not retain out across the re-entrant output-change callback                                                                                   
                                                                                                                                                           
 src/wayland/wayland_connection.cpp:193-195                                                                                                                
                                                                                                                                                           
 matchPendingOutputHeads() can call m_outputChangeCallback(). The callback enters shell reconciliation and may re-enter Wayland dispatch; m_outputs can  then be mutated and invalidate the out pointer. outputName() subsequently dereferences out->done.                                                         
                                                                                                                                                           
 Capture readiness before invoking the callback, or reacquire the output through findOutputByWl(wlOut) afterward. The surrounding code already follows the invariant that vector-backed output references must not survive callbacks.

rogervn:
No worries, it makes sense. I went with matching them individually and added the tests for coverage. LMK if there's anything else I could do here!

Thanks for the review!

ItsLemmy:
Thanks a lot.

PR #3763, opened 2026-08-02
feat(launcher): Hide/Unhide and multi-backend Uninstall in the app context menu

PriyanshusSGupta:
## Summary

Adds a right-click context menu for apps in the launcher's Browse Apps view with three actions:

1. **Hide / Unhide** — toggles the freedesktop `NoDisplay=` key on the `.desktop` entry. System entries (`/usr/share/applications`) get a user-level override copy in `~/.local/share/applications` (full copy, atomic write), preserving `Exec`/`Icon`. Works on every distro — pure freedesktop standard.

2. **Uninstall** — runtime-detected, per-backend removal, Arch-focused but distro-safe:
   - **Standard/AUR**: `shelly remove standard|aur <pkg>` with `pacman -Rns` fallback
   - **Flatpak**: `shelly remove flatpak <id>` with `flatpak uninstall -y --app <id>` fallback
   - **AppImage**: `shelly remove appimage <path>` when shelly-managed
   - **Custom/unowned** (Steam shortcuts, PWAs, hand-made launchers): **Remove launcher** — plain `.desktop` unlink, no escalation
   
   Uninstall runs **interactively in a terminal** (via `terminal_launch::prepareCommand` + `resolvePrivilegeEscalator`) so pacman/shelly prompts — dependency-cascade confirmations, "breaks packages" warnings, polkit — are seen and answered by the user, never silently forced with `--noconfirm`.

3. **Post-action verification + list refresh** — backend-specific checks (`pacman -Q`, `flatpak info`, file existence) drive the toast, and the launcher re-gathers results unconditionally.

## Why this design

- No hard dependencies: pacman/shelly/flatpak presence is detected at runtime. On non-Arch distros the menu degrades gracefully to Hide + Remove launcher only.
- `resolveAppSource()` classifies each entry (X-Flatpak key, X-AppImage keys/Exec, pacman `-Qo`/`-Qm`) — see `src/system/desktop_entry_override.{h,cpp}`.
- The interactive-terminal pattern fixes real failures where non-interactive `pkexec pacman -Rns` stalls on cascade/break warnings.

## Files changed

- `src/system/desktop_entry_override.{h,cpp}` — new: `setDesktopEntryHidden()`, `resolveAppSource()`, `resolveOwningPackage()`, `shellyListsAppImage()`
- `src/shell/launcher/launcher_panel.{h,cpp}` — context menu entries, per-backend dispatch, confirm dialog, `runUninstall()`/`runRemoveLauncherFile()`
- `assets/translations/en.json` — new `launcher.context-menu.*` keys (en only, per contributing guide)
- `meson.build` — register new source file

## Checks

- `clang-format` clean, `clang-tidy` clean, `i18n-check.py` OK
- `meson test`: 66/66 pass
- Manually tested: Hide/Unhide, standard uninstall (avahi break-warning case), cascade confirmation, Flatpak backend, AppImage (shelly-managed), custom "Remove launcher"

### DIFF
--- a/assets/translations/en.json
+++ b/assets/translations/en.json
@@ -644,7 +644,15 @@
     "context-menu": {
       "open": "Open",
       "pin-to-dock": "Pin to dock",
-      "unpin-from-dock": "Unpin from dock"
+      "unpin-from-dock": "Unpin from dock",
+      "hide": "Hide",
+      "unhide": "Unhide",
+      "uninstall": "Uninstall",
+      "uninstall-cancel": "Cancel",
+      "uninstall-confirm": "Uninstall",
+      "uninstall-done": "Package uninstalled:",
+      "uninstall-failed": "Uninstall failed",
+      "hide-failed": "Could not hide this app"
     },
     "empty": {
       "no-results": "No results found",
--- a/assets/translations/en.json
+++ b/assets/translations/en.json
@@ -652,6 +652,14 @@
       "uninstall-confirm": "Uninstall",
       "uninstall-done": "Package uninstalled:",
       "uninstall-failed": "Uninstall failed",
+      "uninstall-aur-suffix": " (AUR)",
+      "uninstall-flatpak-suffix": " (Flatpak)",
+      "uninstall-appimage-suffix": " (AppImage)",
+      "remove-launcher": "Remove launcher",
+      "remove-launcher-confirm": "Remove launcher for",
+      "remove-launcher-done": "Launcher removed",
+      "remove-launcher-failed": "Failed to remove launcher",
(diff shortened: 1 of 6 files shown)


ItsLemmy:
Thanks for the PR, but we won't be merging this.
- This approach is too Arch base  and a lot of us don't use Arch.
- There is currently not a single use of "pacman" "flatpak" or any package manager in noctalia core.
- The "degrades gracefully" mention on non Arch distros is not really a thing, since it will just hide entries that you can never bring back.
- The launcher is not an app store and it's not its responsibility to uninstall apps.

This could be a cool plugin if you feel like working on this.

PR #3840, opened 2026-08-08
fix(cc/network): stop wifi toggle from snapping back before state catches up

### LINKED ISSUE #3837: [FEATURE]: Smooth enable/disable wifi
### Submission checklist

### Feature type

UI / visual improvement

### Feature summary

When I toggle the wifi on/off it force refresh (look rec)

### Motivation / use case

Not good visualy

### Proposed solution

Wait 1sec before force update the state

### Alternatives considered

_No response_

https://discord.com/channels/1401598189823590460/1533573713222565888

### Additional context

https://github.com/user-attachments/assets/b8059117-2e02-4057-b8c6-b86cd579eb6a

TotalyEnglizLitrate:
## Summary

Add a wifi-specific pending state (target/since/ flag) that holds the toggle at the requested value until state confirms the flip or a timeout elapses, reusing same pattern as the existing connect/disconnect m_actionPending guard, but tracked separately.

## Motivation

Previously, the toggle was synced to s.wirelessEnabled unconditionally every frame, so a user click could get reverted/flicker while the backend was still applying the change.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Build / packaging

## Related Issue
Closes #3837 

## Testing
Verify no jitter when toggling wifi

## Manual Coverage
- [x] Tested on Niri
- [ ] Tested on Hyprland
- [ ] Tested on Sway

## Screenshots / Videos

https://github.com/user-attachments/assets/09011c1c-4456-43bb-af9a-459bc0dcb291

## Checklist

- [x] This PR is ready for review.
- [x] This PR does not change user-facing configuration or behavior.

## Additional Notes

N/A

### DIFF
--- a/src/dbus/network/inetwork_service.h
+++ b/src/dbus/network/inetwork_service.h
@@ -2,7 +2,6 @@
 
 #include "dbus/network/network_types.h"
 
-#include <cstdint>
 #include <functional>
 #include <string>
 #include <vector>
@@ -15,6 +14,7 @@ class INetworkService {
 public:
   using ChangeCallback = std::function<void(const NetworkState&, NetworkChangeOrigin)>;
   using WirelessFeedbackCallback = std::function<void(bool enabled)>;
+  using WirelessEnabledCompletion = std::function<void(bool success)>;
 
   virtual ~INetworkService() = default;
 
@@ -33,7 +33,7 @@ public:
   virtual bool deactivateVpnConnection(const VpnConnectionInfo& vpn) = 0;
   [[nodiscard]] virtual bool canActivateWiredConnection() const noexcept { return false; }
   virtual bool activateWiredConnection() { return false; }
-  virtual void setWirelessEnabled(bool enabled) = 0;
+  virtual void setWirelessEnabled(bool enabled, WirelessEnabledCompletion onComplete = {}) = 0;
   virtual void disconnect() = 0;
   virtual void forgetSsid(const std::string& ssid) = 0;
   [[nodiscard]] virtual bool hasSavedConnection(const std::string& ssid) const = 0;
--- a/src/dbus/network/iwd_service.cpp
+++ b/src/dbus/network/iwd_service.cpp
@@ -340,9 +340,10 @@ bool IwdService::activateAccessPoint(const AccessPointInfo& ap, const std::strin
   return activateAccessPoint(ap);
 }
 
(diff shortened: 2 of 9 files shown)


ItsLemmy:
- [P1] src/shell/control_center/tabs/network_tab.cpp:740-746 — Rapid reversal treats stale state as confirmation.                                          
   Start with Wi-Fi off, click on, then click off before NetworkManager reports the first write. The second target is false, which already equals the stale s.wirelessEnabled, so m_wifiTogglePending clears on the next update. When the first asynchronous write subsequently reports true, the toggle snaps on before the second write restores false—the jitter this PR intends to prevent. Serialize requests, disable the toggle while pending, or require confirmation associated with the latest request rather than target equality alone.                                                                     
     
 - [P1] src/shell/control_center/tabs/network_tab.cpp:741 — The six-second timeout has no wake-up mechanism.                                                
   syncCurrentCard()evaluates the deadline only during update/layout. NetworkTab does not request frame ticks or schedule a timer, and the toggle animation only keeps frames running briefly. In particular, NetworkManagerService’s asynchronous write-error callback clears its internal pending value and logs the error without invalidating the panel. After a rejected or dropped request, the pending target can therefore remain displayed indefinitely until an unrelated network/UI update occurs. Schedule an update at the deadline or keep requesting frame ticks while the operation is pending.                 
   
 - [P2] src/shell/control_center/tabs/network_tab.h:93-95 — Pending toggle state survives panel closure.                                                    
   NetworkTab::onClose()clears m_actionPending but not the newly added m_wifiTogglePending. If the panel closes while a request is unresolved and reopens  before the deadline, the newly created toggle can be overwritten with the stale target even if the network changed externally while the panel was closed. Clear the Wi-Fi pending state in onClose() alongside the existing action state.

TotalyEnglizLitrate:
Ran into the spinner bug while testing the wifi toggle fix, intial commit (32d02e41b3d5) was almost entirely claude because I couldn't figure out the nm stuff, tested yesterday, went over it now to clean up the code, and made sure it actually did what it said.

ItsLemmy:
Thanks for the PR, made it even more robust.   Unfortunately it's hard to test 100% for me because my adapter toggles instantly, hope @Cleboost can check if that really fixes his issue.

PR #3863, opened 2026-08-10
feat(audio): add Bluetooth codec selector drop-down

### LINKED ISSUE #3634: [FEATURE] Select audio codec
### Submission checklist

### Feature type

New functionality

### Feature summary

Selectable audio codec in the sound options panel.

### Motivation / use case

I am currently using wiremix (TUI) to switch codec. Sometimes Bluetooth headphones like to switch to a very low quality phone call codec and have to be manually switched back. Some people may also like to switch between quality and latency optimised codecs depending on current usage.

### Proposed solution

In the sound options panel, under 'output', have a drop down menu with available codecs for the current default output device.

### Alternatives considered

Using a TUI to switch codecs. It works but is not very pretty.

There was a [v4 pull request](https://github.com/noctalia-dev/noctalia/pull/2396) for this feature, but currently none for v5.

### Additional context

_No response_

GhostEagle68:
## Summary

Adds a **Codec** drop-down menu to the Output card in the Audio control-center tab, letting users switch the codec of a connected Bluetooth audio device (SBC / SBC-XQ / AAC / LDAC / aptX / HFP) without leaving the shell.

`PipeWireService` already tracks PipeWire `Device` objects for route/port handling, so this extends that to subscribe to `SPA_PARAM_EnumProfile` / `SPA_PARAM_Profile`. Only for devices reporting `device.api = "bluez5"`, so non-Bluetooth cards see no additional registry traffic. Applying a selection builds a `SPA_TYPE_OBJECT_ParamProfile` pod with `save = true`, the same mechanism `wpctl set-profile` uses, so WirePlumber persists the choice across reconnects with no extra state on our side.

The row hides itself unless the active output is a Bluetooth device exposing two or more codec choices, matching the existing behavior of the Effects profile row directly above it.

## Motivation

Bluetooth headsets frequently drop to a low-quality handsfree codec commonly after handling a call. As of now there is currently no in-shell way to switch back. Users have to reach for `wiremix`, `wpctl`, or `pavucontrol`. Some also want to trade quality against latency depending on what they are doing.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Build / packaging

## Related Issue

Closes #3634

## Testing

```sh
just build          # clean, no new warnings
just test           # 75/75 passing
just format         # applied
just lint           # clang-tidy clean on changed files
python3 tools/i18n-check.py   # OK, 874 keys
```

Manual, on a Galaxy Buds3 Pro (multi-codec: SBC / SBC-XQ / AAC, plus HFP variants):

- Codec row appears once the device is the active output, listing `SBC`, `SBC-XQ`, `AAC`, `HFP`
- Selecting a codec applies immediately; the active profile is reflected with a checkmark
- Row stays hidden for a USB wireless headset (Logitech G733; not `bluez5`)
- Renders correctly alongside the EasyEffects profile row when both are present

## Manual Coverage

- [ ] Tested on Niri
- [x] Tested on Hyprland
- [ ] Tested on Sway

## Screenshots / Videos

with effects

## Checklist

## Additional Notes

**Why the labels are derived rather than taken verbatim.** PipeWire reports these profiles for the test device:

| name | description | priority |
|---|---|---|
| `off` | Off | 0 |
| `a2dp-sink-sbc` | High Fidelity Playback (A2DP Sink, codec SBC) | 133 |
| `a2dp-sink-sbc_xq` | High Fidelity Playback (A2DP Sink, codec SBC-XQ) | 132 |
| `a2dp-sink` | High Fidelity Playback (A2DP Sink, codec AAC) | 134 |
| `headset-head-unit-cvsd` | Headset Head Unit (HSP/HFP, codec CVSD) | 5 |
| `headset-head-unit-msbc` | Headset Head Unit (HSP/HFP, codec MSBC) | 6 |
| `headset-head-unit` | Headset Head Unit (HSP/HFP, codec LC3-SWB) | 7 |

Three things fall out of this:

1. Seven profiles are only four meaningful choices and the three headset variants are one decision to a user ("the mode with a microphone"), so they are collapsed.
2. Matching codec names by substring is unsafe: `headset-head-unit-msbc` contains `sbc` and would be mislabeled as the A2DP SBC codec. Each profile is therefore reduced to a codec *key* first, with headset profiles classified before any codec parsing.
3. The unsuffixed names (`a2dp-sink`, `headset-head-unit`) carry the device's *best* codec and the highest priority in their group, so where several profiles share a key the highest-priority one wins. Picking the first would have selected CVSD, the worst option available.

The active selection is matched by codec key rather than profile index, so if PipeWire reports the active profile as a variant that was folded into another entry, the surviving entry still shows as selected.

Labels are codec and protocol acronyms only (`SBC`, `AAC`, `HFP`, …), so nothing here needs translating and no i18n plumbing was added to `src/pipewire`.

**Known limitation.** If a user manually selects a virtual sink (for example "Easy Effects Sink") as their output device, the Codec row hides, since that sink has no device behind it. This is not the default EasyEffects behavior as it intercepts the application streams rather than taking over the default sink. But resolving the real downstream device would need link-graph walking, which seemed disproportionate here. Happy to add it if wanted.

### DIFF
--- a/assets/translations/en.json
+++ b/assets/translations/en.json
@@ -183,6 +183,8 @@
   "control-center": {
     "audio": {
       "application-volumes": "Application Volumes",
+      "choose-codec": "Choose a codec",
+      "codec": "Codec",
       "choose-effects-profile": "Choose a profile",
       "effects-profile": "Effects",
       "input-device-prefix": "Input",
--- a/assets/translations/en.json
+++ b/assets/translations/en.json
@@ -186,7 +186,7 @@
       "choose-codec": "Choose a codec",
       "choose-effects-profile": "Choose a profile",
       "codec": "Codec",
-      "codec-tooltip": "Codec used for this Bluetooth device. Higher-quality codecs may use more bandwidth and battery.",
+      "codec-tooltip": "Select codec to use for this Bluetooth device. Higher-quality codecs may use more bandwidth and battery.",
       "effects-profile": "Effects",
       "input-device-prefix": "Input",
       "no-application-audio": "No application audio",
--- a/assets/translations/en.json
+++ b/assets/translations/en.json
@@ -184,8 +184,9 @@
     "audio": {
       "application-volumes": "Application Volumes",
       "choose-codec": "Choose a codec",
-      "codec": "Codec",
       "choose-effects-profile": "Choose a profile",
+      "codec": "Codec",
+      "codec-tooltip": "Codec used for this Bluetooth device. Higher-quality codecs may use more bandwidth and battery.",
       "effects-profile": "Effects",
       "input-device-prefix": "Input",
       "no-application-audio": "No application audio",
--- a/assets/translations/en.json
+++ b/assets/translations/en.json
@@ -186,7 +186,7 @@
       "choose-codec": "Choose a codec",
       "choose-effects-profile": "Choose a profile",
       "codec": "Codec",
-      "codec-tooltip": "Select codec to use for this Bluetooth device. Higher-quality codecs may use more bandwidth and battery.",
+      "codec-tooltip": "Select codec to use for this device. Please note that higher-quality codecs can use more bandwidth and battery.",
       "effects-profile": "Effects",
       "input-device-prefix": "Input",
       "no-application-audio": "No application audio",
--- a/src/pipewire/pipewire_service.cpp
+++ b/src/pipewire/pipewire_service.cpp
@@ -2023,6 +2023,26 @@ void PipeWireService::setSourceVolume(std::uint32_t id, float volume) {
 }
 void PipeWireService::setSourceMuted(std::uint32_t id, bool muted) { setNodeMuted(id, muted); }
 void PipeWireService::setDefaultSource(std::uint32_t id) { setDefaultNode(id, "default.audio.source"); }
+void PipeWireService::setDeviceProfile(std::uint32_t deviceId, std::int32_t profileIndex) {
+  const auto it = m_devices.find(deviceId);
+  if (it == m_devices.end() || it->second.proxy == nullptr) {
+    return;
+  }
+
+  std::uint8_t buffer[256];
+  spa_pod_builder builder;
+  spa_pod_builder_init(&builder, buffer, sizeof(buffer));
+
+  spa_pod_frame frame;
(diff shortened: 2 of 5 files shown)


ItsLemmy:
### [P2] Do not parse localized PipeWire descriptions for codec identity                                                                                         

pipewire_service.cpp:569-580 

The preferred A2DP profile derives its codec from the literal English marker "codec ". PipeWire localizes this description. For example, its German catalog      
translates it to:                                                                                                                                                

```text                                                                                                                                                          
High Fidelity-Wiedergabe (A2DP-Senke, Codec %s)
```                                                                                                                                                              

The case-sensitive search fails, producing {key = "a2dp", label = "A2DP"} instead of the actual codec. Thus non-English users see a generic A2DP option for the  
preferred codec.                                                                                                                                                 

Use non-localized protocol data—such as the profile index/codec enum or another stable field—rather than parsing SPA_PARAM_PROFILE_description.                  

### [P2] Replace profile snapshots instead of only upserting them                                                                                                

pipewire_service.cpp:1555-1561        

Every SPA_PARAM_EnumProfile refresh only inserts or overwrites returned indexes. It never removes profiles omitted by a later enumeration. PipeWire signals      
profile-set changes through the parameter serial and then enumerates only currently valid profiles; it does not provide tombstones for removed entries.          

When Bluetooth capabilities change while the Device global survives—reconnect, endpoint loss, role change—removed codecs remain in DeviceData::profiles. The     
dropdown can then offer an unsupported codec and send its stale profile index.                                                                                   

Stage each enumeration as a snapshot and replace profiles when complete, or otherwise clear it at the start of a coordinated re-enumeration.                     

### [P2] Include PipeWire BAP profiles or explicitly limit the feature to classic Bluetooth                                                                      

pipewire_service.cpp:558-583                                                                                                                                              

The classifier accepts only a2dp-sink* and headset/HFP names. PipeWire 1.6.8 also exposes Bluetooth LE Audio profiles as bap-sink, bap-duplex, and codec-suffixed variants. These devices still report device.api = "bluez5", but every BAP profile receives an empty key, so the Codec row remains hidden—typically including LC3-capable LE Audio devices.                                                                                                         

Handle the BAP profile names and output directions, or narrow the PR’s stated Bluetooth support to A2DP/HFP.

ItsLemmy:
3 weeks without activity, closing.

PR #3874, opened 2026-08-11
fix(plugin-store): keep scroll position across sheet body rebuilds

### LINKED ISSUE #3872: [BUG] Plugin Store periodically resets scroll position
### Submission checklist

### Bug description

The plugin browser and individual plugin pages frequently reset their scroll position to the top of the page.

### Steps to reproduce

1. Open the plugin browser interface
2. Scroll down the page (any amount is fine)
3. Wait up to 30 seconds
4. Observe that the page scrolls back up to the top of the page.

1. Open the plugin browser interface
2. Click on any plugin with a sufficiently long description
3. Scroll down
4. wait up to 30 seconds
5. Observe the page scrolling back up to the top.

### Expected behavior

The page stays where it should be

### Actual behavior

The page jumps back to the top of the page

### Logs / error output

```text
There aren't any outputs to the log when the issue occurs
```

### Noctalia version

v5.0.0 (5.0.0_beta.8-1-dirty)

### Compositor

Hyprland

### Distribution

Arch-based

### Environment information

```text
Installation method: AUR
GPU: Intel Integrated Graphics
```

### Additional context

_No response_

TheBinaryLoop:
## Summary

The plugin store's catalog grid and plugin detail page jumped back to the top roughly every 30 seconds. Two things combine to cause it:

1. `SettingsWindow::onExternalOptionsChanged()` runs `requestSceneRebuild()`, which calls `m_editorSheetPopup->rebuildBody()`. That destroys and repopulates the entire sheet body, so any node-local state is gone. `UPowerService` is one of its callers, and `upowerd` polls the battery every 30s — on a laptop the energy/rate values genuinely change, so the open sheet is rebuilt unprompted at that cadence.
2. `ScrollViewState` is the existing mechanism for surviving exactly this, but it did not work for sheet bodies: `ScrollView::doLayout` wrote the clamped offset back into the bound state on every pass, and the first pass after a rebuild measures the content unconstrained (`height() == 0`, so `viewportH == contentH`). That pass reports `maxScrollOffset == 0`, so it zeroed the remembered offset before the real viewport height arrived on the following pass.

So this PR:

- Binds the catalog grid and the detail page to `ScrollViewState` members on `PluginStoreContent` that outlive the rebuilt nodes. `openDetail` resets the detail state so opening a different plugin still starts at the top.
- Replaces the `stashScrollOffset`/`restoreScrollOffset` pair added in 32a3ad31. That approach only covered the three interaction sites that knew a rebuild was coming, so it could not help with rebuilds the store never initiated, and it did not cover the detail view at all.
- Makes `ScrollView` write to bound state only once there is a scrollable extent. This also repairs bound state for the other users of it — `SettingsSheetPopup` binds `m_scrollState` for its scrollable body, so the plugin settings editor sheet had the same latent loss.

`VirtualGridView` never hit the `doLayout` trap because it always hands its `ScrollView` an explicit height, which is why the grid half was salvageable with a stash and the detail half was not.

## Motivation

Fixes the reported behavior: scrolling the store or a plugin's description, then having the view snap back to the top a few seconds later, repeatedly. On a laptop it is unavoidable — the battery poll alone is enough to trigger it.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Build / packaging

## Related Issue

Closes #3872

## Testing

- `meson compile -C build-debug` — clean
- `meson test -C build-debug --print-errorlogs` — 79/79 pass
- `clang-format --dry-run -Werror` on the three changed files — clean
- `run-clang-tidy -p build-debug -warnings-as-errors='*'` on the two changed `.cpp` files — clean

Manual verification: the store sheet can only be opened by clicking and this machine has no pointer/key injection tool, so I drove it with a temporary env-gated harness (removed before commit) that auto-opened the store, fired `onExternalOptionsChanged()` on a 30s repeat as a stand-in for the UPower tick, forced a scroll offset, and sampled `scrollOffset()`/`maxScrollOffset()`.

| | grid | detail |
| --- | --- | --- |
| before | 400 → **0** on every tick | 300 → **0** |
| after | 400 → 400 → 400 | 300 → 300 |

Note for anyone reproducing this way: with the display asleep no layout passes run, so the sampled offsets are pre-layout values and mean nothing. The numbers above are from a run with the display kept awake.

## Manual Coverage

- [ ] Tested on Niri
- [x] Tested on Hyprland
- [ ] Tested on Sway

## Screenshots / Videos

None — the change has no visual output of its own; the observable difference is the scroll offset table above.

## Checklist

## Additional Notes

The `ScrollView::doLayout` change is shared UI code, so it is worth a careful look. The behavioral delta is narrow: when a bound scroll view's content fits its viewport, the remembered offset is no longer overwritten with zero. Nothing renders differently in that state (there is nothing to scroll), and if the content grows again the offset comes back, which is the intent of binding state in the first place. Callers that want a genuine reset already do it explicitly — for example `settings_window_scene.cpp` sets `m_contentScrollState.offset = 0.0f` on section change, and `openDetail` now does the same.

Left deliberately out of scope: `onExternalOptionsChanged` is a blunt "rebuild everything" signal, so a battery tick rebuilds the whole settings scene every 30 seconds on a laptop. Narrowing it would touch every caller and deserves its own PR rather than riding along with a bug fix.

### DIFF
--- a/src/shell/settings/plugin_store_content.cpp
+++ b/src/shell/settings/plugin_store_content.cpp
@@ -1,7 +1,6 @@
 #include "shell/settings/plugin_store_content.h"
 
 #include "config/config_service.h"
-#include "core/deferred_call.h"
 #include "core/input/key_symbols.h"
 #include "core/input/keybind_matcher.h"
 #include "i18n/i18n.h"
@@ -142,25 +141,6 @@ namespace settings {
   }
 
   PluginStoreContent::~PluginStoreContent() = default;
-
-  void PluginStoreContent::stashScrollOffset() {
-    if (m_grid != nullptr && !isDetailView()) {
-      m_pendingRestoreScrollOffset = m_grid->scrollView().scrollOffset();
-    }
-  }
-
-  void PluginStoreContent::restoreScrollOffset() {
-    if (!m_pendingRestoreScrollOffset.has_value()) {
-      return;
-    }
-    const float offset = *m_pendingRestoreScrollOffset;
-    m_pendingRestoreScrollOffset.reset();
-    VirtualGridView* grid = m_grid;
-    if (grid == nullptr) {
-      return;
-    }
-    DeferredCall::callLater([grid, offset]() { grid->scrollView().setScrollOffset(offset); });
-  }
 
   void PluginStoreContent::setOnRebuildNeeded(std::function<void()> cb) { m_onRebuildNeeded = std::move(cb); }
 
@@ -472,7 +452,6 @@ namespace settings {
                 }
 
                 applyFilter();
-                stashScrollOffset();
                 if (m_onRebuildNeeded) {
                   m_onRebuildNeeded();
                 }
@@ -497,7 +476,6 @@ namespace settings {
               .variant = ButtonVariant::Ghost,
               .onClick = [this]() {
                 m_tagFiltersCollapsed = !m_tagFiltersCollapsed;
-                stashScrollOffset();
                 if (m_onRebuildNeeded) {
                   m_onRebuildNeeded();
                 }
@@ -518,7 +496,6 @@ namespace settings {
               .onClick = [this, tag = std::move(tag)]() {
                 m_selectedTag = tag;
                 applyFilter();
-                stashScrollOffset();
                 if (m_onRebuildNeeded) {
                   m_onRebuildNeeded();
                 }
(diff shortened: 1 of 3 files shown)


ItsLemmy:
Is this ready for review?

jinliu:
This works initially, but breaks after clicking into one plugin, then back.

To reproduce:
1. Open plugin store.
2. Click on a plugin.
3. Click "back".
4. Scroll down.
5. Wait for 30s

ItsLemmy:
I'm not sure this is needed anymore, can you guys check if you are still affected?

jinliu:
> I'm not sure this is needed anymore, can you guys check if you are still affected?

Can confirm that it no longer automatically scrolls back.

However, now when clicking into a plugin, then clicking "back" in the top-left corner, it always scrolls to the top. Seems some previous fix to keep the scrolling position was overridden.

PR #3876, opened 2026-08-11
feat(cli): implement dynamic runtime completions for plugins

oldirtty:
## Summary

Added a `runtimeProvider` field to the `Positional` schema. This allows the completion generators (bash, zsh, fish) to inject shell functions that fetch choices dynamically at runtime. 

Implemented this specifically for plugins by adding `plugins_enabled` and `plugins_disabled` providers, hooking them into `msg plugins enable`, `msg plugins disable`, and `msg plugin <entry>`.

## Motivation

Typing out long `author/plugin` names by hand is tedious, and it's hard to remember all the exact names off the top of your head. Initially, I just wanted to implement this for the `msg plugin` command so that hitting `<TAB>` would only suggest currently enabled plugins. However, I realized the other commands could easily make use of this too, like `msg plugins disable` (shows only enabled plugins) and `msg plugins enable` (shows only disabled plugins).

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Build / packaging

## Related Issue

None 

## Testing

- Generated and sourced completions for `bash`, `zsh`, and `fish`.
- Tested `noctalia msg plugins disable <TAB>` and verified it only lists enabled plugins.
- Tested `noctalia msg plugins enable <TAB>` and verified it only lists disabled plugins.
- Tested `noctalia msg plugin <TAB>` to ensure the runtime provider triggers correctly on subcommands.
- ran `just format` and `just test`

## Manual Coverage

- [x] Tested on Niri
- [ ] Tested on Hyprland
- [ ] Tested on Sway

## Screenshots / Videos

image

## Checklist

## Additional Notes

The `runtimeProvider` is can be extensible. If we ever want to add dynamic completions other type of commands in the future, we just need to drop a new provider tag in the schema and add a quick shell function in `completions.cpp`.

### DIFF
--- a/src/cli/completions.cpp
+++ b/src/cli/completions.cpp
@@ -3,6 +3,7 @@
 #include "cli/parse.h"
 #include "cli/schema_root.h"
 
+#include <algorithm>
 #include <print>
 #include <span>
 #include <string_view>
@@ -218,6 +219,11 @@ namespace noctalia::cli {
         "  local words=\"$(noctalia msg plugins list 2>/dev/null | awk '$4 == \"disabled\" {print $1}')\"\n"
         "  COMPREPLY=( $(compgen -W \"$words\" -- \"$cur\") )\n"
         "}\n\n"
+        "_noctalia_plugin_prefix() {\n"
+        "  local words=\"$(noctalia msg plugins list 2>/dev/null | awk '$4 == \"enabled\" {print $1\":\"}')\"\n"
+        "  compopt -o nospace 2>/dev/null\n"
+        "  COMPREPLY=( $(compgen -W \"$words\" -- \"$cur\") )\n"
+        "}\n\n"
         "_noctalia_completions() {\n"
         "  local cur=\"${COMP_WORDS[COMP_CWORD]}\"\n"
         "  local path=noctalia\n"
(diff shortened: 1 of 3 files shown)


ItsLemmy:
1. [P1] msg plugin completes an invalid entry ID — src/cli/schema_msg.h:175-176              
The positional requires author/plugin:entry, but plugins_enabled emits only $1 from       
plugins list, which is author/plugin (src/cli/completions.cpp:214, :361, :503). The IPC   
router performs exact entry-ID matching, so the completed value cannot identify an        
endpoint.
                                    
Reproduced with generated Bash and Fish completions:                                      
```text                                                                                   
noctalia msg plugin ali<TAB>
# becomes:
noctalia msg plugin alice/clock 
```                                                                                       
Both shells append a space, preventing continuation with :entry. Add a provider that      
returns concrete enabled entry IDs. If only plugin-prefix completion is intended, emit    
author/plugin: and suppress the trailing separator in Bash, Zsh, and Fish.

2. [P1] Changed file fails required clang-tidy — src/cli/completions.cpp:421-422             
```cpp                                                                                    
std::replace(safeName.begin(), safeName.end(), ':', '-');
```                                                                                       
This triggers modernize-use-ranges, which is enabled and treated as an error:             
```text                                                                                   
```                                                                                       
Use:                                                                                      
```cpp                                                                                    
std::ranges::replace(safeName, ':', '-');
```                                                                                       
This matches the PR’s failed CI state: build and tests passed; lint failed.

oldirtty:
Thanks for the thorough review! Just pushed the fixes. Let me know if everything looks good now.

image

ItsLemmy:
Thanks!

PR #3900, opened 2026-08-12
feat(plugin-ui): expose input frame visibility

blackbartblues:
## Summary

- Expose the existing native input chrome toggle as the optional declarative `ui.input` property `frameVisible`.
- Keep the current framed appearance as the default; plugin authors opt out with `frameVisible = false`.
- Introduce plugin API 27 and cover frame-free multiline inputs in the reconciler test.

## Motivation

Plugin UIs sometimes place editable text directly on a custom surface, such as a sticky note or document canvas. The native `Input` control already supports hiding its background and border while retaining text editing, selection, cursor, and focus behavior, but the declarative plugin UI could not request it. Plugins therefore had to accept unrelated input chrome or replace a real editor with non-editable content.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Build / packaging

## Related Issue

None.

## Testing

- `meson compile -C build-debug ui_tree_reconciler_test plugin_manifest_test`
- `meson test -C build-debug ui_tree_reconciler plugin_manifest --print-errorlogs`
- `just format` with clang-format 22.1.8
- `git diff --check`

## Manual Coverage

- [ ] Tested on Niri
- [x] Tested on Hyprland
- [ ] Tested on Sway

The locally built shell was launched on Hyprland 0.56.2 with the Clipper plugin. Both the single-line title and multiline body used `frameVisible = false`: their native gray background and border disappeared while the editable text remained visible on the parent sticky-note surface. Omitting the property continues to use the existing framed appearance.

The Clipper panel also exposed an unrelated pre-existing Luau stack-reservation bug in deeply nested declarative UI trees. The manual check used a local-only guard for that crash; the guard is intentionally excluded from this focused PR and does not affect the `frameVisible` implementation or its unit coverage.

## Screenshots / Videos

The visual result was verified with the Clipper sticky-note surface. The reconciler test additionally asserts that the native input background node becomes hidden.

## Checklist

## Additional Notes

`frameVisible` maps directly to `Input::setFrameVisible`. Omitting the property preserves existing plugin behavior.

The branch was rebased onto the latest `main` after `noctalia.getSetting` claimed plugin API 26. `frameVisible` now uses the next API level, 27.

### DIFF
--- a/src/scripting/plugin_api.h
+++ b/src/scripting/plugin_api.h
@@ -26,7 +26,8 @@ namespace scripting {
   inline constexpr std::uint32_t kDirectArgvPluginApiVersion = 24;
   inline constexpr std::uint32_t kWallpaperMaskPluginApiVersion = 25;
   inline constexpr std::uint32_t kGetSettingPluginApiVersion = 26;
-  inline constexpr std::uint32_t kCurrentPluginApiVersion = kGetSettingPluginApiVersion;
+  inline constexpr std::uint32_t kInputFrameVisibilityPluginApiVersion = 27;
+  inline constexpr std::uint32_t kCurrentPluginApiVersion = kInputFrameVisibilityPluginApiVersion;
 
   static_assert(kOldestSupportedPluginApiVersion <= kCurrentPluginApiVersion);
 
--- a/src/ui/ui_tree_reconciler.cpp
+++ b/src/ui/ui_tree_reconciler.cpp
@@ -476,10 +476,11 @@ namespace ui {
       static const std::unordered_set<std::string> kGraph = {"width",   "height",    "flexGrow",   "opacity",
                                                              "visible", "values",    "values2",    "color",
                                                              "color2",  "lineWidth", "fillOpacity"};
-      static const std::unordered_set<std::string> kInput = {"width",    "height",   "flexGrow",    "opacity",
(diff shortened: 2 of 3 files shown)


ItsLemmy:
I dont really see a need for that, aside from creating UI that drift from the core.

blackbartblues:
That concern makes sense, and I agree that plugins should not replace core controls with bespoke imitations. The use case here is the opposite: keeping a real core `Input` embedded in a plugin-owned semantic surface (for example, a sticky note or document canvas), instead of swapping between a label and an input or reimplementing text editing.

`frameVisible` only exposes the existing `Input::setFrameVisible` switch; it does not expose arbitrary input styling. Its default remains `true`, so existing plugin UIs do not change. With it set to `false`, core still owns text editing, caret, selection, focus, scrolling, and input behavior. Only the input fill and border are omitted, allowing the parent surface to serve as the visual frame.

Here is the same Clipper note UI with identical data in both states:

**`frameVisible = true`**

Clipper notes with framed native inputs

**`frameVisible = false`**

Clipper notes with frame-free native inputs

The second case still uses the native core input throughout; it just avoids nesting a gray input surface inside the sticky-note surface. I should have included this concrete before/after example in the PR description initially.

ItsLemmy:
Thanks

PR #3906, opened 2026-08-12
feat(plugins): add native panel context menus

### LINKED ISSUE #3899: [FEATURE] Expose native pointer-anchored context menus to Luau plugin panels
### Submission checklist

### Feature type

Developer / API improvement

### Feature summary

Expose Noctalia's native `ContextMenuPopup` to declarative Luau plugin panels through a pointer-anchored `panel.openContextMenu(...)` API.

A plugin should be able to open a native context menu from an `onRightClick` callback, have the shell anchor it at the exact triggering pointer event, and receive the selected action ID in a Luau callback. The shell should own positioning, styling, Wayland input serials, popup exclusivity, and lifecycle cleanup.

### Motivation / use case

Noctalia already has a native context-menu implementation with XDG popup positioning, edge flipping/sliding, scrolling, keyboard navigation, Escape handling, and consistent shell styling. It is used by native surfaces such as the launcher.

Luau plugins cannot currently use it. Although `ui.button` supports `onRightClick`, that callback is value-less: the pointer coordinates and Wayland serial are discarded before the callback reaches the plugin runtime. A plugin therefore cannot safely open a menu at the triggering cursor position.

This is needed for dense, card-based plugin UIs where secondary actions should not permanently occupy card space. A clipboard manager is one concrete use case: copy, paste, pin, delete, and related actions fit naturally in a context menu. The capability should remain generic and reusable by other plugin panels.

The implementation should be compositor-independent. It should use the existing Wayland/XDG popup path rather than compositor IPC or a global cursor-position query.

#### Public Luau API

Add an imperative, panel-specific API:

```luau
function openCardMenu()
  return panel.openContextMenu({
    items = {
      { id = "copy", label = "Copy" },
      { id = "paste", label = "Paste" },
      { kind = "separator" },
      { kind = "header", label = "Manage entry" },
      { id = "pin", label = "Pin", enabled = true },
      { id = "delete", label = "Delete" },
    },
    onActivate = "onCardMenuAction",
    context = currentCardId,
    maxVisible = 12,
  })
end

function onCardMenuAction(actionId, context)
  -- actionId is the selected item's public string ID.
  -- context is the optional scalar captured when the menu was opened.
end

panel.render(
  ui.button({
    text = "Clipboard entry",
    onRightClick = "openCardMenu",
  })
)
```

Proposed contract:

- `items` is required.
- `onActivate` is the name of a Luau function.
- `context` is optional and may be a string, number, or boolean. It is returned as the callback's second argument so one handler can serve many rendered cards without generating per-item callback functions.
- `maxVisible` is optional. The host retains control of menu sizing and clamps the value.
- `panel.openContextMenu()` returns `true` when the request is accepted.
- It returns `false` when called without a current pointer context or when the panel is no longer active.
- A malformed request table produces a descriptive Luau argument error, consistent with other strict bindings.

Initial item schema:

| `kind` | Required fields | Behavior |
|---|---|---|
| omitted / `"item"` | `id`, `label` | Actionable menu item |
| `"separator"` | none | Non-interactive separator |
| `"header"` | `label` | Non-interactive group heading |

`enabled = false` is supported for normal items.

The first version intentionally does not expose native submenu, swatch, checkbox, radio, arbitrary width/style, or `onDismiss` facilities. These can be designed separately without changing the basic API.

#### Implicit pointer anchor

The plugin should not provide `x`, `y`, or a Wayland input serial. The shell should capture them from the pointer event that invoked `onRightClick`.

Expected internal flow:

```text
Wayland right-button event
  -> InputDispatcher
  -> Button / UiTreeReconciler
  -> one-shot callback pointer context
  -> Luau callback calls panel.openContextMenu()
  -> transient ScriptResult request
  -> PluginPanel
  -> native ContextMenuPopup
```

This avoids compositor-specific IPC, prevents plugins from spoofing off-surface anchors, and preserves the exact triggering position even if the pointer moves while the worker callback is running.

Existing Luau callback signatures should remain unchanged; pointer metadata is host-only state and is not automatically exposed to scripts.

#### Pointer event propagation

1. Preserve the pointer event serial (and event time if useful for future APIs) through `InputDispatcher` and `InputArea::PointerData`.
2. Let `Button` retain pointer data for its right-click callback while keeping the existing void callback API available to native callers.
3. In `UiTreeReconciler`, map control-local coordinates to panel-surface logical coordinates.
4. Add an optional pointer context to `ControlCallback`.
5. Copy it into the `ScriptSnapshot` for only that callback invocation.

#### One-shot runtime request

The open operation must be represented as a transient request, not as persistent declarative UI state:

- Add internal request/item/pointer-context types to the scripting runtime.
- Let `PluginBindingContext` collect at most one context-menu request during a call.
- Return it as a dedicated optional field in `ScriptResult`.
- Explicitly exclude it from `ScriptPatch` and `replayState`.

This is important because replaying a patch after subscription, panel reconstruction, or hot reload must never reopen an old context menu.

Additional runtime rules:

- Pointer context exists only during the direct pointer callback.
- A call may submit at most one menu request.
- Results from a stale runtime generation are rejected.
- If the same result requests `panel.close()`, closing the panel takes precedence.
- Native popup creation occurs only on the main UI thread.

#### Validation and resource limits

Suggested initial limits:

- at most 64 items;
- `maxVisible` clamped to 1–30;
- action ID at most 128 bytes;
- label at most 512 bytes;
- callback name at most 128 bytes;
- no empty or duplicate action IDs;
- at least one actionable item.

Separators, headers, and disabled items must never activate.

Public action IDs remain strings. `PluginPanel` can build a temporary string-to-`int32_t` mapping for `ContextMenuControlEntry`, then translate the native activation back to the public string ID.

#### Native popup integration

`PluginPanel` should own one `ContextMenuPopup` and the mapping/state for its currently open menu.

Open sequence:

1. Close any previous plugin context menu and any open select dropdown.
2. Verify that this plugin panel is still the active panel and has a valid parent surface.
3. Clamp the captured anchor point to the current parent-surface dimensions.
4. Build a small anchor rectangle around the point, following the launcher context-menu pattern.
5. Pass the preserved input serial to `ContextMenuPopupRequest`.
6. Apply the shell's content scale, shadows, automatic width constraints, and XDG positioner behavior.
7. Register it with `PanelManager` only after successful popup creation.

Add an optional input-serial override to `ContextMenuPopupRequest`. Existing native callers that omit it continue to use `WaylandConnection::lastInputSerial()`.

The existing XDG `slide` and `flip` constraint adjustments should keep the popup visible near screen edges.

#### Lifecycle and cleanup

The popup must close on:

- action activation;
- Escape;
- click outside;
- opening another mutually exclusive popup;
- panel close or panel switch;
- plugin hot reload;
- plugin disable/removal;
- parent Wayland surface destruction.

Activation should be guarded by both a `PluginPanel` lifetime token and runtime generation. A stale action must not run after reload or destruction.

`PanelManager::setActivePopup/clearActivePopup` and `beginAttachedPopup/endAttachedPopup` must remain balanced exactly once. A failed popup initialization must not leave a dangling active-popup pointer or attached-popup count.

Select dropdowns and plugin context menus should be mutually exclusive.

#### Plugin API compatibility

This is a new public plugin capability and should increment the current Plugin API version (currently 5 in the inspected tree) to 6 while retaining support for API 3–5.

A plugin using `panel.openContextMenu` declares `plugin_api = 6`. Older cores then reject it with the existing compatibility error instead of running with a missing binding.

If another change consumes API 6 before implementation begins, this proposal should use the next available version.

#### Likely implementation areas

- `src/scripting/plugin_api.h`
- `src/scripting/plugin_bindings.h/.cpp`
- `src/scripting/script_runtime_types.h`
- `src/scripting/script_runtime.cpp`
- `src/render/scene/input_area.h/.cpp`
- `src/render/scene/input_dispatcher.h/.cpp`
- `src/ui/controls/button.h/.cpp`
- `src/ui/ui_tree_reconciler.h/.cpp`
- `src/ui/controls/context_menu_popup.h/.cpp`
- `src/shell/panel/plugin_panel.h/.cpp`
- potentially a small popup-exclusivity helper in `PanelManager`
- tests, Meson registration, README, and plugin API documentation

#### Automated test plan

Add coverage for:

- pointer coordinate and serial propagation without changing existing callback arguments;
- local-control to panel-surface coordinate mapping;
- valid menu request parsing;
- invalid types, bounds, empty IDs, and duplicate IDs;
- returning `false` without pointer context;
- optional scalar `context` preservation;
- public string ID -> native ID -> public string ID mapping;
- inert separator/header/disabled entries;
- clearing the request between callbacks;
- excluding the request from runtime replay state;
- rejecting stale generations after hot reload;
- API 6 compatibility and regression coverage for API 3–5;
- balanced popup registration/attachment cleanup.

Validation should include targeted Meson tests, the full debug test suite, lint, formatting, and a full debug build.

#### Manual acceptance criteria

- The menu opens at the exact right-click location.
- XDG flip/slide keeps it visible at every screen edge.
- Long menus scroll.
- Arrow keys, Home, End, Enter, and Escape work.
- Clicking outside dismisses the menu without closing the panel.
- An action is delivered exactly once with the correct `context`.
- Select dropdowns and context menus are mutually exclusive.
- Closing, switching, reloading, disabling, or removing a plugin with an open menu does not crash or dispatch stale actions.
- Fractional scaling and multi-output placement remain correct.
- Right-click does not arm declarative drag and drop.
- The implementation works through the generic Wayland path on Hyprland, Niri, MangoWC, and other supported compositors without compositor-specific code.

### Alternatives considered

1. **Pass `x` and `y` directly to every `onRightClick` callback.**  
   This changes callback semantics, exposes internal coordinates unnecessarily, and still does not safely carry the correct Wayland serial across the worker boundary.

2. **Let plugins query the global cursor position.**  
   Wayland intentionally does not provide a generic global cursor-position API. Compositor-specific IPC would be fragile and would not be portable.

3. **Render a menu inside the declarative panel tree.**  
   This would not provide a native grabbing popup, XDG edge constraints, keyboard routing, or consistent popup lifecycle, and it would duplicate existing shell behavior.

4. **Store the open request in `ScriptPatch`.**  
   Patches are replayed. This could reopen the menu after subscription, reconstruction, or hot reload, so the request needs a one-shot channel.

5. **Expose all native `ContextMenuControlEntry` fields immediately.**  
   Submenus are not a self-contained public feature in the current popup component, and several native fields need stable public semantics. A smaller initial contract is easier to maintain and extend.

- Related pointer-event API discussion: https://github.com/noctalia-dev/noctalia/issues/3341
- Declarative plugin-panel drag and drop: https://github.com/noctalia-dev/noctalia/pull/3542
- Existing native implementation: `ContextMenuPopup`, `ContextMenuControl`, and the launcher app-actions menu in this repository.

### Additional context

A follow-up community-plugin PR can migrate Clipper to this API after the core contract is accepted and merged. Keeping that migration separate allows the core API and lifecycle guarantees to be reviewed independently.

The proposed implementation should start from a clean branch/worktree based on the maintainer-selected target branch and should not depend on unrelated local feature work.

blackbartblues:
## Summary

- add a generic `panel.openContextMenu(request)` API for plugin panels
- anchor the native `ContextMenuPopup` at the exact originating right-click position and press serial
- support action rows, disabled actions, headers, separators, scalar callback context, and bounded scrolling
- keep pointer metadata host-owned and carry the menu request as a one-shot result that is never replayed
- support regular and persistent plugin panels, including popup exclusivity, focus-grab integration, configured popup shadows, and lifecycle cleanup

## Plugin API

```luau
local opened = panel.openContextMenu({
  items = {
    { kind = "header", label = "Clipboard" },
    { kind = "item", id = "copy", label = "Copy" },
    { id = "delete", label = "Delete", enabled = canDelete },
    { kind = "separator" },
  },
  onActivate = "onContextMenuAction",
  context = entryId,
  maxVisible = 12,
})

function onContextMenuAction(actionId, context)
  -- Plugin-defined behavior.
end
```

Action rows accept an explicit `kind = "item"` or an omitted kind. The call returns `false` when it is not made from a live direct pointer callback. Malformed requests raise descriptive Luau argument errors.

Initial limits match the accepted issue contract: 64 items, `maxVisible` 1–30, 128-byte IDs/callback names, and 512-byte labels. IDs must be unique and each menu must contain at least one enabled action.

## Lifecycle and safety

- `panel.close()` wins when the same callback also requests a menu.
- Requests from a closed, reopened, or reloaded panel generation are ignored.
- Menus close on panel close, reload, disable/unregister, replacement, outside click, Escape, or activation.
- Popup-level deferred activation and dismissal use a lifetime token, so immediate panel unregistration cannot dereference a destroyed popup.
- Activation callbacks are also guarded against panel destruction and stale open generations.
- The original right-click callback signature remains unchanged; Wayland coordinates and serials are never exposed to Luau.
- Plugin menus use the effective `shell.shadow` configuration.
- The API is introduced as plugin API 28.

## Testing

- `context_menu_popup_test`: normal activation and activation followed by immediate `PanelManager::unregisterPanel()`
- `plugin_bindings_test`: valid requests, explicit/implicit action kinds, missing pointer context, scalar context, duplicate IDs, empty-action rejection, and invalid context
- `ui_tree_reconciler_test`: right-click pointer metadata propagation
- `input_dispatcher_test`: initiating press coordinates, serial, and time
- `plugin_manifest_test`: API-level regression coverage
- debug core build completed successfully

## Companion updates

- Luau declarations: https://github.com/noctalia-dev/official-plugins/pull/46
- Plugin developer documentation: https://github.com/noctalia-dev/noctalia-docs/pull/226

Closes #3899


blackbartblues:
@ItsLemmy ?

ItsLemmy:
### P1 — Deferred activation can use a destroyed popup                                             

PluginPanel now dynamically owns ContextMenuPopup in src/shell/panel/plugin_panel.cpp:379-380, and 
unregistering a plugin immediately erases the panel at src/shell/panel/panel_manager.cpp:473-485.  

Menu activation is deferred with a raw ContextMenuPopup*:                                          

```cpp                                                                                             
DeferredCall::callLater([self, onActivate, e]() {
self->close();
// ...
});
```                                                                                                

src/ui/controls/context_menu_popup.cpp:175-186                                                     

If plugin configuration reload/unregistration runs after the pointer event but before the next     
deferred-call drain, PluginPanel::~PluginPanel() destroys the popup. The deferred callback then    
calls self->close() on freed memory. The m_alive guard in the plugin activation callback runs too  
late; the popup has already been dereferenced.                                                     

Use a lifetime token for ContextMenuPopup’s deferred activation/dismissal callbacks, or otherwise  
guarantee the popup survives queued callbacks. Add coverage for activation followed by immediate   
panel unregister.                                                                                  

### P1 — Plugin API 27 is already allocated                                                        

PR #3900 merged and current main defines:                                                          

```cpp                                                                                             
kInputFrameVisibilityPluginApiVersion = 27;
```                                                                                                

This PR still defines the context-menu API as 27. The branch is currently DIRTY. Rebase and        
allocate API 28.                                                                                   

### P2 — Explicit kind = "item" violates the accepted API contract                                 

src/scripting/plugin_bindings.cpp:708-722 accepts only "separator" and "header" when kind is       
present. Issue #3899 defines action rows as omitted or "item". Therefore this valid request raises 
an argument error:                                                                                 

```luau                                                                                            
{ kind = "item", id = "copy", label = "Copy" }
```                                                                                                

Accept "item" as ScriptContextMenuItemKind::Action and cover it in plugin_bindings_test.           

### P2 — Plugin menus ignore configured shadow direction/alpha                                     

PluginPanel::openContextMenu() creates the popup and opens it without calling setShadowConfig().   
Other native panel context menus copy config().shell.shadow before opening. Plugin menus           
consequently use the ShadowConfig defaults instead of the user’s configured shadow direction and   
alpha.                                                                                             

Expose the effective shadow config through the panel host and apply it before open().              

Required cross-repository updates                                                                  

- official-plugins/noctalia.d.luau still has no panel.openContextMenu declaration or request/item  
types. Project guidance requires plugin-facing bindings to update this file in a separate        
commit.                                                                                          
- noctalia-docs/src/content/docs/noctalia/plugins/development/declarative-ui.mdx still says        
persistent panels have no context menus. Update as the documented follow-up.

ItsLemmy:
Thanks

PR #3907, opened 2026-08-13
fix: session and lock-screen reliability (screencopy freeze, dropped lock requests, logind session lookup)

### LINKED ISSUE #3848: [BUG] Native lock screen intermittently hangs indefinitely on "authenticating", unrecoverable without switching to a TTY and rebooting
### Submission checklist

### Bug description

The native v5 lock screen intermittently freezes on the authenticating state after a password is submitted, roughly one in a handful of unlocks. There is no visible recovery path once it happens; the compositor eventually force-releases input devices after roughly a minute. The only way back to a usable session is switching to a TTY and issuing a reboot. The compositor/session itself never recovers on its own, though the underlying system is otherwise responsive enough to reboot cleanly, so this is scoped to the compositor/Noctalia session rather than a kernel-level hang.

### Steps to reproduce

1. Launch noctalia (reproduced both via `noctalia -d` detached mode and the documented `exec-once = noctalia` in the compositor config; same intermittent hang under both)
2. Lock the session via the native lock screen
3. Submit the password to unlock
4. Intermittently (not every attempt), the screen stays on "authenticating" indefinitely

### Expected behavior

Password submission completes authentication and unlocks the session, or fails with a visible error if the password is wrong.

### Actual behavior

The lock screen hangs indefinitely on "authenticating" with no error, no timeout, and no way to cancel or retry. Only recoverable by switching to a TTY and rebooting.

### Logs / error output

```text
09:04:26.128 [INF] [lockscreen] session lock requested
09:04:26.133 [INF] [lockscreen] session is locked
09:04:27.716 [DBG] [pam] authenticating user='admin' service='noctalia-lock'
   ... complete silence in the log for 68 seconds, no activity anywhere in the app ...
09:05:36.026 [DBG] [timer] timer callback ... took 172.5ms
09:05:36.027 [INF] [seat] touch: released
09:05:36.027 [INF] [seat] pointer: released
   ... 15 more seconds ...
09:05:51.039 [INF] [app] shutdown
```

### Noctalia version

noctalia-git-5.0.0.r5028.g19721f1de-1

### Compositor

Hyprland

### Distribution

Arch-based

### Environment information

```text
Init system: dinit (Artix Linux, not systemd)
Session backend: userspawn

Note: this build includes a local source patch, not upstream. My PKGBUILD's
prepare() step renames the hardcoded PAM service in
src/shell/lockscreen/lock_screen.cpp from "login" (stock upstream, per the
FAQ) to "noctalia-lock":

  sed -i 's/const std::string pamService = "login";/const std::string pamService = "noctalia-lock";/' src/shell/lockscreen/lock_screen.cpp

I made this patch partway through debugging to isolate the lock screen's
PAM call from the system's real `login` service, which my session-tracking
backend (previously turnstile which I also switched out in an attempt to troubleshoot this but is now userspawn) hooks into for session
accounting. /etc/pam.d/noctalia-lock does not ship with the package; I
created it myself (auth+account include system-auth, no session stack) to
give the renamed service something to authenticate against.
```

### Additional context

Because the lock screen's PAM service name is normally hardcoded with no
runtime override (confirmed via the FAQ: "Noctalia v5 authenticates through
PAM using the login service"), I patched it locally to a dedicated service
name so I could rule out interference from session-tracking PAM modules
(pam_turnstile.so / pam_rundir.so under userspawn) that live in the real
`login` service's session stack. This eliminated an earlier, different
symptom I was chasing (a spurious session-close event correlating with a
Wayland disconnect), but the hang described above persisted even after
that isolation.

To rule the PAM stack out entirely, I tested progressively:
1. Confirmed the faillock tally file for the account was empty with no stale lock, ruling out tally database contention.
2. Confirmed YESCRYPT_COST_FACTOR is unset (default cost 5), ruling out a misconfigured slow hash.
3. As a final, conclusive test, replaced noctalia-lock's entire auth stack with just pam_permit.so, a module that does nothing but return success immediately with zero I/O, no locking, no crypto:

#%PAM-1.0
auth      required   pam_permit.so
account   include    system-auth

The hang still occurred with this trivial stack in place. Since pam_permit.so cannot itself block on anything, this rules out the PAM module chain entirely.

Stock "login" service, my isolated "noctalia-lock" service, and a trivial always-succeed stack all reproduce the same intermittent hang. That progression points at Noctalia's own PAM handling rather than the system's PAM or session configuration: likely a deadlock or stalled synchronization between the PAM auth worker and the UI/main thread. The intermittency is consistent with a race condition rather than a deterministic bug.

Any pointers on where the PAM conversation callback is implemented in src/auth/ would help narrow this further, but functionally this needs a timeout/cancellation path on the auth call so a stalled PAM conversation cannot take down the entire compositor session.

nocstah:
## Summary

Three independent fixes to session/lock reliability, found while running Noctalia as my daily shell on Fedora + Hyprland. Each is a separate commit and they can be reviewed (or cherry-picked) independently:

1. **`fix(capture)`** — bound the blocking screencopy wait so a frame that never completes cannot freeze the shell.
2. **`fix(lockscreen)`** — log lock requests that are dropped because a lock is already active.
3. **`fix(logind)`** — resolve the logind session via the user's Display session when the PID lookup fails.

> **Updated:** the PAM fork-deadlock fix that was originally the first commit here has been split into **#3908**, since it closes #3848 and deserves to be reviewed on its own. This PR no longer contains it. The two branches touch no files in common and can be merged in either order.

### 1. Blocking screencopy can spin forever

`captureOutputBlocking()` looped `wl_display_roundtrip()` with no deadline. When the compositor never delivers `ready`/`failed` for a frame — which I reproduced with lock-screen desktop snapshots on Hyprland, on both outputs, every time — the loop spins forever inside `LockScreen::lock()` on the main loop. The whole shell freezes, IPC goes dark, and the session lock is never even requested, so the screen stays unlocked.

Now bounded to 2 seconds, cancelling the in-flight capture on timeout (the completion callback holds stack references), letting the caller fall back to the wallpaper background.

### 2. Silently ignored lock requests

`lock()` returned `true` silently when `isActive()`. A stale `m_locked` flag — which I saw after an external locker plus a compositor lock-restore sequence — then made every subsequent lock request look successful while no-op'ing. Logging the state makes the condition diagnosable; a shell restart clears it.

### 3. logind session resolution under the systemd user manager

Both session lookups assumed the shell runs inside the login session's cgroup. Started from the systemd user manager it does not: `user@.service` sits outside that cgroup, so `GetSessionByPID` answers `NoSessionForPID`, and `XDG_SESSION_ID` is not in that manager's environment either.

The whole logind integration was dark as a result: the session lock monitor logged `disabled: session path unavailable`, so `loginctl lock-session` never reached the lock screen; the idle-inhibit monitor never armed; and brightness fell back to the `auto` session path, which resolves against the caller and is the same dead end. Asking logind for this user's Display session as a last resort fixes all three, which is why `brightness_service.cpp` carries the same fallback in that commit.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Build / packaging

## Related Issue

None directly. Companion PR: #3908 (PAM fork deadlock, closes #3848).

## Testing

- `just build release` — clean, no new warnings, with these three commits applied directly on top of `main` after the split.
- `clang-format` v22.1.8 run over every touched file.
- Running as my daily shell since 2026-08-10 on Fedora 44 / Hyprland, through many lock/unlock and suspend/resume cycles. The screencopy freeze reproduced on every lock attempt beforehand and none since. The logind fix is verifiable from the log line: `logind session lock monitor active (/org/freedesktop/login1/session/_33)` where it previously read `disabled: session path unavailable`.

## Manual Coverage

- [ ] Tested on Niri
- [x] Tested on Hyprland
- [ ] Tested on Sway

Single machine only — MacBookPro15,1, Fedora 44, Hyprland, two external 4K outputs plus the internal panel. I have no Niri or Sway install to verify against. The capture fix is a deadline on an existing wait and is compositor-independent in nature; the logind fix is specific to being started from the systemd user manager rather than to any compositor.

## Screenshots / Videos

Not applicable — no UI, visual, or layout changes.

## Checklist

## Additional Notes

The 2-second screencopy deadline is a chosen constant. If you would rather it were configurable, or a different default, that is an easy change.

Happy to split these three further if you would prefer one PR per fix.


decksters-lab:
Im excited for this!

nocstah:
Split the PAM fork-deadlock commit out into #3908 — it closes #3848 and is easier to review on its own. This PR is now the remaining three fixes (screencopy timeout, dropped-lock-request logging, logind session fallback) and no longer touches `src/auth/` or `src/main.cpp`. Both branches build cleanly on top of `main` independently and share no files, so they can be merged in either order.

Ly-sec:
Thanks for splitting these fixes out @nocstah . The lock-request logging change looks good, and the logind session-resolution change is a useful direction.

I would request changes on the screencopy timeout before merging. The deadline is checked only after `wl_display_roundtrip()` returns. If that call itself blocks because the Wayland connection stops making progress, the timeout cannot fire. Repeated blocking roundtrips with `sleep_for()` also do not fit our poll-based event-loop model.

Could this use a deadline-aware wait on the Wayland file descriptor instead, then dispatch events while progress is available? Please also add coverage for the timeout and cancellation path, including that a late completion callback cannot access the stack references captured by `captureOutputBlocking()`.

For the logind fallback, please extract the duplicated session-resolution logic into logind-owned shared code. `User.Display` is a sensible best-effort path for a shell started by the systemd user manager, but it may select a different session when a user has multiple concurrent sessions, so that behavior should be explicit and tested.

nocstah:
Thanks for the review — all three addressed in the two follow-up commits.

> The deadline is checked only after `wl_display_roundtrip()` returns…

Fixed — the roundtrip is gone. `captureOutputBlocking()` now waits on the Wayland fd directly: `dispatch_pending` → `prepare_read` → `flush` → `poll(fd, remaining)` → `read_events` → `dispatch_pending`. The wait is bounded by what's left of the budget rather than being a blocking call with a deadline checked around it, and there's no `sleep_for` — each pass either makes progress or returns when its slice of the deadline expires.

> Please also add coverage for the timeout and cancellation path…

`tests/screencopy_blocking_test.cpp`. The capture and wait sides moved into `screencopy_blocking.{h,cpp}` behind small injectable ops, so the tests drive a virtual clock with no compositor: synchronous completion, completion after N pumps, timeout with cancellation, no single wait exceeding the remaining budget, one wait consuming the whole budget, and dispatch failure.

The last case is the one you asked about specifically: completion state now lives in a `shared_ptr` that gets flagged abandoned when we give up, so the callback captures nothing from the caller's stack. The test fires a completion *after* `runBlockingCapture()` has returned and asserts the caller's frame and error are untouched.

> For the logind fallback, please extract the duplicated session-resolution logic into logind-owned shared code.

Done — `src/dbus/logind/logind_session.{h,cpp}`. Both `LogindService` and `BrightnessService` call `logind::resolveSession()`, which returns the path plus the lookup that answered (`XDG_SESSION_ID`, pid, or user Display); both log that source. `BrightnessService` keeps its `auto` path only for when nothing resolves.

> …it may select a different session when a user has multiple concurrent sessions, so that behavior should be explicit and tested.

Agreed. The resolver now reads the user's full `Sessions` list and, when there's more than one, warns naming the session it guessed instead of returning it silently:

```
[WRN] [logind] logind: 2 concurrent sessions for this user, falling back to the
      Display session 3 (/org/freedesktop/login1/session/_33) — it may not be the
      session this process runs in
```

The lookups are injectable, so `tests/logind_session_test.cpp` covers the order (id wins, no fallback consulted), a stale `XDG_SESSION_ID` falling through to the pid, the systemd-user-manager case, the multi-session guess, no display session, and every lookup dark.

Left as two follow-up commits so the review diff stays readable — happy to squash into the originals if you'd prefer.

ItsLemmy:
@Ly-sec

Ly-sec:
Hey @nocstah I am very sorry for the super late reply! There are a few things that still need to be solved, after that I'll gladly merge it.

1. **Pre-lock capture has an N × 2-second timeout.** Each output is captured sequentially before the session-lock object is created. Use one batch deadline or concurrent captures so a stalled compositor cannot delay locking per monitor.

2. **logind's no-display value is mishandled.** `User.Display` uses `("", "/")` when no display session exists, but the resolver only rejects an empty path. Reject the empty ID/root path and test the real sentinel.

3. **Wayland flush backpressure is incomplete.** After `wl_display_flush()` returns `EAGAIN`, the code polls only for `POLLIN`. Poll for `POLLOUT` and retry the flush within the same deadline.

4. **The added tests fail clang-tidy.** Two raw-string warnings and one redundant member-initializer warning are treated as errors.

## Verification

Configure, build, targeted tests, formatting, and production-file lint passed. Linting the added test files failed. The PR is currently conflicting with `main`; retain current main's `SetLockedHint` behavior during resolution.

Issue #3277, opened 2026-07-05
[BUG] Lockscreen password authentication blocks on pam_fprintd due to hardcoded "login" PAM service


MeinardEdrei:
### Submission checklist

- [x] I have searched existing issues and confirmed this is not a duplicate.
- [x] I am using the latest available version of Noctalia.

### Bug description

In Noctalia v5, lockscreen password authentication is hardcoded to use the `"login"` PAM service inside `src/shell/lockscreen/lock_screen.cpp`.

Although the comments in the source code indicate that the developers intended to bypass/strip `pam_fprintd.so` from this stack (since Noctalia drives the biometric fingerprint reader itself directly over D-Bus), the code simply passes `"login"` directly to `PamAuthenticator`.

Because `/etc/pam.d/login` includes fingerprint authentication (via `system-local-login` -> `pam_fprintd.so` on many systems like Arch Linux), the password prompt blocks for 30 seconds waiting for a fingerprint swipe before it times out and finally verifies the entered password via `pam_unix.so`.

### Steps to reproduce

1. Install Noctalia on a system with a fingerprint reader configured under `/etc/pam.d/login` (via `system-local-login`).
2. Set `fingerprint = false` in `noctalia/config.toml` (or keep it enabled—the bug occurs in both states).
3. Lock the session.
4. Type your password and press Enter.

### Expected behavior

Password authentication should be evaluated immediately (e.g., using `system-auth` or `common-auth` which don't invoke biometric blockers, or a custom `noctalia` PAM service file).

### Actual behavior

Pressing Enter on the password input freezes the lock screen authentication status for 30 seconds while `pam_fprintd` waits for a fingerprint swipe in the background. Only after the 30-second timeout does the password successfully verify.

### Logs / error output

```text
Relevant journalctl output shows a 31-second delay between the Fingerprint daemon starting up and the fallback unix password authentication executing:

    Jul 05 15:02:28 chico systemd[1]: Starting Fingerprint Authentication Daemon...
    Jul 05 15:02:28 chico systemd[1]: Started Fingerprint Authentication Daemon.
    ...
    Jul 05 15:02:59 chico noctalia[23507]: pam_unix(login:account): setuid failed: Operation not permitted
    Jul 05 15:03:29 chico systemd[1]: fprintd.service: Deactivated successfully.
```

### Noctalia version

5.0.0_beta1-1

### Compositor

Niri

### Distribution

Arch-based

### Environment information

```text
[...]

PR #3908, opened 2026-08-13
fix(auth): run PAM in a re-exec'd helper, not a raw fork child (fixes lock screen hanging on "authenticating")

### LINKED ISSUE #3848: [BUG] Native lock screen intermittently hangs indefinitely on "authenticating", unrecoverable without switching to a TTY and rebooting

nocstah:
## Summary

Run the PAM conversation in a re-exec'd helper process instead of a raw `fork()` child, so it can no longer inherit a locked mutex from the threaded shell and hang the lock screen forever.

This is the fix for #3848.

## Motivation

The PAM conversation ran in a `fork()` child of the heavily threaded shell. After `fork()` only the calling thread survives in the child, but the full memory image comes along — including any mutex another thread happened to hold at that instant. Malloc arena, logging, locale: if one of those was locked at fork time, it stays locked forever in the child, which then hangs before it ever reaches PAM while the parent blocks reading the result pipe. The lock screen sits on `authenticating` with no error and no timeout, and the session cannot be unlocked.

That is exactly the failure described in #3848: intermittent (roughly one unlock in a handful), no visible recovery path, log completely silent after `[pam] authenticating` — 68 seconds of nothing in the reporter's excerpt — and only recoverable by switching to a TTY and rebooting. It is intermittent precisely because it depends on which lock happens to be held at the moment of the fork, which is why it reproduces under load and on some machines far more than others. The reporter saw it on both Hyprland and mango; I hit it twice in one morning on idle-lock unlock attempts on Hyprland.

Because only async-signal-safe functions are legal between `fork()` and `exec()` in a multithreaded process, the fix is to reach `exec()` immediately rather than to do PAM work in the child:

- The fork now `exec`s straight into a hidden `noctalia pam-helper <service>` mode, so PAM runs in a clean, single-threaded process image with no inherited locks.
- The password goes in over stdin and the `Result` comes back over stdout.
- Only async-signal-safe calls happen between fork and exec.
- `SIGPIPE` is blocked on the auth thread, so a helper that dies early surfaces as `EPIPE` rather than killing the shell.

`pam-helper` is a hidden argv mode rather than a separate installed binary: it re-execs `/proc/self/exe` and is dispatched in `main()` alongside the existing `--daemonize` handling, so there is no new artifact to package.

Note this is a **different root cause** from #3277 (the hardcoded `login` PAM service blocking on `pam_fprintd`). This PR does not change the service name, so #3277 is unaffected and still stands.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Build / packaging

## Related Issue

Closes #3848

## Testing

- `just build release` — clean, no new warnings, with this commit applied directly on top of `main`.
- `clang-format` v22.1.8 run over every touched file.
- Running as my daily shell since 2026-08-10 on Fedora 44 / Hyprland, through many lock/unlock and suspend/resume cycles. Before this change I could reproduce the `authenticating` hang within a day of ordinary use; since it, zero occurrences. Correct and incorrect passwords both still behave as expected, including the failure path and retry.
- Verified the helper dying mid-authentication surfaces as an error rather than taking the shell with it.

## Manual Coverage

- [ ] Tested on Niri
- [x] Tested on Hyprland
- [ ] Tested on Sway

Single machine — MacBookPro15,1, Fedora 44, Hyprland, two external 4K outputs plus the internal panel. I have no Niri or Sway install to verify against, though this path is compositor-independent by nature.

## Screenshots / Videos

Not applicable — no UI, visual, or layout changes.

## Checklist

## Additional Notes

Split out of #3907 at the maintainers' convenience — that PR carries three further, unrelated session/lock fixes (a screencopy timeout, a dropped-lock-request log line, and a logind session lookup fallback) and no longer contains this commit. The two do not overlap in files and can be merged in either order.


Ly-sec:
Hey @nocstah , looks good to me. Can you please clean up the comments a bit and resolve the conflict(s)? Afterwards I'll gladly merge it.

PR #3912, opened 2026-08-13
widgets/sysmon: show highlight color setting even without visualization

Ape:
Show the highlight color setting when visualization is None, since glyph and value are still colorized.


ItsLemmy:
Thanks!

PR #3913, opened 2026-08-13
render: weight HSV hue interpolation by chroma

Ape:
Give low-chroma endpoints proportionally less influence over the interpolated hue. This prevents near-achromatic colors from causing unrelated hue detours while preserving shortest-path interpolation and the existing behavior for colors with equal chroma.

image


ItsLemmy:
Thanks
I wanted to keep lerpHSV vanilla since it's mathematically correct.
Made yours a separate helper and used it in both sysmons widget/desktop widget.

PR #3914, opened 2026-08-13
settings: render editor sheets in the Settings window

### LINKED ISSUE #3718: [BUG] (Hyprland) settings' pop-up subwindows disappearing
### Submission checklist

### Bug description

Changing window focus while settings window is tiled makes any pop-up windows disappear (eg. individual widget settings, wallpaper directory selector etc.)
This does not happen if the settings window is fullscreen or floating.
The pop up appears again when the window gains focus again.
possible related to https://github.com/noctalia-dev/noctalia/issues/3609 ?

### Steps to reproduce

1. open noctalia settings
2. tile the window
3. go to wallpaper section
4. try to change wallpaper directory, leave the window that pops up open
5. change focus to another window
6. the wallpaper directory selection pop-up window disappears

### Expected behavior

The pop-up windows should stay visible.

### Actual behavior

They disappear.

### Logs / error output

```text

```

### Noctalia version

noctalia v5.0.0 (97917d9ca07e)

### Compositor

Hyprland

### Distribution

Arch-based

### Environment information

```text
installed from cachy repo
RX 6700XT
```

### Additional context

_No response_

Ape:
## Summary

Fixes #3718.

Render Settings editor sheets and their nested dialogs inside the Settings toplevel instead of creating separate Wayland popup surfaces.

## Design

The original `xdg_popup` implementation was not the best semantic match for editor sheets. An `xdg_popup` represents a transient compositor-managed surface whose visibility and lifetime can depend on focus, grabs, and input serials. A Settings editor is persistent modal content belonging to the Settings window, so compositors could hide it when that window lost focus even though Settings itself remained visible.

This change separates the retained sheet content from its presentation and mounts it in an in-scene modal stack owned by the Settings window. The modal stack provides the backdrop, layout, input isolation, keyboard navigation, and nested-modal ordering while remaining part of the same `xdg_toplevel`.

Color, glyph, and file dialogs opened from Settings use the same modal stack. Their existing popup presenters remain in use when those dialogs are opened elsewhere, preserving the appropriate transient behavior for other shell surfaces.

Modal content is retained across Settings scene rebuilds. Focus is restored by a stable tab position instead of pointers to controls that may have been destroyed, and retained content is rebound to the current dropdown host when reattached. Active dropdowns are dismissed before replacing that host, preventing controls from retaining stale popup context.

This also removes the focus-wait and input-serial workarounds previously needed to open editor sheets as `xdg_popup` surfaces.


ItsLemmy:
Very nice, thanks

Issue #3915, opened 2026-08-13
[BUG] Labwc apply.sh file will override user-defined compositor font


na-cb94f7:
### Submission checklist

- [x] I have searched existing issues and confirmed this is not a duplicate.
- [x] I am using the latest available version of Noctalia.

### Bug description

When using Noctalia and Labwc together, and when a theme is refreshed, I notice that the apply.sh file meant for colors and the used theme will also override the set fonts.

### Steps to reproduce

1. Enable the Labwc template
2. Change Noctalia theme (pre-defined, wallpaper-based, custom...)
3. Call `labwc --reconfigure to apply the new configuration`

### Expected behavior

Only the settings meant for the theme will change, and no changes are done to `rc.xml `except for choosing the Noctalia theme.

### Actual behavior

As the `apply.sh` script changes the text to select the Noctalia font, it also overrides font settings. Since `noctalia` is not a font, Labwc reverts to its default font and the user must manually change the font.

### Logs / error output

```text
No relevant logs to add
```

### Noctalia version

v5.0.0 (5.0.0_beta.8-1-dirty)

### Compositor

Labwc

### Distribution

Arch-based

### Environment information

```text
Installation method : pacman -S noctalia
```

na-cb94f7:
After some testing, this sed command seems to be the problem :

`
sed '/<theme>/,/<\/theme>/s|<name>.*</name>|<name>noctalia</name>|' "$rc_file" >"$tmp_file"`

na-cb94f7:
I've never used `sed` before, so I'm taking a long time to fix this haha. I'm trying to make it so that it looks at different addresses (inside the <theme> block and NOT inside the <font> blocks), but I can't really find documentation on nested addresses. Is that something `sed` can do or is something like `awk` better-suited for that ?

na-cb94f7:
Nevermind, the following seems to work :

`sed '/<theme>/,/<\/theme>/ { /<font/,/<\/font>/! s|<name>.*</name>|<name>noctalia</name>|; }' $testfile`

ItsLemmy:
Fixed by #3918

PR #3918, opened 2026-08-13
Prevent apply.sh from overriding user-defined font

na-cb94f7:
The 'sed' command on line 50 was changed, so that the replace operation would be done inside of the `<theme> `block but ignoring `<font>` blocks inside of the former.

See issue #3915 for details

## Summary

What changed was a `sed` command that would replace all occurences of `<name>*</name>` with `<name>noctalia</name>`, inside of the `<theme>` block.

## Motivation

The problem was that this would override the theme (expected behavior), but also override user-defined fonts (unexpected behavior, especially as it would tell Labwc to use the `noctalia` font which doesnt exist).

The problem has been solved by doing the operation inside of the `<theme>` block, _while_ ignoring the `<font>` blocks.

Now, the user can use the Labwc template to synchronise Labwc with the Noctalia theme, without losing their font declarations.

## Type of Change

- [X] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Build / packaging

## Related Issue

Closes issue #3915 

## Testing

After changing the `sed` command on line 50, I have done the following :

1. Toggled the Wallpaper widget
2. Made sure that I have set my own font declarations for Labwc
3. Changed wallpaper to trigger the apply.sh script, as intended
4. Observed the used Labwc font, which has stayed the same since step 2

Due to the extremely small change, I've done no formal testing though `just format` and such.

## Manual Coverage

- [ ] Tested on Niri
- [ ] Tested on Hyprland
- [ ] Tested on Sway
- [X] Tested on another compositor: Labwc

## Screenshots / Videos

None to be added

## Checklist

## Additional Notes

None to be added

### DIFF
--- a/assets/templates/labwc/apply.sh
+++ b/assets/templates/labwc/apply.sh
@@ -46,8 +46,16 @@ if grep -q '<theme>' "$rc_file"; then
         exit 1
     fi
 
-    if sed -n '/<theme>/,/<\/theme>/p' "$rc_file" | grep -q '<name>.*</name>'; then
-        sed '/<theme>/,/<\/theme>/ { /<font/,/<\/font>/! s|<name>.*</name>|<name>noctalia</name>|; }' "$rc_file" >"$tmp_file"
+    if sed -n '/<theme>/,/<\/theme>/ {
+        /<font[[:space:]>].*<\/font>/d
+        /<font[[:space:]>]/,/<\/font>/d
+        /<name>.*<\/name>/p
+    }' "$rc_file" | grep -q .; then
+        sed '/<theme>/,/<\/theme>/ {
+            /<font[[:space:]>].*<\/font>/b
+            /<font[[:space:]>]/,/<\/font>/b
+            s|<name>.*</name>|<name>noctalia</name>|
+        }' "$rc_file" >"$tmp_file"
     else
         sed '/<theme>/a\    <name>noctalia</name>' "$rc_file" >"$tmp_file"
     fi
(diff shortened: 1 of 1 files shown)


na-cb94f7:
I'll add this : it would be really cool to have such a `noctalia` font, so that you can synchronize fonts between compositors and shell without any additional configuration !

ItsLemmy:
You can use the classic fontconfig, ex: ~/.config/fontconfig/fonts.conf

And let noctalia use "System default"

ItsLemmy:
- [P2] Detect the theme name outside <font> before selecting the replacement branch —              
   assets/templates/labwc/apply.sh:49-50                                                           
   Line 49 still considers any <name> within <theme> a theme name, including font names. For a      
   valid configuration containing fonts but no direct theme name:                                   
```xml                                                     
<theme>
  <font place="ActiveWindow">
    <name>Inter</name>
  </font>
</theme>
```                                                                                              
   the condition succeeds, but the new substitution skips the font block and writes the file        
   unchanged. <name>noctalia</name> is never inserted, so the template is not applied. Labwc’s      
   canonical rc.xml.all demonstrates this valid structure with an omitted theme name and nested     
   font names. The existence check must use the same font exclusion as the replacement.             
 - [P2] Handle single-line <font> elements without leaking the sed range —                          
   assets/templates/labwc/apply.sh:50                                                               
   With valid compact XML such as:                                                                  
 ```xml                                                                                           
<font place="ActiveWindow"><name>Inter</name></font>
<name>Clearlooks</name>
 ```                                                                                              
   sed starts the /<font/,/<\/font>/ range on the first line but does not test the terminating      
   address on that same line. The range therefore remains active and skips the following theme      
   <name>, leaving Clearlooks unchanged. Avoid a two-address range for font exclusion, or           
   explicitly handle elements whose opening and closing tags share a line.

PR #3927, opened 2026-08-14
feat(widgets/volume): add "Hide When Inactive" toggle to volume widget

### LINKED ISSUE #3917: [FEATURE] Option to hide input volume panel widget when microphone not in use
### Submission checklist

### Feature type

UI / visual improvement

### Feature summary

An option to hide the input volume panel widget when not in use. This allows the microphone's volume to be adjusted/muted when the microphone is in use, while not taking up space when it either isn't in use or no microphone is plugged in.

### Motivation / use case

This would contribute to a cleaner bar that only exposes controls for devices when applicable for people who want a clean UI that only shows relevant information. The lack of a microphone icon when this option is turned on would imply that the microphone is not in use.

### Proposed solution

With this feature, the input volume widget that shows the volume of the current input device would have a toggle in the widget's settings that allows the widget to be hidden when not in use.

### Alternatives considered

- Using the privacy indicator to see the microphone in use and using keyboard controls to adjust/mute the mic volume. This comes with the downside of not being able to immediately see what volume the mic is set at or whether or not it is actually muted.
- Using the input volume widget, without autohiding. This is unnecessary most of the time when the mic is unplugged or not in use. When the mic is unplugged it simply shows 0% with a crossed-out microphone icon, which is unproductive for people who prefer an interface that only shows relevant information.

_No response_

### Additional context

_No response_

Aron-dxd:
## Summary
This pull request adds an optional toggle to hide the volume widget when it is not in use. When enabled, the widget will hide itself unless the device(output or input) is currently in use. In the case of inputs, it's hidden unless some application is capturing audio from a microphone. For outputs, the widget is hidden unless any application is currently playing audio. 

## Motivation
An input volume widget that is always visible even when nothing is capturing audio adds clutter to a bar for those who want only relevant controls. This lets people configure the widget to be hidden when it isn't relevant.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Build / packaging

## Related Issue

Closes #3917

## Testing

- Ran `just format` and `just build`. Both passed.
- `just test` passes 79/80. `process_test` fails in the same manner on a clean `main`
- Built and ran the debug binary directly.

## Manual Coverage

- [x] Tested on Niri
- [ ] Tested on Hyprland
- [ ] Tested on Sway

## Checklist

## Additional Notes
The original issue only asked for a toggle in the case of microphone. A single setting for when a device is not in use in both input and output cases seemed more consistent than adding a setting that silently does nothing for output widgets or conditionally rendering the toggle only when input is selected.  I also didn't find any widgets that do conditional rendering of a setting based on the value of another. 

This PR also only checks capture activity and not device presence. A mic that is connected but idle will be treated the same as no mic present. In both cases, the widget will be hidden when the toggle is enabled.

### DIFF
--- a/assets/translations/en.json
+++ b/assets/translations/en.json
@@ -3506,6 +3506,10 @@
           "description": "Hide the battery widget when fully charged",
           "label": "Hide When Full"
         },
+        "hide-when-inactive": {
+          "description": "Hide the volume widget when not in use",
+          "label": "Hide When Inactive"
+        },
         "hide-when-no-connected-device": {
           "description": "Hide the Bluetooth widget when no device is connected",
           "label": "Hide When No Connected Device"
--- a/assets/translations/en.json
+++ b/assets/translations/en.json
@@ -3507,7 +3507,7 @@
           "label": "Hide When Full"
         },
         "hide-when-inactive": {
-          "description": "Hide the volume widget when not in use",
+          "description": "Hide the microphone widget while no application is capturing audio",
           "label": "Hide When Inactive"
         },
         "hide-when-no-connected-device": {
--- a/src/pipewire/pipewire_service.cpp
+++ b/src/pipewire/pipewire_service.cpp
@@ -1233,13 +1233,15 @@ void PipeWireService::onNodeInfo(std::uint32_t id, const pw_node_info* info) {
   const bool isStream = isProgramStreamClass(nd.mediaClass);
   const bool isPrivacyCandidate = isPrivacyCandidateClass(nd.mediaClass);
   const bool wasStreamReady = nd.streamClassificationReady;
+  const bool streamStateChanged = isStream && (info->change_mask & PW_NODE_CHANGE_MASK_STATE) != 0;
   if (isStream) {
     nd.streamClassificationReady = true;
   }
   if ((isStream && (!wasStreamReady || filterPropsChanged))
       || wasProgramStream != isStream
       || wasPrivacyCandidate
-      || isPrivacyCandidate) {
+      || isPrivacyCandidate
+      || streamStateChanged) {
     rebuildState();
   }
 
--- a/src/pipewire/pipewire_service.cpp
+++ b/src/pipewire/pipewire_service.cpp
@@ -1151,6 +1151,18 @@ void PipeWireService::onNodeInfo(std::uint32_t id, const pw_node_info* info) {
   const bool wasPrivacyCandidate = isPrivacyCandidateClass(nd.mediaClass);
   bool filterPropsChanged = false;
 
+  if (info->state == PW_NODE_STATE_CREATING) {
+    nd.nodeState = AudioNodeState::Creating;
+  } else if (info->state == PW_NODE_STATE_SUSPENDED) {
+    nd.nodeState = AudioNodeState::Suspended;
+  } else if (info->state == PW_NODE_STATE_IDLE) {
+    nd.nodeState = AudioNodeState::Idle;
+  } else if (info->state == PW_NODE_STATE_RUNNING) {
+    nd.nodeState = AudioNodeState::Running;
+  } else {
+    nd.nodeState = AudioNodeState::Error;
+  }
+
   if (info->props != nullptr) {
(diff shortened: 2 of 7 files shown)


ItsLemmy:
### [P1] Required clang-tidy check fails                                                           
                                                                                            
src/shell/bar/widgets/volume_widget.cpp:113                                                        
                                                                                            
```cpp                                                                                             
isActive = std::any_of(captures.begin(), captures.end(), ...);
```                                                                                                
                                                                                            
The repository’s required scoped lint reports:                                                     
                                                                                            
```text                                                                                            
[modernize-use-ranges,-warnings-as-errors]
```                                                                                                
                                                                                            
Use:                                                                                               
                                                                                            
```cpp                                                                                             
isActive = std::ranges::any_of(captures, [](const auto& capture) {
return capture.kind == PrivacyCaptureKind::Microphone;
});
```                                                                                                
                                                                                            
### [P2] Output stream existence does not mean active playback                                     
                                                                                            
src/shell/bar/widgets/volume_widget.cpp:117                                                        
                                                                                            
```cpp                                                                                             
isActive = !m_audio->state().programOutputs.empty();
```                                                                                                
                                                                                            
programOutputs is the application-volume inventory. PipeWireService::rebuildState() adds every     
classified Stream/Output/Audio node without considering its pw_node_info::state. Applications can  
retain idle or suspended streams after playback stops, so hide_when_inactive = true can leave the  
output widget visible indefinitely.                                                                
                                                                                            
PipeWire distinguishes RUNNING, IDLE, and SUSPENDED node states, but NodeData/AudioNode currently  
discard that state. Either:                                                                        
                                                                                            
- retain node state and count only running output streams; or                                      
- scope this setting to input widgets rather than advertising incorrect output behavior.           
                                                                                            
 Reference: PipeWire node states                                                                    
 (https://docs.pipewire.org/group__pw__node.html#gac493839c0b0a1d0b935969f2e990a2df).

Aron-dxd:
Fixed the warning and added PipeWire node state to `AudioNode`. Now check is only for RUNNING streams.

ItsLemmy:
The prior clang-tidy finding is fixed. The output check now tests AudioNodeState::Running, but the 
 state snapshot is not refreshed when playback transitions.                                         
                                                                                                    
 ### [P1] Rebuild AudioState when node state changes                                                
                                                                                                    
 src/pipewire/pipewire_service.cpp:1154-1164                                                        
                                                                                                    
 onNodeInfo() writes the latest PipeWire state into nd.nodeState, but the rebuildState() condition  
 at lines 1239–1243 does not consider whether that state changed.                                   
                                                                                                    
 For an already-classified Stream/Output/Audio, a state-only callback has:                          
                                                                                                    
 - wasStreamReady == true                                                                           
 - filterPropsChanged == false                                                                      
 - wasProgramStream == isStream                                                                     
 - neither privacy condition set                                                                    
                                                                                                    
 Therefore rebuildState() is skipped. m_audio->state().programOutputs retains its old state, so the 
 widget can:                                                                                        
                                                                                                    
 - remain visible after playback becomes idle or suspended; or                                      
 - remain hidden when an existing idle stream starts running.                                       
                                                                                                    
 Capture the previous state, detect a transition, and include that transition in the rebuild        
 condition. Alternatively, handle PW_NODE_CHANGE_MASK_STATE explicitly. The externally exposed      
 AudioNode::nodeState must be synchronized before consumers inspect it.

Aron-dxd:
Can you check now? 

On a side note, when I was on google meet, the stream stayed in the Running state even after ending the call(without closing the meet tab). But that is probably something with browsers/google meet.

ItsLemmy:
Scoped hide_when_inactive to input only via visibleWhen{"device", {"input"}}. That left the AudioNodeState work with no consumer, so I reverted pipewire_service.* to main, as you saw with Meet, browsers keep streams RUNNING after playback stops, so it wasn't a reliable signal anyway.

Also switched to the existing syncWidgetVisibility() pattern and added a test for the gating. Thanks for the PR!

PR #3928, opened 2026-08-14
feat(calendar): right-click a day to open the default calendar app

andresparrab:
## Summary

v4's calendar launched `gnome-calendar` (hardcoded, gated on that specific app being installed) when a date was clicked. v5's rewrite dropped any external-app handoff entirely — clicking a day in the built-in Calendar just selects it, with no way to jump out to a real calendar app to add or edit an event.

This restores that behavior on **right-click** (left-click still just selects the date, unchanged), generalized to whatever the user has configured as their default calendar app via the `text/calendar` MIME association — the same association `xdg-mime query default text/calendar` and every desktop environment's "Default Applications" settings panel already read/write. Nothing is hardcoded to a specific app (Evolution, Thunderbird, GNOME Calendar, Morgen, whatever the user has set, all work identically), with one deliberate exception noted below.

Two follow-up commits round it out: closing the calendar panel after a successful handoff (no point leaving it open behind the app you just switched to), and a Thunderbird-specific improvement (see below).

## What changed

- **`src/net/default_app_open.{h,cpp}`** (new): a small helper, `net::openDefaultAppForMimeType(mimeType)`, that resolves and launches the default `GAppInfo` for a MIME type via GIO (`g_app_info_get_default_for_type` / `g_app_info_launch`). It mirrors the existing `net::openInBrowser()` pattern in the same directory — same shape, just GIO's native app-launch API instead of shelling out to `xdg-open`. `gio/gio.h` is already linked and used elsewhere in the codebase (`src/system/icon_resolver.cpp`), so this introduces no new dependency.
- **`src/ui/controls/calendar_view.cpp`**: adds `button->setOnRightClick(...)` on each day-cell button, right after the existing `button->setOnClick(selectDate)`. `Button::setOnRightClick` already existed on the `Button` class (used elsewhere for plugin buttons/shortcuts) — calendar day cells just never wired it up. On a successful launch, it also calls `PanelManager::instance().closePanel()`.
- **Thunderbird special-case**: Thunderbird is a mail client first — focusing an already-running instance just resumes whatever tab (usually Mail) was open last, not useful from a calendar click. Where the resolved default app is specifically Thunderbird (checked via `g_app_info_get_id`), it's launched with `-calendar` instead of the generic path — verified against `thunderbird --help` directly ("Go to the calendar tab."), not guessed. Every other app still gets the plain generic launch, unchanged.
- **`meson.build`**: registers the new source file.

## Test plan

- [x] `meson`/`ninja` release build compiles clean, zero new warnings
- [x] Ran the built binary against a real desktop session (niri) in place of the packaged build, on 3 separate machines
- [x] Right-click on a day cell launches the configured default `text/calendar` app with no argument, so it opens focused/ready rather than trying to import a dummy file
- [x] Verified against Thunderbird specifically: right-click reliably lands on the Calendar tab regardless of which tab was previously open, and the calendar panel closes on success
- [x] Left-click day selection, month navigation, and the agenda side panel are unaffected
- [x] Bar, other panels, and plugins unaffected by the change
- [ ] Not tested against Evolution/GNOME Calendar specifically (no local install), but the generic path is identical to what `xdg-open`/every DE's default-apps setting already relies on, so behavior should be app-agnostic by construction for anything without a special-case

## Open questions for maintainers

- Is right-click the right gesture for this, or would you prefer it live elsewhere (e.g. a toolbar button, a long-press, a setting to enable/disable)?
- Any objection to depending on `GAppInfo`/GIO here specifically (vs. shelling out to `xdg-open` the way `net::openInBrowser` does)? I chose the native GIO call since it avoids a subprocess for a no-argument "just launch the app" case, but happy to switch to `xdg-open` shelling if that's the preferred house style for this kind of thing.
- **New**: is a per-app special-case (Thunderbird's `-calendar`) something you'd want upstream at all, or is that scope creep for this PR? It's a real quality improvement for a common multi-mode PIM app, but it does mean the file now has to know about one specific app by name rather than staying fully generic. Could split it into a separate PR if that's cleaner, or extend it to Evolution too (same problem, different flag) if the general approach is welcome.
- Naming/placement of `net::openDefaultAppForMimeType` — open to suggestions if there's a more idiomatic convention I'm missing.

Happy to adjust based on feedback.


ItsLemmy:
Thanks

PR #3931, opened 2026-08-14
fix(theme): scale container saturation offset by headroom

### LINKED ISSUE #3924: [BUG] *_container roles blow out to full saturation on pastel-accent palettes (HSL offset has no headroom check)
### Bug description

The `*_container` colour roles are derived from their base role with a fixed HSL offset of roughly `S +0.15, L -0.35`. The offset is applied with no headroom check, so for any palette whose accents already sit high on the saturation scale the `+0.15` clamps at or near `S = 1.0` while `L` still drops into the midtones. The result is a fully saturated primary sitting on the sRGB gamut wall, where a muted container tint was intended.

This is not specific to one palette, but it only becomes visible on ones with pastel accents. Built-in **Catppuccin** is the clearest case: `primary` `#cba6f7` (a soft mauve) derives `primary_container` `#6b02e9` — a neon purple.

Measured on the four accent roles, same machine, same session, by rendering a template containing only `{{colors.X.default.hex}}` and `{{colors.X_container.default.hex}}`:

**Built-in Catppuccin — pastel accents, S already 0.81–0.92**

| role | source | S | L | derived | S | L | ΔS | ΔL |
|---|---|---|---|---|---|---|---|---|
| `primary` | `#cba6f7` | 0.84 | 0.81 | `#6b02e9` | 0.98 | 0.46 | +0.15 | −0.35 |
| `error` | `#f38ba8` | 0.81 | 0.75 | `#c8043a` | 0.96 | 0.40 | +0.15 | −0.35 |
| `tertiary` | `#94e2d5` | 0.57 | 0.73 | `#1ba891` | 0.72 | 0.38 | +0.15 | −0.35 |
| `secondary` | `#fab387` | 0.92 | 0.75 | `#cf4f00` | 1.00 | 0.41 | +0.08 | −0.35 |

`secondary` shows the clamp directly: the requested `+0.15` can only deliver `+0.08` before saturation tops out.

**Built-in Nord — same transform, accents start at S 0.25–0.43, results are fine**

| role | source | S | L | derived | S | L |
|---|---|---|---|---|---|---|
| `primary` | `#8fbcbb` | 0.25 | 0.65 | `#2e6b69` | 0.40 | 0.30 |
| `error` | `#bf616a` | 0.42 | 0.56 | `#56171d` | 0.58 | 0.21 |
| `tertiary` | `#5e81ac` | 0.32 | 0.52 | `#172a40` | 0.47 | 0.17 |
| `secondary` | `#88c0d0` | 0.43 | 0.67 | `#226e83` | 0.59 | 0.32 |

**`source = "wallpaper"` (m3-tonal-spot) is also unaffected** — it goes through Material Color Utilities with HCT gamut mapping rather than this offset, e.g. `primary` `#d4bbfc` → `primary_container` `#503c74`.

So the pattern is: `builtin` / `community` / `custom` sources take the HSL offset; `wallpaper` does not; and the offset produces an unusable colour whenever the source accent's saturation is already high.

### Steps to reproduce

1. `noctalia msg color-scheme-set builtin Catppuccin`
2. `noctalia msg templates-apply`
3. Render any template referencing `{{colors.primary_container.default.hex}}` — or just read a generated file that uses the role.

With the `kcolorscheme` template enabled, it lands in `~/.config/kdeglobals` as `[Colors:Window] BackgroundAlternate` and `[WM] activeBackground`. With `niri` enabled, as `tab-indicator inactive-color` in `~/.config/niri/noctalia.kdl`.

### Expected behavior

`primary_container` is a muted, low-emphasis surface tinted toward the accent — usable as a container background behind `on_primary_container` text.

### Actual behavior

On pastel-accent palettes it is a maximally saturated accent. In practice: Dolphin's alternating list rows render neon purple against a dark grey window, and the niri tab indicator's inactive state is brighter than its active state.

### Suggested fix

Scale the saturation step by the headroom available rather than adding a constant — e.g. `S' = S + 0.15 * (1 - S)` — or clamp on the resulting chroma. Either keeps Nord's output essentially unchanged while pulling Catppuccin's back into container territory.

A local workaround, for anyone hitting this before a fix: copy `kde/kcolorscheme.colors` and `niri/niri.kdl` out of `/usr/share/noctalia/assets/templates/`, remap `primary_container` to `surface_container_high` (KDE) and `surface_variant` (niri), register them under `[theme.templates.user.*]`, and drop the two ids from `builtin_ids`.

### Noctalia version

`v5.0.0 (v5.0.0-beta.5)`, Fedora package `noctalia-5.0.0~beta.5-1.fc44`

### Compositor

niri 26.04

### Distribution

Fedora 44

### Environment information

Fedora 44 KDE base, niri session (no KWin/Plasma shell), `QT_QPA_PLATFORMTHEME=kde` via `plasma-integration`. Templates enabled: `gtk3`, `gtk4`, `ghostty`, `starship`, `btop`; `kcolorscheme` and `niri` are the two affected ones. Values above were produced by a single-purpose user template so they are read straight from the palette rather than inferred from a rendered config.

endafk:
## What

`makeContainerDark()` derives `*_container` colour roles from their base role with a fixed HSL offset of `S +0.15, L -0.35`, applied with no headroom check. On palettes whose accents already sit high on the saturation scale, the `+0.15` clamps at `S = 1.0` while `L` still drops into the midtones, turning a muted container tint into a fully saturated, near-neon colour on the sRGB gamut wall (e.g. built-in Catppuccin's `primary` `#cba6f7` becomes `primary_container` `#6b02e9`).

This fixes both places that use the pattern:
- `fixed_palette.cpp` (builtin + community palette derivation)
- `custom_schemes.cpp` (custom-scheme derivation)

## Fix

Scale the saturation step by the remaining headroom instead of adding a constant, as suggested in the issue:

```cpp
S' = S + 0.15 * (1 - S)
```

## Verification

I re-derived the issue's numbers from source to confirm both the bug and the fix, rather than taking the formula on faith:

- **Reproduced the bug**: recomputing `S +0.15 / L -0.35` from Catppuccin's actual HSL values lines up with the hex values reported in the issue (`primary` becomes `#6b02e9`, `secondary` clamps at `S=1.00`, etc.), confirming this is the right code path.
- **Confirmed the fix moves saturation the right way on the reported case**: Catppuccin's derived saturations drop meaningfully off the gamut wall. `primary` goes 0.99 to 0.86, `secondary` 1.00 to 0.93, `tertiary` 0.72 to 0.64, instead of pinning at 1.0.
- **Checked the low-saturation control isn't broken**: on Nord (S 0.25 to 0.43, the case the issue says is "fine" today), the new formula shifts saturation by only about 0.04 to 0.065. That's a small, real change, not a no-op, but nowhere near enough to be visually significant. Flagging this precisely rather than claiming byte-identical output.

`source = "wallpaper"` is unaffected either way; it goes through Material Color Utilities / HCT gamut mapping rather than this HSL offset.

Compiles clean in isolation; couldn't do a full `meson` build (missing `wayland-protocols` on my machine). Formatted with `clang-format` per `just format`.

Fixes #3924

### DIFF
--- a/src/theme/custom_schemes.cpp
+++ b/src/theme/custom_schemes.cpp
@@ -621,7 +621,10 @@ namespace noctalia::theme {
 
       auto makeContainerDark = [](const Color& base) {
         auto [h, s, l] = base.toHsl();
-        return fromHsl(h, std::min(s + 0.15, 1.0), std::max(l - 0.35, 0.15));
+        // Scale the saturation step by the remaining headroom instead of adding a
+        // constant: a flat +0.15 clamps at S = 1.0 for already-saturated accents,
+        // producing a maximally saturated "container" instead of a muted one.
+        return fromHsl(h, s + 0.15 * (1.0 - s), std::max(l - 0.35, 0.15));
       };
       const Color primary_container = makeContainerDark(primary_adjusted);
       const Color secondary_container = makeContainerDark(secondary_adjusted);
--- a/src/theme/fixed_palette.cpp
+++ b/src/theme/fixed_palette.cpp
@@ -231,7 +231,11 @@ namespace noctalia::theme {
 
     auto makeContainerDark = [](const Color& base) {
       auto [h, s, l] = base.toHsl();
-      return Color::fromHsl(h, std::min(s + 0.15, 1.0), std::max(l - 0.35, 0.15));
+      // Scale the saturation step by the remaining headroom instead of adding a
+      // constant: on already-saturated accents (e.g. pastel palettes with S near
+      // 1.0) a flat +0.15 clamps at the gamut wall while L still drops into the
+      // midtones, producing a neon "container" instead of a muted one.
+      return Color::fromHsl(h, s + 0.15 * (1.0 - s), std::max(l - 0.35, 0.15));
     };
     auto makeContainerLight = [](const Color& base) {
       auto [h, s, l] = base.toHsl();
--- a/tests/scheme_test.cpp
+++ b/tests/scheme_test.cpp
@@ -1,3 +1,4 @@
+#include "theme/builtin_palettes.h"
 #include "theme/color.h"
 #include "theme/contrast.h"
 #include "theme/palette_generator.h"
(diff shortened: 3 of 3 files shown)


ItsLemmy:
Thanks, I added an extra test to ensure this does not break.

PR #3937, opened 2026-08-14
feat(translations): add "focus existing window" option to multiple la…

### LINKED ISSUE #3859: [FEATURE] Tray: focus the app's already-mapped window on left click (v5.0.0-beta.7 repro; SNI Activate is a no-op for Slack)
### Submission checklist

### Feature type

New functionality

### Feature summary

On a tray left click, when the application already has a mapped toplevel, focus that window through the compositor instead of relying solely on the StatusNotifierItem `Activate` call.

This was previously reported as #3490 and closed with "We dont accept V4 issues anymore. We recommend everyone to upgrade to v5." The report did target v5 as well, but to remove any doubt this issue is a fresh reproduction made entirely on **v5.0.0-beta.7**, with the v5 code path and the v5 API names.

### Motivation / use case

Environment: Noctalia 5.0.0_beta.7-1 (Arch, `noctalia` package), niri 26.04 (8ed0da4), two outputs DP-1 / DP-2, Slack 4.49.89 (`slack-desktop-wayland`, native Wayland via `--ozone-platform=wayland`).

Reproduction, with `niri msg event-stream` recording:

1. Slack has a mapped toplevel: `id=44 ws=2` on DP-1.
2. Focus is on a different output: `niri msg action focus-window --id 3` (Chrome, ws=1, DP-2).
3. Invoke the tray item's method directly, bypassing the shell so the click handling itself cannot be blamed:
   `busctl --user call :1.67 /StatusNotifierItem org.kde.StatusNotifierItem Activate ii 100 100`
4. Result: the D-Bus call returns successfully, **the niri event stream records zero events**, Slack stays `id=44 ws=2` unfocused, focus stays on Chrome.

The click is dead. Slack's `Activate` handler only shows a window that is hidden; with a window already mapped it does nothing, and no xdg-activation token is ever involved, so there is nothing the compositor could honor.

The complementary case works: with the window closed to tray, `Activate` makes Slack map a fresh toplevel, which lands on the active workspace and takes focus normally. So the dead click happens exactly in the case where the user most expects "click tray icon, take me to the app" — and it is invisible when the window happens to be on the currently viewed output, which makes it read as a random failure.

### Proposed solution

In the tray left-click handler — v5 `src/shell/bar/widgets/tray_widget.cpp:785`, currently:

```cpp
(void)m_tray->activateItem(itemId, x, y);
```

1. Try to match the tray item to an open toplevel. `CompositorPlatform::windowsForApp(const std::string& idLower, const std::string& wmClassLower, wl_output* outputFilter = nullptr)` (`src/compositors/compositor_platform.h:105`) already provides the lookup, and SNI ids match app-ids under case-insensitive containment (Slack's SNI id `Slack_status_icon_1` vs app-id `slack`). Where that heuristic is not enough, the item's owner PID is available from `GetConnectionUnixProcessID` on its bus name and can be matched against the toplevel's pid.
2. On a hit, and when the window is not already focused: `CompositorPlatform::activateToplevelInfo(const ToplevelInfo&)` (`compositor_platform.h:115`). Compositor-side focus needs no activation token and brings the workspace/output switch with it — verified here: `niri msg action focus-window --id 44` moves focus DP-2 to DP-1 correctly, which is the same operation the taskbar widget already performs.
3. Otherwise fall back to the current `activateItem()` call, preserving open-from-tray and each app's own toggle-to-hide semantics.

Happy to submit a PR for this if the approach is acceptable.

### Alternatives considered

- Relying on apps to implement xdg-activation when handling SNI `Activate`: outside Noctalia's control, and Electron's tray path has behaved this way for years. Slack in particular is closed source with no public tracker.
- Using the taskbar widget instead, with `show_all_outputs = true` and `only_active_workspace = false`: this does work, because the taskbar focuses through the compositor. It is a workaround, not a fix — it forces a different widget and a much longer taskbar on users who want a tray.
- Locally patching the tray widget: lost on every package update.

### Additional context

Compositor: niri 26.04. All behavior above was verified with direct `busctl` calls to the SNI `Activate` method and with `niri msg event-stream`, so shell-side click handling is ruled out as the cause.

Orfeo42:
## Summary

Left-clicking a tray icon now raises the application's already-open window instead of relying solely on the StatusNotifierItem `Activate()` call.

- New tray widget option `focus_existing_window` (default `true`, label "Focus Existing Window").
- On left click the widget resolves the item to a toplevel, and:
  - a matching window that is not focused is activated through `CompositorPlatform::activateToplevelInfo()`, and `Activate()` is not sent;
  - a matching window that is already focused is a deliberate no-op (no `Activate()` fallthrough, so apps that toggle visibility on `Activate` do not hide the window the user just clicked towards);
  - no matching window means the previous behavior, `TrayService::activateItem()`.
- The SNI `ItemIsMenu` property is now read into `TrayItemInfo`. Items that declare themselves menu-only never receive `Activate()`; a left click opens their menu through the same path right click already uses. This rule applies regardless of the new option's value.
- Item-to-window matching lives in a new header `src/shell/tray/tray_window_match.h` behind a `WindowLookup` seam so the decision logic is unit-testable without a live compositor. The click decision is a pure function, `tray::decideTrayClick()`.
- `tray::windowMatchCandidates()` is added next to the existing `pinMatchCandidates()` in `src/shell/tray/tray_identifier.h`. It reuses the same variant expansion but orders identity-strong fields first (`Id`, bus name, process name, item name, object path, icon names) and descriptive fields last (`Title`, tooltip title/description), then lowercases and deduplicates. Descriptive fields collide with unrelated applications' windows far more easily, and compositor lookups compare against a lowercased app id verbatim. `pinMatchCandidates()` is untouched, so pinning behavior is unchanged.
- `canActivateWindow()` and `newestActivatableWindow()` moved from the anonymous namespace in `src/shell/dock/dock.cpp` to a shared `src/shell/common/window_activation.h` (namespace `shell`) so the tray and the dock apply the same activatability rule. `matchesActiveWindow()` and `nextActivatableWindow()` stay in `dock.cpp` — they encode dock-specific cycling that the tray does not need.

## Motivation

On Wayland an application usually cannot raise itself. When a tray icon is clicked, the SNI `Activate()` call reaches the app, but the app's own attempt to focus its window has no activation token behind it, so nothing visible happens. The common expectation — click the tray icon, get the window — silently fails for a large set of applications.

Doing the activation shell-side fixes this: the shell holds the compositor connection and can activate the toplevel directly. `Activate()` remains the fallback for the case where no window exists yet (the icon is the app's only entry point and the app must create the window itself).

## Type of Change

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Refactoring
- [ ] Build / packaging

Refactoring covers the `window_activation.h` extraction out of `dock.cpp`; the dock's behavior is unchanged.

## Related Issue

Closes #3859

## Testing

Commands run from the repository root:

- `just format` (clang-format v22)
- `meson setup --reconfigure build-debug`
- `ninja -C build-debug noctalia` — links clean, no new warnings
- `meson test -C build-debug` — 81 passed, 0 failed
- `just lint` — clang-tidy with `-warnings-as-errors=*` over 561 files, clean

New unit test `tests/tray_window_match_test.cpp` (registered in `meson.build`) covers:

- no candidate matches any window;
- newest window wins, asserted through explicit `ToplevelInfo::order` values rather than list position;
- windows that cannot be activated are rejected;
- transient unique bus names produce no candidates and trigger no compositor lookup;
- focus detection by handle and by identifier, plus the "no active toplevel" case;
- candidate ordering: an item whose process name matches one window and whose title matches a different window resolves to the process-name window;
- the click decision for every action — activate, focus, already-focused no-op, and menu-only.

## Manual Coverage

- [x] Tested on Niri
- [x] Tested on Hyprland
- [ ] Tested on Sway

Manual verification on a live session is still outstanding; see Additional Notes.

## Screenshots / Videos

No visual change. The only UI addition is a toggle in the tray widget settings, rendered by the existing widget settings generator.

## Checklist

## Additional Notes

- Default value: `focus_existing_window` defaults to `true`, which changes left-click behavior for existing users on upgrade. The reasoning is that the broken case is the common one — an `Activate()` that the compositor renders inert — and users hitting it would have to discover a setting to get working behavior. Happy to flip the default to opt-in if maintainers prefer no behavior change on upgrade; it is a one-line change.
- Menu-only items: honoring `ItemIsMenu` on left click also corrects a pre-existing deviation from the SNI specification, where `Activate()` was sent to items that declare they only support their menu.
- Matching limits: an application whose tray item advertises neither its app id nor a process name matching its toplevel `app_id` will not resolve to a window and falls back to `Activate()`, exactly as before. Matching is exact on normalized identity keys, not substring based, to avoid focusing the wrong application.
- Window selection: the newest activatable window is used. There is no cycling between multiple windows of the same application, unlike the dock — the tray icon is a single-target affordance and cycling would need its own UX decision.
- Manual matrix: still to be run on a live session; I will fill in the Manual Coverage boxes before marking this ready for review.

### DIFF
--- a/assets/translations/en.json
+++ b/assets/translations/en.json
@@ -3468,6 +3468,10 @@
           "description": "Allow mouse wheel and trackpad scroll actions on this widget",
           "label": "Enable Scroll"
         },
+        "focus-existing-window": {
+          "description": "Left-click focuses this app's already-open window instead of only sending the tray Activate action, which many apps ignore once their window is open",
+          "label": "Focus Existing Window"
+        },
         "focused-color": {
           "description": "Color role used for the focused workspace; fixed hex colors are also supported",
           "label": "Focused Color"
--- a/meson.build
+++ b/meson.build
@@ -1199,6 +1199,7 @@ if build_tests
     'taskbar_widget',
     'time_format',
     'tray_identifier',
+    'tray_window_match',
     'triad_workspace_backend',
     'ui_tree_reconciler',
     'upower_charge_limit',
@@ -1208,6 +1209,7 @@ if build_tests
     'widget_action',
     'widget_definition',
     'wallpaper_shuffle_state',
+    'window_activation',
     'workspace_alert_service',
   ]
 
--- a/src/app/application_ui.cpp
+++ b/src/app/application_ui.cpp
@@ -694,7 +694,9 @@ void Application::initPanelManagerAndPanels() {
           &m_wayland, &m_configService, &m_thumbnailService, &m_wallpaperScanner, &m_themeService
       )
   );
-  m_panelManager.registerPanel("tray-drawer", std::make_unique<TrayDrawerPanel>(m_trayService.get(), &m_configService));
+  m_panelManager.registerPanel(
+      "tray-drawer", std::make_unique<TrayDrawerPanel>(&m_compositorPlatform, m_trayService.get(), &m_configService)
+  );
   m_panelManager.registerPanel("polkit", std::make_unique<PolkitPanel>(&m_configService, [this]() {
                                  return m_polkitAgent.get();
                                }));
--- a/src/dbus/tray/tray_service.cpp
+++ b/src/dbus/tray/tray_service.cpp
@@ -201,6 +201,20 @@ namespace {
     return fallback;
   }
 
+  bool get_item_property_bool_from(
+      const std::map<std::string, sdbus::Variant>& properties, std::string_view propertyName, bool fallback
+  ) {
+    const auto* value = findProperty(properties, propertyName);
+    if (value == nullptr) {
+      return fallback;
+    }
+    try {
+      return value->get<bool>();
+    } catch (const sdbus::Error&) {
+    }
+    return fallback;
+  }
+
   using IconPixmapTuple = std::tuple<std::int32_t, std::int32_t, std::vector<std::uint8_t>>;
   using IconPixmapStruct = sdbus::Struct<std::int32_t, std::int32_t, std::vector<std::uint8_t>>;
   using StatusNotifierTextTuple = std::tuple<std::string, std::vector<IconPixmapTuple>, std::string, std::string>;
@@ -2050,6 +2064,7 @@ void TrayService::refreshItemMetadata(const std::string& itemId) {
           next.statusNotifierDescription = std::move(statusNotifierDescription);
           next.status = get_item_property_string_from(properties, "Status", cur.status);
           next.needsAttention = (StringUtils::toLower(next.status) == "needsattention");
+          next.itemIsMenu = get_item_property_bool_from(properties, "ItemIsMenu", cur.itemIsMenu);
 
           if (next.itemName == "chrome_status_icon_1" && !next.statusNotifierTitle.empty()) {
             next.itemName = next.itemName + "::" + next.statusNotifierTitle;
--- a/src/dbus/tray/tray_service.cpp
+++ b/src/dbus/tray/tray_service.cpp
@@ -1725,40 +1725,6 @@ void TrayService::requestProcessNameForItem(const std::string& itemId, const std
   }
 }
 
-void TrayService::requestActivateCapabilityForItem(const std::string& itemId) {
-  const auto proxyIt = m_itemProxies.find(itemId);
-  if (proxyIt == m_itemProxies.end()) {
-    return;
-  }
-
-  try {
-    proxyIt->second->callMethodAsync("Introspect")
-        .onInterface("org.freedesktop.DBus.Introspectable")
-        .withTimeout(kItemPropertyTimeout)
-        .uponReplyInvoke([this, itemId](std::optional<sdbus::Error> error, std::string introspectionXml) {
-          if (error.has_value()) {
-            kLog.debug("tray introspect failed id={} err={}", itemId, error->what());
-            return;
-          }
-
-          auto itemIt = m_items.find(itemId);
-          if (itemIt == m_items.end()) {
-            return;
-          }
-
-          const bool hasActivate = sniActivateMethodPresent(introspectionXml);
-          if (itemIt->second.hasActivateMethod == hasActivate) {
-            return;
-          }
-          itemIt->second.hasActivateMethod = hasActivate;
-          kLog.debug("tray item id={} hasActivateMethod={}", itemId, hasActivate);
-          emitChanged();
-        });
-  } catch (const sdbus::Error& e) {
-    kLog.debug("tray introspect dispatch failed id={} err={}", itemId, e.what());
-  }
-}
-
 std::uint32_t TrayService::connectionPidForBusName(const std::string& busName) const {
   if (busName.empty() || m_dbusProxy == nullptr || !looks_like_dbus_name(busName)) {
     return 0;
@@ -1856,7 +1822,6 @@ void TrayService::registerOrRefreshItem(const std::string& busName, const std::s
           itemId, sdbus::createProxy(m_bus.connection(), sdbus::ServiceName{busName}, sdbus::ObjectPath{objectPath})
       );
       attachItemProxySignals(itemId, *proxyIt->second);
-      requestActivateCapabilityForItem(itemId);
     }
 
     notifyWatcherItemRegistered(itemId);
@@ -2015,7 +1980,6 @@ void TrayService::resolvePathOnlyItemProxy(const std::string& itemId) {
                     }
 
                     attachItemProxySignals(itemId, *proxyIt->second);
-                    requestActivateCapabilityForItem(itemId);
                     m_pathOnlyResolutionsInFlight.erase(itemId);
                     refreshItemMetadata(itemId);
                     emitChanged();
--- a/src/dbus/tray/tray_service.cpp
+++ b/src/dbus/tray/tray_service.cpp
@@ -836,8 +836,7 @@ bool TrayService::itemUsesDBusMenu(const std::string& itemId) const {
   if (it == m_items.end()) {
     return false;
   }
-  const std::string_view menuPath = it->second.menuObjectPath;
-  return !menuPath.empty() && menuPath != "/NO_DBUSMENU";
+  return trayItemHasDBusMenu(it->second);
 }
 
 namespace {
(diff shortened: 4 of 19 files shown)


ItsLemmy:
### [P1] Required clang-tidy check fails                                                                                                
                                                                                                                                         
 src/shell/bar/widgets/volume_widget.cpp:113                                                                                             
                                                                                                                                         
 ```cpp                                                                                                                                  
isActive = std::any_of(captures.begin(), captures.end(), ...);
 ```                                                                                                                                     
                                                                                                                                         
 The repository’s required scoped lint reports:                                                                                          
                                                                                                                                         
 ```text                                                                                                                                 
[modernize-use-ranges,-warnings-as-errors]
 ```                                                                                                                                     
                                                                                                                                         
 Use:                                                                                                                                    
                                                                                                                                         
 ```cpp                                                                                                                                  
isActive = std::ranges::any_of(captures, [](const auto& capture) {
  return capture.kind == PrivacyCaptureKind::Microphone;
});
 ```                                                                                                                                     
                                                                                                                                         
 ### [P2] Output stream existence does not mean active playback                                                                          
                                                                                                                                         
 src/shell/bar/widgets/volume_widget.cpp:117                                                                                             
                                                                                                                                         
 ```cpp                                                                                                                                  
isActive = !m_audio->state().programOutputs.empty();
 ```                                                                                                                                     
                                                                                                                                         
 programOutputs is the application-volume inventory. PipeWireService::rebuildState() adds every classified Stream/Output/Audio node      
 without considering its pw_node_info::state. Applications can retain idle or suspended streams after playback stops, so                 
 hide_when_inactive = true can leave the output widget visible indefinitely.                                                             
                                                                                                                                         
 PipeWire distinguishes RUNNING, IDLE, and SUSPENDED node states, but NodeData/AudioNode currently discard that state. Either:           
                                                                                                                                         
 - retain node state and count only running output streams; or                                                                           
 - scope this setting to input widgets rather than advertising incorrect output behavior.                                                
                                                                                                                                         
 Reference: PipeWire node states (https://docs.pipewire.org/group__pw__node.html#gac493839c0b0a1d0b935969f2e990a2df).

ItsLemmy:
### [P1] Remove non-English catalog edits — assets/translations/de.json:3423 and equivalent changes across 21 other non-English catalogs

Repository policy requires new translation keys to be added only to assets/translations/en.json; translators or the localization import process own other catalogs. This PR adds apparently generated translations to every catalog and explicitly leaves the corresponding checklist  
item unchecked. Keep the English entry and revert all other translation-file changes.                                                          

### [P2] “Newest” selection relies on vector position, not ToplevelInfo::order — src/shell/common/window_activation.h:16-24                        
newestActivatableWindow()returns the last activatable element and never examines window.order. The new test assigns orders 1 and 2, but also  supplies the windows in that same order, so it does not verify the claimed contract. 

This is observable on Hyprland:                           
CompositorPlatform::windowsForApp() appends ext-only windows after wlr windows (src/compositors/compositor_platform.cpp:219-237), regardless of their relative creation order. A tray click can therefore select an older ext-only window instead of the newest application window. Either  select by a reliable ordering field or explicitly establish and test an ordered lookup contract, including the mixed Hyprland result.

ItsLemmy:
I'm being extra cautious with this one, since the tray is fragile beast.

src/shell/bar/widgets/tray_widget.cpp
  ● 801 [correctness] The new click-decision switch falls through to `m_itemActivated()` for every outcome including `OpenMenu`, so opening a tray item's menu from the drawer immediately destroys the panel that is the menu popup's parent.
  ● 788 [correctness] With `focus_existing_window` defaulting to `true`, `Activate()` is never delivered whenever any toplevel matches, breaking tray icons whose show/hide toggle is driven by `Activate`.
src/shell/tray/tray_window_match.h
  ●  60 [correctness] `decideTrayClick` returns `OpenMenu` for any item advertising `ItemIsMenu`, without verifying it actually exports a DBusMenu, turning left-click into a silent no-op for items that don't implement SNI `ContextMenu()`.
  ●  20 [correctness] `findNewestWindowForTrayItem` matches via `windowsForApp` → `app_identity::matchesLower` with empty desktop-entry/StartupWMClass hints, which is exact after normalization and does no app-id tail matching, so reverse-DNS app_ids never match the SNI item id.
  ●  42 [correctness] `trayWindowIsFocused` compares `identifier` strings that are neither unique nor consistent across protocols, producing both false positives and false negatives.
src/shell/common/window_activation.h
  ●  43 [correctness] `newestActivatableWindow` is not a pure move of the dock helper: it ranks by `order` within the non-ext group and only falls back to ext-only, whereas the original returned the last activatable element of the vector.

Summary of PR 3937 review (tray click → focus existing window):

The core issue is finding (1) — the new decision switch in tray_widget.cpp falls through to m_itemActivated() on every branch, so OpenMenu from inside the tray drawer tears down the panel that parents the menu popup. That one is a straightforward bug and should be fixed before merge.

Findings (2) and (3) are about the match/decision logic being too narrow: OpenMenu is picked from the ItemIsMenu hint alone (no DBusMenu check), and window matching skips desktop-entry tail matching so reverse-DNS app_ids (KeePassXC, Nextcloud, Telegram) silently never match — which is a good chunk of the apps the feature targets.

Findings (4) and (6) are decisions rather than defects: the extracted newestActivatableWindow quietly changes the dock's Hyprland launch-focus pick despite the PR claiming dock behavior is unchanged, and focus_existing_window defaulting to true removes the hide half of Activate-driven toggles for existing users.

Orfeo42:
> I'm being extra cautious with this one, since the tray is fragile beast.
> 
> src/shell/bar/widgets/tray_widget.cpp ● 801 [correctness] The new click-decision switch falls through to `m_itemActivated()` for every outcome including `OpenMenu`, so opening a tray item's menu from the drawer immediately destroys the panel that is the menu popup's parent. ● 788 [correctness] With `focus_existing_window` defaulting to `true`, `Activate()` is never delivered whenever any toplevel matches, breaking tray icons whose show/hide toggle is driven by `Activate`. src/shell/tray/tray_window_match.h ● 60 [correctness] `decideTrayClick` returns `OpenMenu` for any item advertising `ItemIsMenu`, without verifying it actually exports a DBusMenu, turning left-click into a silent no-op for items that don't implement SNI `ContextMenu()`. ● 20 [correctness] `findNewestWindowForTrayItem` matches via `windowsForApp` → `app_identity::matchesLower` with empty desktop-entry/StartupWMClass hints, which is exact after normalization and does no app-id tail matching, so reverse-DNS app_ids never match the SNI item id. ● 42 [correctness] `trayWindowIsFocused` compares `identifier` strings that are neither unique nor consistent across protocols, producing both false positives and false negatives. src/shell/common/window_activation.h ● 43 [correctness] `newestActivatableWindow` is not a pure move of the dock helper: it ranks by `order` within the non-ext group and only falls back to ext-only, whereas the original returned the last activatable element of the vector.
> 
> Summary of PR 3937 review (tray click → focus existing window):
> 
> The core issue is finding (1) — the new decision switch in tray_widget.cpp falls through to m_itemActivated() on every branch, so OpenMenu from inside the tray drawer tears down the panel that parents the menu popup. That one is a straightforward bug and should be fixed before merge.
> 
> Findings (2) and (3) are about the match/decision logic being too narrow: OpenMenu is picked from the ItemIsMenu hint alone (no DBusMenu check), and window matching skips desktop-entry tail matching so reverse-DNS app_ids (KeePassXC, Nextcloud, Telegram) silently never match — which is a good chunk of the apps the feature targets.
> 
> Findings (4) and (6) are decisions rather than defects: the extracted newestActivatableWindow quietly changes the dock's Hyprland launch-focus pick despite the PR claiming dock behavior is unchanged, and focus_existing_window defaulting to true removes the hide half of Activate-driven toggles for existing users.

Thanks everybody help me understand all the point i'm missing, you are very kind :heart: 

Next week i'm going to work on this points and try to make it right!

craciu25yt:
is this supposed to be in prod?

ItsLemmy:
when a PR is not merged it's not in the code base.

PR #3944, opened 2026-08-14
widget: fix effects profile glyphs

Ape:
The audio widget effect profile glyphs setting didn't do anything before this fix. The regression was introduced in commit 03071fce9.

The declarative widget resolver only looked for values in WidgetConfig::settings, while nested TOML string maps such as effects_profile_glyphs are stored in WidgetConfig::tables. Resolve string maps from tables while retaining the settings fallback for programmatic callers.

### DIFF
--- a/src/shell/bar/widget_definition.h
+++ b/src/shell/bar/widget_definition.h
@@ -387,6 +387,20 @@ namespace noctalia::bar {
       }
     }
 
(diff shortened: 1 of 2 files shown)


ItsLemmy:
Cool, sorry for the regression. I dont really use easyeffects often so I did not notice.

PR #3945, opened 2026-08-14
refactor: make widget factory dispatch data-driven

Ape:
Replace the built-in widget type condition chain with a static creator table and ranges-based lookup. Keep plugin widgets as the dynamic fallback while preserving widget-specific construction requirements.


ItsLemmy:
Thanks

PR #3946, opened 2026-08-15
feat(input): add readline-style shorcuts

xirzo:
## Summary

Added a subset of [readline-style](https://en.wikipedia.org/wiki/GNU_Readline#Emacs_keyboard_shortcuts) shortcuts to inputs (e.g. Ctrl+W to delete a word, Ctrl+A to move caret to start) and made app launcher use them.

## Motivation

When you make a typo in app launcher, it's usually faster to delete the whole word (or line) and rewrite it rather than moving the caret.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Build / packaging

## Related Issue

None

## Testing

- Added [tests](https://github.com/xirzo/noctalia/blob/210007d0528649e32c8076ef2b88757a76f2af0b/tests/input_readline_shortcuts_test.cpp) for new shortcuts;
- Manually went through each keybinding in app launcher.

## Manual Coverage

- [ ] Tested on Niri
- [x] Tested on Hyprland
- [ ] Tested on Sway

## Screenshots / Videos

<p align="center">
showcase
</p>

## Checklist

## Additional Notes

Docs updated:  noctalia-dev/noctalia-docs#224

All new shortcuts:

| Shortcut | Action |
|---|---|
| `Ctrl+A` | Move caret to start of line |
| `Ctrl+E` | Move caret to end of line |
| `Ctrl+B` / `Ctrl+F` | Move caret back / forward one character |
| `Alt+B` / `Alt+F` | Move caret back / forward one word |
| `Ctrl+W` | Delete previous word |
| `Ctrl+U` | Clear the entire line |
| `Ctrl+K` | Delete from caret to end of line |
| `Alt+D` | Delete next word |

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1154,6 +1154,7 @@ if build_tests
     'image_source_log',
     'input_area',
     'input_dispatcher',
+    'input_readline_shortcuts',
     'input_password_mode',
     'inotify',
     'ipc_service',
--- a/src/shell/launcher/launcher_panel.cpp
+++ b/src/shell/launcher/launcher_panel.cpp
@@ -1070,6 +1070,7 @@ void LauncherPanel::create() {
           .controlHeight = Style::controlHeight * scale,
           .horizontalPadding = Style::spaceMd * scale,
           .clearButtonEnabled = true,
+          .lineEditing = true,
           .surfaceOpacity = panelCardOpacity(),
           .onChange =
               [this](const std::string& text) {
--- a/src/ui/builders.cpp
+++ b/src/ui/builders.cpp
@@ -243,6 +243,9 @@ namespace ui {
     if (props.passwordMode.has_value()) {
       control->setPasswordMode(*props.passwordMode);
     }
+    if (props.lineEditing.has_value()) {
+      control->setLineEditingEnabled(*props.lineEditing);
+    }
     if (props.invalid.has_value()) {
       control->setInvalid(*props.invalid);
     }
--- a/src/ui/builders.h
+++ b/src/ui/builders.h
@@ -146,6 +146,7 @@ namespace ui {
     std::optional<float> horizontalPadding = std::nullopt;
     std::optional<bool> clearButtonEnabled = std::nullopt;
     std::optional<bool> passwordMode = std::nullopt;
+    std::optional<bool> lineEditing = std::nullopt;
     std::optional<bool> invalid = std::nullopt;
     std::optional<bool> frameVisible = std::nullopt;
     std::optional<bool> embeddedOnSolidPrimary = std::nullopt;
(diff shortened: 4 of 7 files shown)


ItsLemmy:
### [P2] Alt+F implements the wrong Readline word boundary                                                                                            
                                                                           
src/ui/controls/input.cpp:1228-1231 uses:                                                                                                             
                                                                                                                                               
```cpp                                                                                                                                                
m_cursorPos = nextWordStartForByteOffset(m_cursorPos);
```                                                                                                                                                   
                                                                                                              
GNU Readline’s forward-word (M-f) moves to the end of the next word, not the start of the following word. From the beginning of "foo bar", Alt+F  should stop after "foo"; typing x should produce "foox bar". The implementation instead stops before "bar", producing "foo xbar".                     
                                                                                                                                               
The new test currently codifies the incorrect behavior. Use nextWordEndForByteOffset() and update that expectation.                                   
                                                                                                                                               
Reference: GNU Bash manual, Commands For Moving (https://www.gnu.org/software/bash/manual/html_node/Commands-For-Moving.html)

ItsLemmy:
Thanks

BigBoyBarney:
Why would you change the keybinds just for the launcher without an option to disable?

> [...] and made app launcher use them.

Why was this even merged like this lol. Absolutely horrendous feature. It changed the default behaviour and deviates from the rest of Noctalia. No ordinary person would expect GNU keybinds. Additionally, these keybinds make 0 sense for the launcher. It has no multiline support and usually people search for a couple of keys at most to find their app. This leaves most of these keybinds without purpose. Seems very arbitrary to me. Not once did I think "I do wish I could go to the beginning of the line in the launcher with ctrl + A!"

 See #4530 as well.

Given that https://github.com/noctalia-dev/noctalia/pull/4137 is not merged yet, this behaviour can't be disabled either, which is super weird. It's the only text box in Noctalia that follows these weird key bindings.

Would it be possible to (if not revert this default) add an option to disable this monstrosity? Like merging #4137.

xirzo:
> No ordinary person would expect GNU keybinds.

safari searchbar, zsh, bash, python repl, node repl, etc. do use such keybindings so i think it's safe to say that a lot of people are used to readline-style keybindings.

> It has no multiline support and usually people search for a couple of keys at most to find their app.

app-launcher can do more than just launching apps, for instance doing math with /calc, in which case those keybindings help with navigation, more than that readline-style keybindings do not require multiline support.

although I do agree that making this optional would be nice and makes absolute sense (#4137)

what i don't appreciate is the way you've delivered the information, you can make a point without devaluation of other people's job

BigBoyBarney:
I could and should have phrased my comment better. I didn't mean to devalue your work, though I do think that making this the default out of the blue is effectively a feature regression. I apologise for being crude.

PR #3949, opened 2026-08-15
fix(network): support WPA3 SAE connections

### LINKED ISSUE #3841: [BUG] Cannot connect to WPA3-Personal/SAE networks because NetworkManager connection is forced to wpa-psk
### Submission checklist

### Bug description

Noctalia cannot create a first-time connection to a WPA3-Personal/SAE-only access point.

The password prompt is displayed normally, but the connection fails because Noctalia asks NetworkManager to use `wpa-psk` for every secured access point.

NetworkManager rejects that configuration because an SAE-only access point does not support PSK key management.

The Noctalia service log reports:

```text
AddAndActivateConnection2 failed ssid=<redacted>
err=[org.freedesktop.NetworkManager.Settings.Connection.InvalidProperty]
802-11-wireless-security.key-mgmt: Access point does not support PSK but setting requires it
 ```

### Steps to reproduce

 1. Use NetworkManager as the network backend.
 2. Ensure there is no saved NetworkManager profile for the target SSID.
 3. Open Noctalia's network panel.
 4. Select a WPA3-Personal/SAE-only access point.
 5. Enter the correct password.
 6. Attempt to connect.

### Expected behavior

Noctalia should detect that the access point advertises SAE and create a connection using:

 ```text
802-11-wireless-security.key-mgmt = sae
 ```

 Alternatively, it should let NetworkManager infer the compatible authentication method.

### Actual behavior

Noctalia creates the connection settings with:

 ```text
802-11-wireless-security.key-mgmt = wpa-psk
 ```

 NetworkManager rejects the settings before association begins.

### Logs / error output

```text

```

### Noctalia version

5.0.0-beta.7

### Compositor

Niri

### Distribution

NixOS

### Environment information

```text
Installation method: Nix
GPU: intel
```

### Comparison with nmcli

 Connecting to the same access point with the same password succeeds:

 ```console
nmcli device wifi connect '<ssid>' password '<password>'
 ```

 The resulting NetworkManager profile contains:

 ```text
802-11-wireless-security.key-mgmt: sae
 ```

 This confirms that the access point, password, Wi-Fi hardware, driver, supplicant, and NetworkManager SAE support are working.

Relevant code:

 https://github.com/noctalia-dev/noctalia/blob/e9230c81beb2513e4a4b09884f37e4f22fc246de/src/dbus/network/network_manager_service.cpp

 ### Root cause

 Noctalia reads the access point's WpaFlags and RsnFlags, but currently reduces them to a single secured Boolean.

 Later, addAndActivateAccessPoint() treats every secured access point as WPA-PSK:

 ```cpp
settings["802-11-wireless-security"]["key-mgmt"] =
    sdbus::Variant{std::string("wpa-psk")};
 ```

 As a result, the distinction between WPA2-Personal/PSK and WPA3-Personal/SAE is lost.

 ### Suggested fix

 Preserve whether the access point advertises the NetworkManager SAE security flag:

 ```cpp
constexpr std::uint32_t kNm80211ApSecKeyMgmtSae = 0x00000400;

info.supportsSae =
    (rsnFlags & kNm80211ApSecKeyMgmtSae) != 0U;
 ```

 Then select compatible key management when creating the connection:

 ```cpp
const auto keyMgmt = ap.supportsSae ? "sae" : "wpa-psk";

settings["802-11-wireless-security"]["key-mgmt"] =
    sdbus::Variant{std::string(keyMgmt)};
 ```

 A transition-mode WPA2/WPA3 access point advertising SAE can prefer SAE.

 It may be even better to delegate more of the security selection to NetworkManager, if its D-Bus API supports doing so while retaining Noctalia's password
 prompt and secret-agent flow.

 ### Verification

 A local build with the SAE capability preserved and key-mgmt=sae selected was tested as follows:

 1. Remove the previously saved profile.
 2. Select the WPA3-only network through Noctalia.
 3. Enter the password through Noctalia.
 4. Observe successful AddAndActivateConnection2.
 5. Confirm that Noctalia persists the connection.
 6. Confirm that the resulting profile has key-mgmt=sae.
 7. Confirm working IPv4 and IPv6 connectivity.

 WPA2-Personal connections continue to use wpa-psk.

 ### Scope

 This report concerns WPA3-Personal/SAE.

 It does not cover WPA-Enterprise/802.1X configuration, which is separately tracked in:

 https://github.com/noctalia-dev/noctalia/issues/3216

g1ibby:
## Summary

- Preserve NetworkManager's SAE capability from access-point discovery through activation.
- Select `sae` for SAE-capable access points and retain `wpa-psk` for WPA2-Personal access points.
- Prefer SAE for WPA2/WPA3 transition-mode access points.
- Preserve the selected BSSID's SAE capability during SSID deduplication.
- Refresh cached control-center network rows when the advertised SAE capability changes.

## Motivation

First-time connections to WPA3-Personal/SAE-only access points failed because Noctalia reduced NetworkManager's WPA/RSN capability flags to a generic `secured` value and always created secured connections with:

```text
802-11-wireless-security.key-mgmt = wpa-psk
 ```

 SAE-only access points require:

 ```text
802-11-wireless-security.key-mgmt = sae
 ```
This change preserves the SAE capability of the selected NetworkManager access-point object through discovery, deduplication, password prompting, and activation. WPA2 access points continue to use WPA-PSK.

 The network panel's cached row identity now also includes SAE capability. If an existing SSID's advertised capability changes while the panel remains open, the row is rebuilt with the current AccessPointInfo instead of activating with stale key management.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Build / packaging

## Related Issue

 Fixes #3841

## Testing

Ran the focused regression in the Nix development environment:

 ```sh
nix develop --command just test debug network_manager_security --print-errorlogs
 ```

 Result:

 ```text
network_manager_security: 1/1 passed
 ```

## Manual Coverage

- [x] Tested on Niri
- [ ] Tested on Hyprland
- [ ] Tested on Sway

## Checklist

## Additional Notes

The NetworkManager SAE bit and key-management mapping are kept together in a small allocation-free helper. This gives production code and regression tests one source of truth without requiring D-Bus or a running NetworkManager instance in the focused test.
Other network backends retain their existing behavior because AccessPointInfo::supportsSae defaults to false.
When multiple BSSIDs advertise the same SSID, Noctalia retains the SAE capability of the selected representative access-point object rather than combining capabilities across BSSIDs. This matches the access-point object passed to AddAndActivateConnection2.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1165,6 +1165,7 @@ if build_tests
     'location_service',
     'math_provider',
     'monitor_selector',
+    'network_manager_security',
     'notification_dnd',
     'notification_filter',
     'notification_history_utf8',
--- a/src/dbus/network/network_manager_security.h
+++ b/src/dbus/network/network_manager_security.h
@@ -1,0 +1,18 @@
+#pragma once
+
+#include <cstdint>
+#include <string_view>
+
+namespace network_manager_security {
+
+  inline constexpr std::uint32_t kNm80211ApSecKeyMgmtSae = 0x00000400U;
+
+  [[nodiscard]] constexpr bool supportsSae(std::uint32_t rsnFlags) noexcept {
+    return (rsnFlags & kNm80211ApSecKeyMgmtSae) != 0U;
+  }
+
+  [[nodiscard]] constexpr std::string_view keyManagement(bool supportsSae) noexcept {
+    return supportsSae ? "sae" : "wpa-psk";
+  }
+
+} // namespace network_manager_security
--- a/src/dbus/network/network_manager_service.cpp
+++ b/src/dbus/network/network_manager_service.cpp
@@ -1,6 +1,7 @@
 #include "dbus/network/network_manager_service.h"
 
 #include "core/log.h"
+#include "dbus/network/network_manager_security.h"
 #include "dbus/system_bus.h"
 #include "system/rfkill_helper.h"
 
@@ -389,7 +390,8 @@ bool NetworkManagerService::addAndActivateAccessPoint(
   ConnectionSettings settings;
   if (ap.secured) {
     // Minimal secured-wifi settings — NM fills in ssid from the specific_object.
-    settings["802-11-wireless-security"]["key-mgmt"] = sdbus::Variant{std::string("wpa-psk")};
+    settings["802-11-wireless-security"]["key-mgmt"] =
+        sdbus::Variant{std::string(network_manager_security::keyManagement(ap.supportsSae))};
     if (psk.has_value()) {
       settings["802-11-wireless-security"]["psk"] = sdbus::Variant{*psk};
     }
@@ -1693,6 +1695,7 @@ void NetworkManagerService::refreshAccessPoints(std::function<void()> onComplete
                                           }();
                                           info.secured =
                                               (wpaFlags != k_nm80211ApSecNone) || (rsnFlags != k_nm80211ApSecNone);
+                                          info.supportsSae = network_manager_security::supportsSae(rsnFlags);
                                           if (!info.ssid.empty()) {
                                             apState->aps.push_back(std::move(info));
                                           }
@@ -1774,6 +1777,7 @@ void NetworkManagerService::finishRefreshAccessPoints(
       it->path = ap.path;
       it->devicePath = ap.devicePath;
       it->secured = ap.secured;
+      it->supportsSae = ap.supportsSae;
     }
   }
   std::ranges::sort(deduped, [](const AccessPointInfo& a, const AccessPointInfo& b) {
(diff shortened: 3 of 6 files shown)


ItsLemmy:
##  [P2] Include supportsSae in the control-center row identity

src/shell/control_center/tabs/network_tab.cpp:817-827                           

The PR body says cached rows are rebuilt when SAE capability changes, but NetworkTab::structureKey() still keys each row only on SSID, secured, active, and saved status. AccessPointRow stores its own AccessPointInfo, while syncLiveMetrics() updates only signal strength.

Therefore, if a refresh changes the selected BSSID from WPA2 to SAE—or the same BSSID changes from transition mode to SAE-only, without changing those existing key fields, the row is retained with stale supportsSae. Clicking it then calls activateAccessPoint() with the old value and can still submit key-mgmt = wpa-psk, reproducing the failure this PR addresses. 
      
Add ap.supportsSae to structureKey(), as the PR description already intends. The focused helper test does not exercise this discovery → row  cache → activation path, which is why the omission passes.

g1ibby:
@ItsLemmy You're right. I had already added `ap.supportsSae` to `NetworkTab::structureKey()` in a local follow-up commit, forgot to push it. That's why the PR description and remote diff were out of sync. I pushed commit

ItsLemmy:
Thanks!

PR #3993, opened 2026-08-17
fix(lockscreen): preserve fcitx context after unlock

takumiando:
## Summary

Stop registering the session lock surface with `text-input-v3`.

This prevents the lock screen from leaving Chromium's fcitx input context inactive after unlocking. Password input continues to be handled through `wl_keyboard`.

## Motivation

When unlocking while Chrome is focused, fcitx may stop accepting input in Chrome. Focusing another application such as Ghostty restores fcitx, after which it also works
again in Chrome.

The lock screen was creating its own `text-input-v3` context for the password field. Removing that context prevents fcitx from being left in the inactive state when focus
returns to Chrome.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Build / packaging

## Related Issue

N/A

## Testing

- Built successfully with:
  - `nix develop --command just build debug`
- Manually verified on Niri:
  1. Focus Chrome.
  2. Lock and unlock the session.
  3. Confirm that fcitx works immediately in Chrome without focusing another application first.
- Confirmed that password entry and unlocking still work through `wl_keyboard`.

## Manual Coverage

- [x] Tested on Niri
- [ ] Tested on Hyprland
- [ ] Tested on Sway

## Screenshots / Videos

N/A — this change has no visual impact.

## Checklist

## Additional Notes

The lock-screen password field intentionally no longer uses an input method. Direct keyboard input remains available through `wl_keyboard`.

### DIFF
--- a/src/shell/lockscreen/lock_screen.cpp
+++ b/src/shell/lockscreen/lock_screen.cpp
@@ -182,9 +182,6 @@ void LockScreen::unlock() {
   stopFingerprint();
 
   const bool wasLockedInteractive = m_locked;
-  for (auto& instance : m_instances) {
-    instance.surface->setLockedState(false);
-  }
 
   if (m_lock != nullptr) {
     if (m_locked) {
--- a/src/shell/lockscreen/lock_screen.cpp
+++ b/src/shell/lockscreen/lock_screen.cpp
@@ -182,6 +182,9 @@ void LockScreen::unlock() {
   stopFingerprint();
 
   const bool wasLockedInteractive = m_locked;
+  for (auto& instance : m_instances) {
+    instance.surface->setLockedState(false);
+  }
 
   if (m_lock != nullptr) {
     if (m_locked) {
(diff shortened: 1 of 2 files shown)


ItsLemmy:
I would really like to avoid disabling IME on the lockscreen,  what if someone has a IME composed password?   Could you check with my commits if that helps?

takumiando:
I'm a Japanese speaker and use fcitx5 with Mozc, but I'd never heard of or considered anyone creating a password using an IME.
That said, you're right. it’s not entirely impossible.

takumiando:
Personally, I’d actually prefer the IME to be forcibly disabled on the lock screen.
Would you consider adding an option for that?

takumiando:
I tested `d1a25e5` on Niri with fcitx5 5.1.21 and a Chromium-based
browser, but unfortunately the issue still occurs.

I also tried destroying the `text-input-v3` object before unlocking,
waiting for a `wl_display_roundtrip`, and recreating it only after the
unlock. The issue still occurred in that case as well.

So far, the only approach that fixes the issue is not registering the
lock surface with `text-input-v3` in the first place.

Would an option such as `lockscreen.ime_enabled`, defaulting to true,
be acceptable? Affected users could disable it while IME-composed
passwords would remain supported by default.

ItsLemmy:
You are right, I consulted with other IME users on Discord and they all agreed it was pointless to have IME on password field.  

Thanks for the PR

PR #3995, opened 2026-08-17
feat(plugins): auto-update options All/Official/None

### LINKED ISSUE #3991: [FEATURE] option to auto-update only official plugins
### Submission checklist

### Feature type

Configuration / customization

### Feature summary

rather than a general "auto-update plugins" toggle, there should be a dropdown to auto-update none, official, or all

### Motivation / use case

main reason why one would want to avoid auto-updates of plugins would presumably be the security concerns with malware being pushed in an update. Considering the user is using noctalia, they presumably already trust the official releases.

### Proposed solution

rather than a general "auto-update plugins" toggle, there should be a dropdown to auto-update none, official, or all

### Alternatives considered

_No response_

_No response_

### Additional context

_No response_

AdityaKr015:
## Summary

This PR adds dropdown option in Auto-Update Plugins-  All Sources / Official only/ None.

## Motivation

Users should have feature to choose which plugins they want to auto update.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Build / packaging

## Related Issue

Closes #3991

## Testing

- Ran `just format`,`just build` clean
- Ran`just test` clean
- Tested the debug build and see the dropdown function showing  All sources | Official Only | None

## Manual Coverage

- [x] Tested on Niri
- [ ] Tested on Hyprland
- [ ] Tested on Sway

## Screenshots / Videos

image

## Checklist

- [x] This PR is ready for review.
- [x] I will update end-user documentation after merge.

### DIFF
--- a/assets/translations/en.json
+++ b/assets/translations/en.json
@@ -1376,6 +1376,13 @@
       "panel-position": {
         "auto": "Follow Bar"
       },
+      "plugins": {
+        "auto-update": {
+          "all": "All sources",
+          "none": "None",
+          "official": "Official only"
+        }
+      },
       "screen-position": {
         "bottom-center": "Bottom Center",
         "bottom-left": "Bottom Left",
@@ -1498,7 +1505,7 @@
         "add": "Add Source",
         "add-title": "Add Plugin Source",
         "auto-update": "Auto-Update Plugins",
-        "auto-update-desc": "Check enabled sources for new versions in the background and keep installed plugins current",
+        "auto-update-desc": "Check selected sources for new versions in the background and keep installed plugins current",
         "edit": "Edit source",
         "edit-title": "Edit plugin source",
         "empty": "No sources configured.",
--- a/assets/translations/en.json
+++ b/assets/translations/en.json
@@ -1379,8 +1379,8 @@
       "plugins": {
         "auto-update": {
           "all": "All sources",
-          "none": "None",
-          "official": "Official only"
+          "official": "Official only",
+          "none": "None"
         }
       },
       "screen-position": {
--- a/src/app/application.h
+++ b/src/app/application.h
@@ -188,8 +188,8 @@ private:
   void reloadDmenuProviders();
   // (Re)register plugin-backed panels from the enabled plugin set.
   void reloadPluginPanels();
-  // When [plugins].auto_update is on, pull every git source. Run once at startup and on
-  // a 6h repeating timer so long-lived sessions pick up new plugin versions.
+  // When [plugins].auto_update is on, pull git sources per the configured mode. Run once at
+  // startup and on a 6h repeating timer so long-lived sessions pick up new plugin versions.
   void runPluginAutoUpdate();
   void startTrayService();
   void syncNotificationDaemon();
--- a/src/app/application_plugins.cpp
+++ b/src/app/application_plugins.cpp
@@ -138,11 +138,7 @@ void Application::reloadPluginPanels() {
 }
 
 void Application::runPluginAutoUpdate() {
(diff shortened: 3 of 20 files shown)


Ly-sec:
Hey @AdityaKr015 , there is one small thing that should be changed.

`plugins.auto_update` should accept only the new string modes:

- `"none"`
- `"official"`
- `"all"`

The proposed boolean compatibility path (`true` → `"all"`, `false` → `"none"`) adds a configuration fallback chain which isn't exactly the way we deal with those things.
Also if you can, updating the `noctalia-docs` repo entry would be nice :)

ItsLemmy:
### P1 — “Official only” can update a non-official source                                          

updateAutoUpdateScope()                                                                         
/src/scripting/plugin_manager.cpp#L930-L943 correctly identifies the official entry by name and location,    
but then discards that identity and calls update(source.name). findSource() 
/src/scripting/plugin_manager.cpp#L303-L309 resolves the first source with that name.                        

Duplicate source names are currently accepted as valid. This configuration passed noctalia config  
validate:                                                                                          

```toml                                                                                            
[plugins]
auto_update = "official"

[[plugins.source]]
name = "official"
kind = "git"
location = "https://example.invalid/untrusted"

[[plugins.source]]
name = "official"
kind = "git"
location = "https://github.com/noctalia-dev/official-plugins"
```                                                                                                

The loop matches the second entry, calls update("official"), then updates the first, untrusted     
entry. That violates the trust boundary this mode exists to provide.                               

Reject duplicate source names during configuration parsing, or preserve/pass the exact matched     
PluginSourceConfig into the update operation. Add coverage for this ordering case.                       

### P1 — Remove boolean compatibility for the canonical config key                                 

As Lysec said, no silent fallback.  And this change is not worth a setting migration
I would also reorder the dropdown to: "All sources", "Official only", "None"
Which a better imo better UX since "All sources" is the default.

ItsLemmy:
### P1 -  “Official only” still can fetch a non-official repository                                 
                                                                                                    
 PluginManager::updateSource()                                                                      
 (https://github.com/noctalia-dev/noctalia/blob/4c715ca4881cc18f731339273e5d1d49b5bc4168/src/script 
 ing/plugin_manager.cpp#L742-L747) now receives the exact matching PluginSourceConfig, but          
 gitRepoRoot(source) does not preserve that identity. sourceStorageRoot()                           
 (https://github.com/noctalia-dev/noctalia/blob/main/src/scripting/plugin_source_paths.cpp#L32-L45) 
 derives storage solely from source.name.                                                           
                                                                                                    
 Therefore, duplicate sources named official share the same checkout:                               
                                                                                                    
 ```toml                                                                                            
[[plugins.source]]
name = "official"
kind = "git"
location = "https://github.com/noctalia-dev/official-plugins"

[[plugins.source]]
name = "official"
kind = "git"
location = "https://example.invalid/untrusted"
 ```                                                                                                
                                                                                                    
 If the shared official/repo checkout was created from the untrusted entry, selecting the           
 legitimate config entry still calls git fetch origin in that untrusted checkout. source.location   
 is never checked against the repository’s remote. The new scope-selection test only verifies which 
 config object was selected; it does not verify which repository is fetched.                        
                                                                                                    
 Reject duplicate source names during schema finalization. That matches the existing invariant that 
 source names are identities and storage/locks/materialized paths are keyed by name. Replace the    
 duplicate-acceptance test with a validation failure test.

ItsLemmy:
Changed my mind and added a setting migration, since it would have popped a critical notification for everyone.
Also fixed a git fetch but that I found along the way.

Thanks again!

AdityaKr015:
Awesome and thank you for the merge again

PR #4000, opened 2026-08-18
refactor(control-center): make shortcut registry data-driven

Ape:
Consolidate built-in shortcut metadata, creation, and availability into a single descriptor catalog.


ItsLemmy:
Thanks

PR #4002, opened 2026-08-18
fix(idle): keep idle tracking alive across screensaver inhibits

### LINKED ISSUE #4190: [BUG] Idle wakes screen back up after lock action
### Submission checklist

### Bug description

I have my idle service configured so it first turns screen off, then locks, then finally suspends. I noticed recently that it starts waking up after the lock, and waiting for the timeout to happen again before proceeding to suspend.

This is different from #3165, because all the actions actually do fire properly. It just causes screen to wake for some reason.

### Steps to reproduce

1. Configure idle service to have a lock action that happens while screen is off
2. Wait for screen to turn off (it does)
3. Wait for lock action to trigger -> the screen wakes up
4. For the suspend to happen, the timeout has to wait starting from 0

Idle service section in my config

```toml
[idle]
behavior_order = ["screen_off", "lock", "suspend"]
pre_action_fade_seconds = 5.0
[idle.behavior.screen-off]
timeout = 300
action = "screen_off"
[idle.behavior.lock]
timeout = 305
action = "lock"
[idle.behavior.suspend]
timeout = 310
action = "command"
# this is just a thing that I use to only sleep on certain computers, based on hostname
command = "~/.config/noctalia/scripts/sleep.sh"
```

### Expected behavior

1. Screen turns off after 300 seconds
2. After 5 seconds, it locks (no visible effect as the screen should be off)
3. After 5 more seconds, the system suspends

### Actual behavior

1. Screen turns off after 300 seconds
2. Locked 5 seconds later, but the screen turns back on
3. Another 300 or so seconds later (I assume 310), suspend triggers

### Logs / error output

```text
00:13:25.802 [INF] [idle] idle behavior 'screen-off' pre-action fade 5000ms
00:13:30.802 [INF] [idle] idle behavior 'lock' joined active pre-action fade
00:13:30.812 [INF] [idle] idle behavior 'screen-off' triggered after pre-action fade
00:13:31.114 [INF] [idle] idle behavior 'lock' triggered after pre-action fade
00:13:31.114 [INF] [session] lock requested: compositor=Niri env_hint="niri:niri:niri" xdg_session_id=2 user=bbaovanc
00:13:31.116 [INF] [lockscreen] session lock requested
00:13:31.116 [INF] [session] lock: lock screen requested
00:13:31.117 [INF] [texcache] hit /home/bbaovanc/wallpapers/evgeny-matveev-f4hR4QdFP04-unsplash.jpg (refCount=3)
00:13:31.146 [INF] [texcache] hit /home/bbaovanc/wallpapers/evgeny-matveev-f4hR4QdFP04-unsplash.jpg (refCount=4)
00:13:31.162 [INF] [fingerprint] no fprintd device available: [org.freedesktop.DBus.Error.ServiceUnknown] The name is not activatable
00:13:31.162 [INF] [lockscreen] session is locked
00:18:31.320 [INF] [idle] idle behavior 'screen-off' pre-action fade 5000ms
00:18:36.320 [INF] [idle] idle behavior 'lock' joined active pre-action fade
00:18:36.438 [INF] [idle] idle behavior 'screen-off' triggered after pre-action fade
00:18:36.685 [INF] [idle] idle behavior 'lock' triggered after pre-action fade
00:18:36.685 [INF] [session] lock requested: compositor=Niri env_hint="niri:niri:niri" xdg_session_id=2 user=bbaovanc
00:18:36.685 [INF] [session] lock: lock screen requested
00:18:41.320 [INF] [idle] idle behavior 'suspend' pre-action fade 5000ms
00:18:46.443 [INF] [idle] idle behavior 'suspend' triggered after pre-action fade
```

### Noctalia version

v5.0.0_beta.10-1

### Compositor

Niri

### Distribution

Arch-based

### Environment information

```text
Installation method: Arch package
```

### Additional context

I bisected and found that 9691c87ddaef4e8234d733d568ecf3376aea2159 introduces this issue.

yevhen-sychov:
## Summary

Two fixes to idle handling, both around the same failure: an idle behavior that has already run must not be silently undone, and a D-Bus screensaver inhibit must not permanently strand idle tracking.

**1. Don't run the resume action when re-arming an idled behavior.**
`recreateBehaviorNotification()` ran a behavior's resume action whenever the behavior had already idled. Both callers hit that on the wrong transition: locking the session re-arms every behavior, so a `screen_off` that had already blanked the display lit it straight back up the moment the lock screen appeared. Those behaviors don't need a synthetic resume — the input that eventually wakes the seat delivers a real `resumed` event and runs it. Re-arming now takes an explicit `resumeIfIdled` flag; only unlocking passes `true`, since that is the transition after which a stuck display-off would never be undone.

**2. Keep tracking compositor idle across screensaver inhibits.**
While an app held a D-Bus screensaver inhibit, the 1s heartbeat notification's `idled` event was dropped on arrival. `ext-idle-notify` cannot resend it without an intervening `resumed`, so the event was gone for good: releasing the inhibit re-armed the behavior notifications but never the heartbeat, and the live idle counter stayed pinned at "Active" until the user touched an input device.

The fix records what the compositor reported regardless of the inhibit and gates only the counter, so the seat's real state is known the moment the inhibit lifts. On release, every *waiting* behavior's countdown restarts and counting resumes.

Two smaller behaviors fall out of that:

- Suppressing an idled behavior no longer destroys and recreates its notification. Browsers renew their inhibit during long playback by dropping it and immediately retaking it (Firefox, roughly once a minute, gaps measured at 370–660 ms), and each of those gaps used to restart every countdown. A 1s debounce absorbs them.
- An inhibit taken during the pre-action fade now cancels the fade instead of blanking the screen over freshly started playback.

The screensaver debug logs now carry the resulting lock count, since a leaked lock and a compositor that stopped reporting idle look identical from the outside.

## Motivation

On Hyprland, browsers inhibit over D-Bus rather than through the compositor's `idle-inhibit-unstable-v1`. Pausing or finishing a video therefore left idle tracking wedged indefinitely — the settings panel read "Active" forever and idle behaviors never fired again until an input event arrived. Separately, locking the session while the display was already off turned it back on.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Build / packaging

## Related Issue

Closes #4190 — the wake-after-lock reported there is exactly the resume-on-re-arm path that
this PR's first commit fixes. The diagnosis in that issue's comment (`recreateBehaviorNotification`
running `runResumeBehavior` when `phase == Idled`) matches the change here.

Possibly also #4077 (screen flashes black instead of dimming) — the same re-arm path is a
plausible cause, but I have not reproduced it, so I am not marking it fixed.

## Testing

- `meson compile` — clean, no new warnings.
- `meson test` — passes, aside from three failures (`location_service`, `plugin_i18n`, `time_format`) that reproduce identically on an unmodified checkout in the same out-of-tree build dir; they are an asset-path artifact of my build layout, not related to this change.
- `clang-format` v22.1.8 clean on all three touched files.
- Manual: extended normal desktop use — video playback in Firefox (start, pause, tab-switch, close), long unattended idles, lock/unlock with `screen_off` already applied, and suspend after playback ends.

A side note on the mechanism, since it is not guaranteed by the protocol text: destroying and recreating an `ext_idle_notification_v1` restarts the countdown from zero rather than immediately re-reporting accrued seat idleness. I verified that empirically against Hyprland with a minimal Wayland client (2000 ms timeout → `idled` after 2000 ms, not instantly). The debounce and the "don't recreate on suppression" change both depend on that; on a compositor that instead reports `idled` immediately for an already-idle seat, the re-arm on release would trigger the behavior the instant playback stops. Happy to gate or adjust this if maintainers know of a compositor that behaves the other way.

## Manual Coverage

- [ ] Tested on Niri
- [x] Tested on Hyprland
- [ ] Tested on Sway

## Screenshots / Videos

No visual changes.

## Checklist

## Additional Notes

No config keys, IPC names, or user-facing strings added — the settings panel's existing live idle status simply stops getting stuck.

The branch carries merge commits from syncing upstream `main`; the two commits that are actually mine are `79dd0c0e` and `febb7280`. Happy to rebase into a linear pair if you prefer that.


TheSkyentist:
I run Niri and I can confirm that this fix works on my setup.

ZhangDM-520 (commented):
Taking you up on the request in #4190 ("review very welcome, since you clearly have a reproducer and I do not run Niri"): I run niri, I reproduced #4190, and I built this PR's head (`ce248bf`) on top of 5.1.0 to drive both halves deliberately. The Niri coverage box can be ticked — details below, including an answer to the open question in your description.

## 1. `resumeIfIdled` — confirmed on niri

Timings shortened to dim 10 s / screen off 15 s / lock 20 s, signal `/sys/class/drm/card1-eDP-1/dpms` (kernel DRM state, not log wording).

| | unpatched | with this PR |
| :-- | :-- | :-- |
| screen-off fires | 08:36:16.601, `dpms=Off` | 08:49:03.947, `dpms=Off` |
| session locks | 08:36:21.666 | 08:49:09.046 |
| panel after the lock | **`dpms=On` 36 ms later (21.702)** | stays `Off` |
| chain replay while locked | `dim` re-fires off the re-arm | none |
| on real input | — | `dim` / `screen-off` / `lock` `resumed`, backlight back at its saved level |

So the early return does not leave a behaviour stuck dimmed: the input that wakes the seat still delivers a real `resumed` and the resume action runs, exactly as the commit argues. #4190's first comment identified the mechanism; this is that conclusion with an objective signal behind it.

## 2. Re-arm restarts the countdown from zero — confirmed on niri too

> destroying and recreating an `ext_idle_notification_v1` restarts the countdown from zero rather than immediately re-reporting accrued seat idleness … I verified that empirically against Hyprland

That claim is load-bearing for the debounce, so here is niri's answer. Inhibit released at 08:51:22.314, `screensaver inhibit released; restarting idle countdowns` at 08:51:23.312 (the 1 s debounce), then:

```
08:51:33.312  dim         +10.000 s
08:51:38.312  screen-off  +15.000 s     dpms=Off at 08:51:38.460
08:51:43.313  lock        +20.001 s
```

Millisecond-exact against the re-arm, not instant — niri does not re-report `idled` for an already-idle seat. Hyprland is therefore not the outlier, and the "behaviour fires the instant playback stops" failure mode you were worried about does not occur here either.

## 3. The inhibit half, against a daemon holding both inhibitor kinds

I have a daemon that takes a logind `idle` block *and* a niri `ScreenSaver.Inhibit` cookie while a video plays, and drops both when playback stops — a combination a browser on its own does not produce. `systemd-inhibit` reported 1 entry, `BlockInhibited=idle:handle-power-key`.

* 35 s held, covering all three timeouts: nothing fired, `dpms=On` throughout, no `suppressed` storm in the log.
* release → `screensaver inhibit released; restarting idle countdowns` → the three waiting behaviours restart from the release moment, numbers above.
* Caveat, stated rather than implied: this daemon holds its inhibitors continuously, so the renewal-swallowing branch never ran here. As far as I can tell nobody in this thread has exercised that path.

## 4. One documentation interaction

`locked_timeout` is read in exactly one place — `effectiveTimeoutSeconds()`, called only from `recreateBehaviorNotification()`. Now that an already-`Idled` behaviour returns before that point, `locked_timeout` can no longer re-target a behaviour that has **already** fired when the session locks; it governs only behaviours still waiting at that moment.

That is the outcome you want (the display is already off, and re-targeting it is precisely the bug), so this is not a change request — but the key's scope is narrower than before, and the "give it a small `locked_timeout` so the wake is undone quickly" workaround users reach for stops working once this lands. `locked_timeout` does not appear anywhere in `docs/user/` today, so one line where the idle behaviours are described would cover both.

I have not tested Hyprland or Sway, and have not reviewed the inhibit-count bookkeeping beyond the niri path above.

ZhangDM-520 (approved):
Approving on top of the review comment above.

Both halves are verified on niri 26.04 from this machine, which is the coverage box that was still unticked:

- **Part 1** (`resumeIfIdled`): with a 10/15/20 s chain, `/sys/class/drm/card1-eDP-1/dpms` now stays `Off` across the lock instead of returning `On` 36 ms after `session is locked`, with no chain replay — and the input that wakes the seat still delivers real `resumed` events, so the backlight comes back at the level `dim` had saved.
- **Part 2** (inhibit tracking): exercised by a daemon here holding a logind `idle` block *and* a niri `ScreenSaver.Inhibit` cookie — 35 s inhibited across all three timeouts with nothing firing, then `screensaver inhibit released; restarting idle countdowns` and a millisecond-exact restart (+10.000 s / +15.000 s / +20.001 s). That also confirms niri restarts the countdown from zero on re-arm rather than re-reporting `idled` immediately, so the Hyprland measurement in the description is not a compositor-specific assumption.

The `locked_timeout` scope note in the review is a documentation suggestion, not a blocker.

ironhak:
Hello

I tried installing this PR but the problem is not fixed. The system still wakes app immediatelly after triggering suspend

yevhen-sychov:
> Hello
> 
> I tried installing this PR but the problem is not fixed. The system still wakes app immediatelly after triggering suspend

Hello, I think your symptom is irrelevant to the pr, can you check journalctl and see what wakes machine?

ironhak:
> > Hello
> > I tried installing this PR but the problem is not fixed. The system still wakes app immediatelly after triggering suspend
> 
> Hello, I think your symptom is irrelevant to the pr, can you check journalctl and see what wakes machine?

Hello pelase see https://github.com/noctalia-dev/noctalia/issues/4190#issuecomment-5767912534

shall i open a new issue?

BBaoVanC:
@TheSkyentist what is your setup like? I'm on Niri, and this PR does stop the screen from turning back on for no reason, but it still resets the timer after the lock event and it has to wait the full 5 minutes again before sleeping; notice the 20 second gap before `idle behavior 'suspend' pre-action fade`

```toml
[idle]
behavior_order = ["screen_off", "lock", "suspend"]
pre_action_fade_seconds = 5.0
[idle.behavior.screen-off]
timeout = 10
action = "screen_off"
[idle.behavior.lock]
timeout = 15
action = "lock"
[idle.behavior.suspend]
timeout = 20
action = "command"
command = "~/.config/noctalia/scripts/sleep.sh"
```

```
04:07:00.532 [INF] [idle] idle behavior 'screen-off' pre-action fade 5000ms
04:07:05.531 [INF] [idle] idle behavior 'lock' joined active pre-action fade
04:07:05.553 [INF] [idle] idle behavior 'screen-off' triggered after pre-action fade
04:07:05.836 [INF] [idle] idle behavior 'lock' triggered after pre-action fade
04:07:05.836 [INF] [session] lock requested: compositor=Niri env_hint="niri:niri:niri" xdg_session_id=2 user=bbaovanc
04:07:05.839 [INF] [lockscreen] session lock requested
04:07:05.839 [INF] [session] lock: lock screen requested
04:07:05.842 [INF] [texcache] hit /home/bbaovanc/wallpapers/evgeny-matveev-f4hR4QdFP04-unsplash.jpg (refCount=2)
04:07:06.408 [INF] [fingerprint] using fingerprint device /net/reactivated/Fprint/Device/0
04:07:06.408 [INF] [lockscreen] session is locked
04:07:06.670 [INF] [fingerprint] claimed fingerprint device
04:07:26.464 [INF] [idle] idle behavior 'suspend' pre-action fade 5000ms
04:07:31.715 [INF] [idle] idle behavior 'suspend' triggered after pre-action fade
04:07:31.773 [INF] [logind] logind sleep delay inhibit released
```

TheSkyentist:
@BBaoVanC I'm on the same build, and it behaves the same way for me. I should be clearer about what it fixes for me. Here's my setup:

```
[idle]
behavior_order = ["Dim", "lock", "screen-off"]
pre_action_fade_seconds = 0

[idle.behavior.Dim]
action = "command"
command = "brightnessctl -s set 5%"
resume_command = "brightnessctl -r"
timeout = 60.0

[idle.behavior.lock]
action = "lock"
timeout = 300.0

[idle.behavior.screen-off]
action = "screen_off"
timeout = 315.0
locked_timeout = 10.0
```

On main, when the session locks at 300s, the lock re-armed Dim, which had already fired, and ran its resume_command. So the screen jumped back to full brightness as soon as the lock screen appeared, then dimmed again 60s later. This PR fixed that; the screen stays dimmed until I actually touch an input.

But I still see the same timer reset as you. To get around this I use locked_timeout to control the timers after a lock. So my "screen off" never fires via the 315 timeout (I could set it to be 1000 or whatever as long as its >300), only 10 seconds after it locks after the 300 second timeout. In your case, putting locked_timeout = 5 on suspend would get you roughly the timing you expected. 

I still do find the "behaviour_order" kind of confusing though since it implies some kind of ordering after each one fires, but in practice it doesn't have an effect. Could be worth rethinking how this is presented to the user? But probably beyond the scope of this PR.

TheSkyentist:
I just realized, the locked_timeout option isn't exposed in the GUI for idle commands, but I'm pretty sure it's respected in config.  Might be worth a separate Issue/PR asking for this to be exposed.

BBaoVanC:
> In your case, putting locked_timeout = 5 on suspend would get you roughly the timing you expected.

Hmm, but I'm guessing that's not intended behavior; it worked fine with my config before 9691c87ddaef4e8234d733d568ecf3376aea2159.

TheSkyentist:
@BBaoVanC Ah okay, I think I understand now, and I agree. Looking at 9691c87 (#3388), locking now re-arms _every_ idle behavior so that the ones with a locked_timeout can switch over to it, but re-arming restarts the countdown for all of them. Probably the expected behavior is that only timers with a locked_timeout specified should rearm. That would also make locked_timeout a lot more useful as an opt-in, which hopefully gets exposed for more than just screen off (I opened #4567 for that). 

Not sure if the reset fix belongs in this PR or a separate one, though. Probably since this PR is already approved, maybe after? If this is the expected behavior I can try to tackle the fix after this is merged.

PR #4004, opened 2026-08-18
fix(brightness): disambiguate ddcutil buses for identical monitors

### LINKED ISSUE #3974: [BUG] Two identical monitors: DDC brightness lists the first connector twice and leaves the second uncontrollable
### Submission checklist

### Bug description

With two identical monitors, the monitor tab in control centre shows three cards, it shows the first monitor twice, and the sliders are linked, and it shows the second monitor once and it does not have a functional slider.

The cause is that `ddcutil detect` reports the same `DRM_connector` for both displays, Noctalia then creates two `DisplayInternal` entries with the same `pub.id`, so each lookup then resolves an id to the first match so the second DDC bus is unreachable. In reality they are different but the only difference in the `ddcutil detect` output is the I2C bus ID (see output below).

`rebuildPublic` walks the Wayland outputs which is why the second monitor is still listed in the widget, but Noctalia cannot find the ddc entry so it is uncontrollable.

### Steps to reproduce

1. Connect two monitors of the same model
2. Confirm that EDIDs are identical using `md5sum /sys/class/drm/card*-DP-1/edid /sys/class/drm/card*-DP-2/edid`, the hashes should be the same.
3. Enable ddcutil control in noctalia settings or toml
4. Open Monitor tab in Control Center

### Expected behavior

One monitor card per monitor, each with a working and independent brightness slider.

### Actual behavior

Three cards, DP-1 twice with sliders both controlling the i2c bus of the first monitor. DP-2 once with greyed out controls.
Image

(Interestingly dragging the slider of the first card drags both sliders in lockstep, but dragging the slider of the second card only changes the slider of that card and changes the number of the top slider but not the slider itself. Both affect the brightness of the first monitor though. I assume this is some separate issue but doesn't really matter all that much.)

### Logs / error output

```text
Output of `ddcutil detect --verbose`
Display 1
   I2C bus:  /dev/i2c-5
      DRM_connector:                             card1-DP-1
      /sys/class/drm/card1-DP-1/dpms             On
      /sys/class/drm/card1-DP-1/enabled          enabled
      /sys/class/drm/card1-DP-1/status           connected
      /sys/class/drm/card1-DP-1/connector_id     826
      Driver:                                    nvidia
      I2C address 0x30 (EDID block#)  present:   false
      EDID exists:                               true
      I2C address 0x37 (DDC) responsive:         true
      Is LVDS or EDP display:                    false
      Is laptop display by EDID:                 false
      Is laptop display:                         false
      /sys/bus/i2c/devices/i2c-5/name            NVIDIA i2c adapter 4 at 1:00.0
      PCI device path:                           /sys/devices/pci0000:00/0000:00:01.1/0000:01:00.0/i2c-5
   EDID synopsis:
      Mfg id:               MSI - Microstep
      Model:                G272QPF E2
      Product code:         19667  (0x4cd3)
      Serial number:        
      Binary serial number: 0 (0x00000000)
      Manufacture year:     2023,  Week: 0
      EDID version:         1.4
      Extra descriptor:        
      Video input definition:    0xb5 - Digital Input (DisplayPort), Bit depth: 10
      Supported features:
         DPMS active-off
         Digital display type: RGB 4:4:4 + YCrCb 4:4:4 + YCrCb 4:2:2
         Standard sRGB color space: True
      White x,y:        0.313, 0.329
      Red   x,y:        0.664, 0.329
      Green x,y:        0.279, 0.672
      Blue  x,y:        0.152, 0.044
      Extension blocks: 2
   EDID source: I2C
   EDID hex dump:
        +0          +4          +8          +c            0   4   8   c   
+0000   00 ff ff ff ff ff ff 00 36 69 d3 4c 00 00 00 00   ........6i.L....
+0010   00 21 01 04 b5 3c 22 78 3b 18 15 aa 54 47 ac 27   .!...<"x;...TG.'
+0020   0b 50 54 bf cf 00 d1 c0 71 4f 81 c0 81 40 81 80   .PT.....qO...@..
+0030   95 00 b3 00 d1 fc 56 5e 00 a0 a0 a0 29 50 30 20   ......V^....)P0 
+0040   35 00 55 50 21 00 00 1a 6f c2 00 a0 a0 a0 55 50   5.UP!...o.....UP
+0050   30 20 35 00 55 50 21 00 00 1a 00 00 00 fd 0c 30   0 5.UP!........0
+0060   b4 1e 1e 4e 01 0a 20 20 20 20 20 20 00 00 00 fc   ...N..      ....
+0070   00 47 32 37 32 51 50 46 20 45 32 0a 20 20 02 fa   .G272QPF E2.  ..
   VCP version:         2.1
   Controller mfg:      Novatek
   Firmware version:    0.0
   Monitor correctly uses invalid feature flag in DDC reply packet to indicate unsupported feature.
   Current dynamic sleep adjustment multiplier:  0.00
   Monitor Model Id:  MSI-G272QPF_E2-19667
   Feature definition file MSI-G272QPF_E2-19667.mccs not found.

Display 2
   I2C bus:  /dev/i2c-6
      DRM_connector:                             card1-DP-1
      /sys/class/drm/card1-DP-1/dpms             On
      /sys/class/drm/card1-DP-1/enabled          enabled
      /sys/class/drm/card1-DP-1/status           connected
      /sys/class/drm/card1-DP-1/connector_id     826
      Driver:                                    nvidia
      I2C address 0x30 (EDID block#)  present:   false
      EDID exists:                               true
      I2C address 0x37 (DDC) responsive:         true
      Is LVDS or EDP display:                    false
      Is laptop display by EDID:                 false
      Is laptop display:                         false
      /sys/bus/i2c/devices/i2c-6/name            NVIDIA i2c adapter 5 at 1:00.0
      PCI device path:                           /sys/devices/pci0000:00/0000:00:01.1/0000:01:00.0/i2c-6
   EDID synopsis:
      Mfg id:               MSI - Microstep
      Model:                G272QPF E2
      Product code:         19667  (0x4cd3)
      Serial number:        
      Binary serial number: 0 (0x00000000)
      Manufacture year:     2023,  Week: 0
      EDID version:         1.4
      Extra descriptor:        
      Video input definition:    0xb5 - Digital Input (DisplayPort), Bit depth: 10
      Supported features:
         DPMS active-off
         Digital display type: RGB 4:4:4 + YCrCb 4:4:4 + YCrCb 4:2:2
         Standard sRGB color space: True
      White x,y:        0.313, 0.329
      Red   x,y:        0.664, 0.329
      Green x,y:        0.279, 0.672
      Blue  x,y:        0.152, 0.044
      Extension blocks: 2
   EDID source: I2C
   EDID hex dump:
        +0          +4          +8          +c            0   4   8   c   
+0000   00 ff ff ff ff ff ff 00 36 69 d3 4c 00 00 00 00   ........6i.L....
+0010   00 21 01 04 b5 3c 22 78 3b 18 15 aa 54 47 ac 27   .!...<"x;...TG.'
+0020   0b 50 54 bf cf 00 d1 c0 71 4f 81 c0 81 40 81 80   .PT.....qO...@..
+0030   95 00 b3 00 d1 fc 56 5e 00 a0 a0 a0 29 50 30 20   ......V^....)P0 
+0040   35 00 55 50 21 00 00 1a 6f c2 00 a0 a0 a0 55 50   5.UP!...o.....UP
+0050   30 20 35 00 55 50 21 00 00 1a 00 00 00 fd 0c 30   0 5.UP!........0
+0060   b4 1e 1e 4e 01 0a 20 20 20 20 20 20 00 00 00 fc   ...N..      ....
+0070   00 47 32 37 32 51 50 46 20 45 32 0a 20 20 02 fa   .G272QPF E2.  ..
   VCP version:         2.1
   Controller mfg:      Novatek
   Firmware version:    0.0
   Monitor correctly uses invalid feature flag in DDC reply packet to indicate unsupported feature.
   Current dynamic sleep adjustment multiplier:  0.00
   Monitor Model Id:  MSI-G272QPF_E2-19667
   Feature definition file MSI-G272QPF_E2-19667.mccs not found.

Displays with I2C bus numbers 5, 6 have identical EDIDs.
DRM connector names may not be accurate.

Output of `md5sum /sys/class/drm/card*-DP-1/edid /sys/class/drm/card*-DP-2/edid`
d72b14daa8a3e5f240116d2a87861f7b  /sys/class/drm/card1-DP-1/edid
d72b14daa8a3e5f240116d2a87861f7b  /sys/class/drm/card1-DP-2/edid


Noctalia's log registers two displays under one id — note `connector=DP-1` on both lines:


[INF] [brightness] ddcutil detect parsed 2 candidate display(s)
[INF] [brightness] found ddcutil display connector=DP-1 bus=5 current=80%
[INF] [brightness] found ddcutil display connector=DP-1 bus=6 current=80%



[DBG] [brightness] ddcutil write completion for 'DP-1' ignored (stale epoch 8 vs 13)
[DBG] [brightness] ddcutil write completion for 'DP-1' ignored (stale epoch 14 vs 20)
[DBG] [brightness] ddcutil refresh completion for 'DP-1' ignored (stale epoch 0 vs 102)
```

### Noctalia version

noctalia v5.0.0 (5.0.0_beta.8-1-dirty)

### Compositor

Niri

### Distribution

Arch-based

### Environment information

```text
Installation method: cachyos-extra-znver4/noctalia package
GPU: 4070 super, using nvidia-open-dkms
Monitors: 2x MSI G272QPF E2 on DP-1 and DP-2 slots of the GPU
```

### Is this more suited to an upstream ddcutil fix?
I would argue no, ddcutil has two ways to map an i2c bus to a connector, firstly it can use `/sys/class/drm/cardN-XXX/ddc`, but it seems that NVIDIA's drivers do not create this symlink (so arguably this is a classic nvidia issue but wcyd). Next, it fallsback to comparing EDIDs, but since the EDID of these monitors is identical this doesn't help.

In the [1.4.0 ddcutil release notes it mentions this case specifically](https://www.ddcutil.com/release_notes/#multiple-monitors-having-the-same-edid).

### Is this a duplicate of #1427
No, although the issue presents similarly, that issue was caused by a parsing bug which has now been resolved.

### Potential fixes
In `applyDetectComplection`
Group candidates by `connectorName`, if any names are claimed more than once we do the following for each connector.
1. Read the sysfs EDID through `/sys/class/drm/card*-<connector>/edid` for the claimed connector and for each Wayland output, collect the outputs whose EDID match.
2. Sort the collected candidates by i2c bus, sort the collected outputs by connectorName, and assign the `connectorName` appropriately.

In my case this would give us
Two candidates are claiming the DP-1 connector.
We read that they have the same EDID, and we read that the DP-1 and DP-2 wayland outputs also have this same EDID.
We assign connectorNames such that we now have (bus5, DP-1), (bus6, DP-2)

The sorting and assigning in step 2 is a guess, it's possible that this could result in the bus for DP-1 actually claiming connector name DP-2 or vise versa, but this is still preferable to one of them being unreachable. 
The fix to this guessing could be a config option to pin a connector to a ddc_bus, e.g.
[brightness.monitor.DP-2]
ddc_bus = 6 

If this fix sounds appropriate then I'm happy to look into implementing and then raising a PR for it.

ybbat:
## Summary

- Added duplicate detection so that there will never be multiple displays with the same id
-  Added config capability for pinning a monitor to a certain bus, example below
```
[brightness.monitor.DP-2]
ddc_bus = "i2c-6"
```

## Motivation

Previously if multiple monitors had the same EDID then they would have the same display.pub.id, this caused an extra monitor card to show up in control centre for the first monitor, and the card for the second monitor was greyed out and uncontrollable. This resolves the extra card by adding a duplicate check, and resolves the second monitor being uncontrollable by letting the user override the connector to bus mapping in config.

## Type of Change

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Build / packaging

## Related Issue

Resolves #3974

## Testing

Added new monitor override with this option to config_schema_roundtrip_test.cpp, passed under `just test`.
Verified that without the bus pinning in config my machine does not show an extra instance of DP-1 as it did before this fix.
Verified that with the bus pinning in config then the monitor tab in control centre correctly shows DP-1 and DP-2 and both are controllable.

## Manual Coverage

- [x] Tested on Niri
- [ ] Tested on Hyprland
- [ ] Tested on Sway

## Screenshots / Videos
No UI changes

## Checklist

## Additional Notes

I have moved the optional-string schema codec into `field.h` as a `field()` overload, replacing a file local `optionalStringField` in `config_schema.cpp` which was defined in an anonymous namespace below where I wanted to use it. I also noticed that there were other fields that don't seem specific to a particular config and could probably be moved to `field.h` too, for example `optionalIntField`, `optionalBoolField`, etc. I only moved `optionalStringField` because that was the only one I needed for this change, marking it as something that could probably be fixed in future.

### DIFF
--- a/example.toml
+++ b/example.toml
@@ -277,6 +277,8 @@ enable_ddcutil = false
 # backlight_device = "intel_backlight"  # explicit sysfs device name or path; run: noctalia msg brightness-list-backlight-devices
 # [brightness.monitor.DP-1]
 # backend = "ddcutil"
+# [brightness.monitor.DP-2]
+# ddc_bus = "i2c-6"                 # pin the I2C bus when monitors share an EDID
 
 # ── Night Light ───────────────────────────────────────────────────────────────
 
--- a/example.toml
+++ b/example.toml
@@ -278,7 +278,7 @@ enable_ddcutil = false
 # [brightness.monitor.DP-1]
 # backend = "ddcutil"
 # [brightness.monitor.DP-2]
-# ddc_bus = "i2c-6"                 # pin the I2C bus when monitors share an EDID
+# ddc_bus = 6                       # pin the I2C bus when monitors share an EDID
 
 # ── Night Light ───────────────────────────────────────────────────────────────
 
--- a/src/config/config_types.h
+++ b/src/config/config_types.h
@@ -1269,6 +1269,7 @@ struct BrightnessMonitorOverride {
   std::string match;
   std::optional<BrightnessBackendPreference> backend;
   std::optional<std::string> backlightDevice; // sysfs device name or path, e.g. "intel_backlight"
+  std::optional<std::string> ddcBus;          // DDC bus name, e.g. "i2c-0"
 
   bool operator==(const BrightnessMonitorOverride&) const = default;
 };
--- a/src/config/config_types.h
+++ b/src/config/config_types.h
@@ -1269,7 +1269,7 @@ struct BrightnessMonitorOverride {
   std::string match;
   std::optional<BrightnessBackendPreference> backend;
   std::optional<std::string> backlightDevice; // sysfs device name or path, e.g. "intel_backlight"
-  std::optional<std::string> ddcBus;          // DDC bus name, e.g. "i2c-0"
+  std::optional<std::int32_t> ddcBus;         // DDC bus number, e.g. 6 for i2c-6
 
   bool operator==(const BrightnessMonitorOverride&) const = default;
 };
--- a/src/config/schema/config_schema.cpp
+++ b/src/config/schema/config_schema.cpp
@@ -422,6 +422,7 @@ namespace noctalia::config::schema {
           field(&BrightnessMonitorOverride::match, "match"),
           optionalEnumField(&BrightnessMonitorOverride::backend, "backend", kBrightnessBackendPreferences),
           field(&BrightnessMonitorOverride::backlightDevice, "backlight_device"),
+          field(&BrightnessMonitorOverride::ddcBus, "ddc_bus"),
       };
       return s;
     }
--- a/src/config/schema/config_schema.cpp
+++ b/src/config/schema/config_schema.cpp
@@ -421,19 +421,7 @@ namespace noctalia::config::schema {
       static const Schema<BrightnessMonitorOverride> s = {
           field(&BrightnessMonitorOverride::match, "match"),
           optionalEnumField(&BrightnessMonitorOverride::backend, "backend", kBrightnessBackendPreferences),
-          custom<BrightnessMonitorOverride>(
-              "backlight_device",
-              [](const toml::table& tbl, BrightnessMonitorOverride& out, std::string_view, Diagnostics&) {
-                if (auto v = tbl["backlight_device"].value<std::string>()) {
-                  out.backlightDevice = *v;
-                }
-              },
-              [](toml::table& tbl, const BrightnessMonitorOverride& in) {
-                if (in.backlightDevice.has_value()) {
-                  tbl.insert_or_assign("backlight_device", *in.backlightDevice);
-                }
-              }
-          ),
+          field(&BrightnessMonitorOverride::backlightDevice, "backlight_device"),
       };
       return s;
     }
@@ -1949,24 +1937,8 @@ namespace noctalia::config::schema {
       );
     }
 
-    template <typename Struct>
-    Field<Struct> optionalStringField(std::optional<std::string> Struct::* member, std::string_view key) {
-      return custom<Struct>(
-          key,
-          [member, key](const toml::table& tbl, Struct& out, std::string_view, Diagnostics&) {
-            if (auto v = tbl[key].value<std::string>()) {
-              out.*member = *v;
-            }
-          },
-          [member, key](toml::table& tbl, const Struct& in) {
-            if ((in.*member).has_value()) {
-              tbl.insert_or_assign(key, *(in.*member));
-            }
-          }
-      );
-    }
-
-    // Like optionalStringField but trims; a present-but-empty value stays unset so it inherits the parent.
+    // Like field(std::optional<std::string>…) but trims; a present-but-empty value stays unset so it inherits the
+    // parent.
     template <typename Struct>
     Field<Struct> optionalTrimmedStringField(std::optional<std::string> Struct::* member, std::string_view key) {
       return custom<Struct>(
@@ -2223,7 +2195,7 @@ namespace noctalia::config::schema {
   const Schema<BarMonitorOverride>& barMonitorOverrideSchema() {
     static const Schema<BarMonitorOverride> s = {
         field(&BarMonitorOverride::match, "match"),
-        optionalStringField(&BarMonitorOverride::position, "position"),
+        field(&BarMonitorOverride::position, "position"),
         optionalBoolField(&BarMonitorOverride::enabled, "enabled"),
         optionalBoolField(&BarMonitorOverride::autoHide, "auto_hide"),
         optionalBoolField(&BarMonitorOverride::smartAutoHide, "smart_auto_hide"),
--- a/src/config/schema/field.h
+++ b/src/config/schema/field.h
@@ -185,6 +185,22 @@ namespace noctalia::config::schema {
     };
   }
 
+  template <typename Struct> Field<Struct> field(std::optional<std::string> Struct::* member, std::string_view key) {
+    return Field<Struct>{
+        key,
+        [member, key](const toml::table& tbl, Struct& out, std::string_view, Diagnostics&) {
+          if (auto v = tbl[key].value<std::string>()) {
+            out.*member = *v;
+          }
+        },
+        [member, key](toml::table& tbl, const Struct& in) {
+          if ((in.*member).has_value()) {
+            tbl.insert_or_assign(key, *(in.*member));
+          }
+        }
+    };
+  }
+
   template <typename Struct> Field<Struct> field(std::vector<std::string> Struct::* member, std::string_view key) {
     return Field<Struct>{
         key,
--- a/src/system/brightness_service.cpp
+++ b/src/system/brightness_service.cpp
@@ -23,12 +23,15 @@
 #include <dirent.h>
 #include <filesystem>
 #include <fstream>
+#include <map>
 #include <memory>
 #include <mutex>
 #include <optional>
 #include <queue>
+#include <ranges>
 #include <sdbus-c++/IProxy.h>
 #include <sdbus-c++/Types.h>
+#include <set>
 #include <string>
 #include <string_view>
 #include <sys/epoll.h>
@@ -620,6 +623,53 @@ namespace {
     }
     kLog.info("ddcutil detect parsed {} candidate display(s)", candidates.size());
     return candidates;
+  }
+
+  std::map<int, std::string> getBusOverrides(const BrightnessConfig& config, const WaylandConnection& wayland) {
+    std::map<int, std::string> busOverrides;
+    std::set<int> contestedBuses;
+
+    for (const auto& output : wayland.outputs()) {
+      if (output.connectorName.empty()) {
+        continue;
+      }
+
+      const auto override = std::ranges::find_if(config.monitorOverrides, [&output](const auto& entry) {
+        return !entry.match.empty() && outputMatchesSelector(entry.match, output);
+      });
+      if (override == config.monitorOverrides.end() || !override->ddcBus.has_value()) {
+        continue;
+      }
+
+      const BrightnessBackendPreference preference = backendPreferenceForOutput(config, &output);
+      if (preference == BrightnessBackendPreference::None || preference == BrightnessBackendPreference::Backlight) {
+        kLog.debug(
+            "ignoring ddc_bus '{}' for connector {} because backend is not set to 'ddcutil'", *override->ddcBus,
+            output.connectorName
+        );
+        continue;
+      }
+
+      const auto parsedBus = parseI2cBus(*override->ddcBus);
+      if (!parsedBus.has_value()) {
+        kLog.warn("ddc_bus '{}' for connector {} could not be parsed", *override->ddcBus, output.connectorName);
+        continue;
+      }
+
+      if (const auto [it, inserted] = busOverrides.try_emplace(*parsedBus, output.connectorName); !inserted) {
+        kLog.warn(
+            "ignoring ddc_bus {} because both '{}' and '{}' pin it; only one connector can own a bus", *parsedBus,
+            it->second, output.connectorName
+        );
+        contestedBuses.insert(*parsedBus);
+      }
+    }
+
+    for (const int contestedBus : contestedBuses) {
+      busOverrides.erase(contestedBus);
+    }
+
+    return busOverrides;
   }
 
 } // namespace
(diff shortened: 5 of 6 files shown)


ItsLemmy:
## Verdict

Right approach, real bug fixed, but **two blocking issues**: an explicit `ddc_bus` pin can be silently overruled by ddcutil's guess, and the config value is parsed by a scanner that accepts `"DP-2"` as bus 2.


## Blocking

### B1 a pin loses to candidate order

`applyDetectCompletion` renames pinned candidates but never revokes the connector from unpinned ones. Trace with the issue's own topology (candidates `[(bus5, DP-1), (bus6, DP-1)]`) and a user who pins the *reported* connector to the *other* bus:

```toml
[brightness.monitor.DP-1]
ddc_bus = "i2c-6"
```

- `busOverrides = {6: "DP-1"}`
- candidate `bus5` → unpinned, keeps ddcutil's `DP-1` → claims it (line 1360-1368)
- candidate `bus6` → renamed `DP-1` → `findInternal` hit → dropped (line 1351)

Result: `DP-1` is driven by bus 5 in direct contradiction of the config, `DP-2` stays uncontrollable, and the emitted warning tells the user to pin a connector to bus 6 — which they just did. Explicit config being silently overridden by a guess is exactly what `AGENTS.md` "No fallbacks for identity & configuration" forbids.

The invariant to enforce: **a connector pinned to bus B is claimable only by bus B.** Cheapest fix: have `getBusOverrides` also return the pinned connector names, and skip any candidate that is *not* itself pinned whose `connectorName` is in that set. Full-swap (`DP-1→i2c-6`, `DP-2→i2c-5`) and single-sided-correct (`DP-2→i2c-6`) already work; this closes the third case.

### B2  `ddc_bus` is a string fed to a ddcutil-output scanner

`parseI2cBus` exists to chew `I2C bus:  /dev/i2c-5` lines. Pointing it at a config value makes the accepted syntax undefined. Compiled the verbatim `parseTrailingInteger`/`parseI2cBus` from the branch against candidate values:

| value | parsed |
|---|---|
| `"i2c-6"` | 6 |
| `"6"` | 6 |
| `"/dev/i2c-6"` | 6 |
| `"DP-2"` | **2** |
| `"eDP-1"` | **1** |
| `"i2c-"` | **2** |
| `"6.5"` | **5** |
| `"-6"` | 6 |
| `"0x06"` | 6 |
| `"banana"` | none |

Four accepted spellings for one identity, plus silent misparses of plausible typos (`ddc_bus = "DP-2"` binds bus 2 and reports success). The bus is an `int` end to end: `DisplayInternal::ddcBus` (line 68), `--bus std::to_string(bus)` (line 507-508). Make the config an int:

```cpp
std::optional<std::int32_t> ddcBus;                    // config_types.h
field(&BrightnessMonitorOverride::ddcBus, "ddc_bus")   // field.h:172 overload already exists
```

`optionalIntField(..., Range)` in `config_schema.cpp:2030` is the alternative if you want bounds. Either validates at parse time instead of at detect time, and `example.toml` becomes `ddc_bus = 6`. This is also what the issue originally proposed.

Note this removes the *need* for the `field(std::optional<std::string>)` move — keep that commit anyway, it deletes a duplicated `custom<>` block and lands next to the existing `field(std::optional<std::int32_t>)`, which is the correct home.

## Non-blocking

- **`std::views::transform` with a side-effecting log (line 1317-1328).** Lazy view: the `kLog.info` fires on each dereference, so any future second pass double-logs, and `auto candidate` copies the struct per element. Just resolve inline at the top of the existing loop — one lookup, no view, no copy.
- **`findInternal(candidate.connectorName)` (line 1351)** matches on `pub.id`, not `connectorName`. It happens to work only because every construction site sets `pub.id = connectorName` (lines 831, 1365). Two lines above you already use the correct idiom (`std::ranges::any_of` over `display.connectorName`); mirror it and scope to `backend == RuntimeBackend::Ddcutil`.
- **Naming.** Neighbouring helpers are `backendPreferenceForOutput` / `backlightDeviceForOutput`; `getBusOverrides` breaks the convention and undersells what it returns (bus → pinned connector). `pinnedConnectorsByBus` reads better. `std::map`/`std::set` for ≤ a handful of monitors also drags in two headers for no gain.
- **Trailing space** in the log format string, line 1322: `"...as per config "`.
- **Discoverability gap.** `backlight_device` ships `noctalia msg brightness-list-backlight-devices` (`schema_msg.h:295`) and `example.toml` points at it. `ddc_bus` has no counterpart — the user must run `ddcutil detect` and interpret it. A `brightness-list-ddc-buses` command would make the pair symmetric; fine as a follow-up.

## Good

Dropping the duplicate rather than publishing two internals with the same `pub.id` is strictly better than the status quo and fixes the third card. Erasing *both* sides of a contested bus instead of picking one is the right call. Roundtrip test extended to cover both new optional fields.

## Docs follow-up

`../noctalia-docs/src/content/docs/noctalia/services/brightness.mdx` documents `backlight_device`; `ddc_bus` needs the same treatment (separate repo, separate commit).

ybbat:
> ## Blocking
> ### B1 a pin loses to candidate order
Thanks for catching this, resolved by adding a guard that checks if the connector this bus claims is pinned to a different bus through the config
> ### B2 `ddc_bus` is a string fed to a ddcutil-output scanner
Fixed as suggested, kept it unbounded; the bounded int field silently clamps into the bounds which doesn't seem appropriate for this field.
> ## Non-blocking
> 
>     * **`std::views::transform` with a side-effecting log (line 1317-1328).** Lazy view: the `kLog.info` fires on each dereference, so any future second pass double-logs, and `auto candidate` copies the struct per element. Just resolve inline at the top of the existing loop — one lookup, no view, no copy.
Removed the use of transform in favour of resolving the bus/connectorname at the top of the candidate loop. 
I did change the loop to `auto candidate`, this is so the pin rewrites it in place for this loop rather than using a local `connectorName` variable. This adds a copy for each candidate but it removes a potential source of confusion if a future contributor missed that `connectorName` is correct and `candidate.connectorName` is wrong. This is done eagerly so it isn't a copy per dereference like the previous `transform` code. The copy could be removed by dropping `const` from `applyCompletion` and `applyDetectCompletion` but the copy doesn't cost much here and I didn't want to change mutability in a wider scope than needed.


>     * **`findInternal(candidate.connectorName)` (line 1351)** matches on `pub.id`, not `connectorName`. It happens to work only because every construction site sets `pub.id = connectorName` (lines 831, 1365). Two lines above you already use the correct idiom (`std::ranges::any_of` over `display.connectorName`); mirror it and scope to `backend == RuntimeBackend::Ddcutil`.
Fixed as suggested


>     * **Naming.** Neighbouring helpers are `backendPreferenceForOutput` / `backlightDeviceForOutput`; `getBusOverrides` breaks the convention and undersells what it returns (bus → pinned connector). `pinnedConnectorsByBus` reads better. `std::map`/`std::set` for ≤ a handful of monitors also drags in two headers for no gain.
Renamed to `pinnedConnectorByBus`.
I have kept `std::map`/`std::set`, happy to switch if you feel strongly but they are already included transitively so this is just making it clearer with no additional cost. The gain is having containers that enforce the invariant the variables aim to achieve, making for code that's more concise and easier to maintain correctness.


>     * **Trailing space** in the log format string, line 1322: `"...as per config "`.
Fixed


>     * **Discoverability gap.** `backlight_device` ships `noctalia msg brightness-list-backlight-devices` (`schema_msg.h:295`) and `example.toml` points at it. `ddc_bus` has no counterpart — the user must run `ddcutil detect` and interpret it. A `brightness-list-ddc-buses` command would make the pair symmetric; fine as a follow-up.
I can do this if you think it's needed, but I don't think this command would be very useful.

Thinking through what a user would do, they'd either notice that their monitors aren't controllable or they'd see the warning in the logs, they would go to the docs which would point them at this command. The command would print two rows with identical connector names, make, models, serial numbers, etc, differing only in bus number. The only thing it tells them really is what bus numbers exist, and the warning log already names which bus needs pinning to a connector. They'd still have to then work out which bus corresponds to which monitor; pin one, see which panel is controlled, and swap if it was the wrong one.

So it would basically just be a wrapper around `ddcutil detect` without any extra value beyond being `noctalia msg brightness-list-ddc-buses` instead of `ddcutil detect`. `brightness-list-backlight-devices` is different in that the device names are distinct and list different hardware paths, so the command gives you extra information you can act on.

> ## Docs
Will do, will follow up on that once this PR is done

ItsLemmy:
Thanks! Looking forward to the docs follow up.

PR #4011, opened 2026-08-19
feat(wallpaper): sync wallpaper with light/dark theme

### LINKED ISSUE #1612: Switch wallpaper according to light/dark theme changes

mdj2812:
## Summary

- Add `[wallpaper.theme_sync]` config with global and per-monitor `path_light` / `path_dark` bindings, plus a settings toggle
- Switch wallpapers when the resolved theme changes, with a 150ms startup debounce to avoid double transitions on login
- Assign bindings from the wallpaper panel Dark/Light tabs (and Auto via resolved appearance); theme sync must be enabled in settings first

Closes #1612

## Test plan

- [x] Unit tests (`wallpaper_theme_sync_paths_test`)
- [x] Desktop: `wallpaper-set` updates bindings for resolved mode; theme mode switch applies bound wallpapers
- [x] Global bindings always updated alongside per-monitor overrides
- [x] Manual: wallpaper panel picks on Dark/Light/Auto tabs with theme sync enabled
- [x] Manual: verify startup debounce when theme flips shortly after login

https://github.com/user-attachments/assets/dfb9d412-8148-4492-87a4-68b6edf9a116

Made with [Cursor](https://cursor.com)

### DIFF
--- a/assets/translations/en.json
+++ b/assets/translations/en.json
@@ -3126,7 +3126,7 @@
           "placeholder": "~/Pictures/Wallpapers"
         },
         "theme-sync": {
-          "description": "Automatically switch wallpaper when the resolved light or dark theme changes. Pick a wallpaper under the Dark or Light tab in the wallpaper panel to assign it.",
+          "description": "Automatically switch wallpaper when the resolved light or dark theme changes. Enable this, then pick wallpapers under the Dark or Light tab in the wallpaper panel to assign bindings.",
           "label": "Sync Wallpaper with Theme"
         },
         "edge-smoothness": {
--- a/assets/translations/en.json
+++ b/assets/translations/en.json
@@ -3125,6 +3125,10 @@
           "label": "Light Mode Directory",
           "placeholder": "~/Pictures/Wallpapers"
         },
+        "theme-sync": {
+          "description": "Automatically switch wallpaper when the resolved light or dark theme changes. Pick a wallpaper under the Dark or Light tab in the wallpaper panel to assign it.",
+          "label": "Sync Wallpaper with Theme"
+        },
         "edge-smoothness": {
           "description": "Feathering of transition edges between wallpapers",
           "label": "Edge Smoothness"
--- a/assets/translations/en.json
+++ b/assets/translations/en.json
@@ -3245,7 +3245,7 @@
           "placeholder": "~/Pictures/Wallpapers"
         },
         "theme-sync": {
-          "description": "Automatically switch wallpaper when the resolved light or dark theme changes. Pick a wallpaper under the Dark or Light tab in the wallpaper panel to assign it.",
+          "description": "Automatically switch wallpaper when the resolved light or dark theme changes. Enable this, then pick wallpapers under the Dark or Light tab in the wallpaper panel to assign bindings.",
           "label": "Sync Wallpaper with Theme"
         },
         "edge-smoothness": {
--- a/assets/translations/en.json
+++ b/assets/translations/en.json
@@ -3244,6 +3244,10 @@
           "label": "Light Mode Directory",
           "placeholder": "~/Pictures/Wallpapers"
         },
+        "theme-sync": {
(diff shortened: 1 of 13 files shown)


musaad-sharikh:
Use case from a current 5.1.0 user, in case it's useful for this PR.

I run `[theme] mode = "auto"`, so the desktop switches light/dark on the
sunrise/sunset schedule. I keep one wallpaper folder for light and one for
dark, and I pick the wallpaper I actually want for each mode by hand — no
random, no `[wallpaper.automation]`.

Today the theme flips on schedule but the wallpaper stays whatever was last
applied, so I regularly end up with a dark wallpaper under the light theme and
the other way round until I change it by hand. `directory_light` /
`directory_dark` don't cover this: they scope the pool for the *next* pick and
never re-apply an already-set wallpaper. On my install a resolved dark→light
transition produces no wallpaper change at all.

`[wallpaper.theme_sync]` with `path_light` / `path_dark` is exactly the shape I
need — one deliberately chosen image per mode instead of a pool, assigned from
the Dark/Light tabs so it keeps whichever wallpaper I last picked in that mode,
and re-applied when the *resolved* mode changes. Keying on the resolved mode is
what makes this work with `mode = "auto"` rather than only with a manually
pinned light or dark.

Single monitor here (eDP-1), so the global `path_light` / `path_dark` pair alone
would cover me. No expectations on timing — just noting there's a real user for
this. Happy to test if that's useful.

PR #4014, opened 2026-08-19
refactor(settings): consolidate launcher provider settings

Ape:
Generate launcher provider prefix and global-search settings from a single descriptor table.


ItsLemmy:
Thanks!

PR #4034, opened 2026-08-20
feat(calendar): add vdirsyncer and local vdir calendar support 

xela-zone:
## Summary
Adds support for local `vdir` and `vdirsyncer` calendar directories to Noctalia's calendar service:
 - **`vdir_reader` module**: Depth-bounded recursive collection discovery capable of resolving nested folder hierarchies (e.g. `~/.
  local/share/calendars/<account>/<collection>/`).
 - **Metadata parsing**: Reads standard vdir metadata files (`displayname`, `color`, `order`) and `X-WR-CALNAME` properties with fallback to folder names.
 - **Inotify event monitoring**: Watches discovered collection directories for `.ics` file creations, modifications, and deletions with a 300ms debounce
  directly on the main poll loop.
 - **Settings GUI**: Added a `Local (vdir)` provider option and directory path picker/validator in **Settings → Services → Calendar**.
- **Unit tests**: Added unit tests in **`tests/vdir_reader_test.cpp`.**
## Motivation

I personally already have my caldav synced using vdirsync, to allow one syncing tool to work with multiple tools, such as khard or even someone using a local only calender app. 

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Build / packaging

## Testing
- Added unit tests in `tests/vdir_reader_test.cpp` covering:
  - **Nested Collection Discovery**: Recursive traversal across multi-level directory trees (e.g. `<account>/<collection-uuid>/`).
  - **Direct Collection Discovery**: Resolving accounts pointed directly at a single collection folder.
  - **Directory Filtering**: Ignoring hidden directories (e.g. `.git/`), temporary directories (`.tmp`), and non-directory files.
  - **Metadata File Parsing**: Correct extraction and trimming of `displayname`, `color`, and `order` files, plus collection sorting by `order`.
  - **`X-WR-CALNAME` Fallback**: Extracting calendar titles directly from `.ics` file headers when no `displayname` file is provided.
  - **Event Loading & Mapping**: Parsing `.ics` events from discovered collections with proper mapping of IDs, summaries, calendar names, and color
assignments.

## Manual Coverage

- [x] Tested on Niri
- [ ] Tested on Hyprland
- [ ] Tested on Sway

## Screenshots / Videos

image
after creation / edit panel: 
image

## Checklist

## Additional Notes

written by gemini / antigravity, reviewed and lead by a human


ItsLemmy:
## Blocking                                                                                                                                                                             

### 1. vdir / local dual account type                                                                                                                      

account.type is accepted as either string in five places (calendar_service.cpp:413,668, config_schema.cpp:629, settings_window_popups.cpp:1033,1049,  settings_window_scene.cpp:1882).

Nothing in the tree ever writes "local" — the settings UI emits "vdir" unconditionally (popups.cpp:1526). It is a pure alias.                       

Per project convention                                                                                                                                                                               
│ "Every config key, IPC name, path, and identifier has exactly one canonical name. Never add aliases…"                                                                              

Delete "local" everywhere. Pick vdir.                                                                                                                                                

### 2. Shared ICalParseControl silently truncates events                                                                                                                             

fetchVdir creates one ICalParseControl (remainingRecurrenceWork = 100'000) and threads it through every collection and every file. loadVdirCollectionEvents then discards            
result.status entirely and accountDone(account.id, true, …) reports success.                                                                                                         

Measured against the PR branch (400 .ics files, RRULE:FREQ=DAILY;COUNT=400, 3-year window):                                                                                          

```                                                                                                                                                                                  
shared budget:    100,000 events   remaining=0
per-file budget:  160,000 events   (0 files reported WorkBudgetExceeded)
```                                                                                                                                                                                  

37.5% of events dropped, no log line, account reported healthy. fetchIcs treats WorkBudgetExceeded as a hard failure with kLog.warn (calendar_service.cpp:621); vdir gets neither.   

This is the common case, not a corner: vdir is one-item-per-file by specification, so file counts are high by design. Give each file a fresh control (matching the per-response semantics every other provider gets) and propagate status.                                                                                                                           

### 3. Trailing slash in the configured path ⇒ zero events, silently                                                                                                                 

Verified by running discoverVdirCollections from the branch:                                                                                                                         

```                                                                                                                                                                                  
root='/tmp/noct_probe'    n=1  id='noct_probe'
root='/tmp/noct_probe/'   n=1  id=''
```                                                                                                                                                                                  

buildCollection uses dirPath.filename(), which is empty for a trailing separator. Downstream, serializeCalendarSources skips empty ids (calendar_discovery_state.cpp:18) and         
selectedCalendarSourceIds skips them too (:62), so std::erase_if removes the collection and accountDone(id, true, {}) reports success with no events and no warning.                 

The settings field placeholder is literally ~/.local/share/calendars; users type trailing slashes. Normalise the root before buildCollection, and never emit an empty collection id. 

### 4. Synchronous filesystem + libical work on the main loop                                                                                                                        

Every other provider is async through HttpClient callbacks. fetchVdir walks the tree and parses every file inline inside startRefresh().                                             

Measured at -O2 (the debug build is materially worse), 3,000 files across 3 collections:                                                                                             

```                                                                                                                                                                                  
discover=4ms  load=101ms
```                                                                                                                                                                                  

~6 dropped frames at 60 Hz, on every refresh cycle and again after each 300 ms inotify debounce while vdirsyncer writes. Compounding it, updateVdirWatches() is called from inside   
fetchVdir() (:197) and re-runs discoverVdirCollections on every root — the recursive walk happens twice per refresh.                                                                 

Direction: mirror CalDavClient::State (caldav_client.cpp:69-186) — a single std::jthread queue that discovers, reads and parses off the main thread and marshals the result back via DeferredCall::callLater, guarded by weak_from_this(). Pass the account data in by value (a config reload will reallocate m_activeConfig.accounts under an in-flight job) and  have the worker return the discovered collection paths alongside the events, so the main thread only issues inotify_add_watch/unwatch and never walks the tree itself. 
 
That  removes the stall and the duplicate walk together, gives 5 the path set it needs to reconcile against, and puts the per-file ICalParseControl of #2 and the root normalisation of 3 in the same worker body. fetchVdir then returns without calling accountDone, exactly like fetchCalDav/fetchIcs, so the refresh state machine needs no change.


## Should fix                                                                                                                                                                           

5. inotify watches are add-only. m_watchedVdirPaths never shrinks and Inotify::unwatch is never called. Removing a vdir account, editing its path, or disabling the calendar leaves  
the watches live for the process lifetime, so unrelated directory churn keeps waking the poll loop and firing notifyChanged(). ConfigService::refreshIncludeWatches() (config_service.cpp:1117-1175) is the repo's reconciliation pattern — diff desired against current, unwatch the difference. Mirror it.                                               

Related: no IN_DELETE_SELF / IN_MOVE_SELF. When a collection directory is deleted and recreated (vdirsyncer collection reset) the kernel drops the wd, but the path stays in m_watchedVdirPaths, so the guard at :172 prevents it from ever being re-watched.                                                                                                     

6. Premature notifyChanged(). dispatchPoll notifies all UI callbacks the moment an inotify batch arrives, before any data has been re-read; requestRefresh() notifies again 300 ms later. Drop the first.                                                                                                                                                               

7. Unrelated behavior change in the save path. m_calendarService->requestRefresh() (popups.cpp:1556) sits in the !caldav branch, so it also fires for Google and ICS. It is redundant — onConfigReload() already does m_nextRefreshAt = now (calendar_service.cpp:314) — and it runs before the override write has been reloaded, scheduling a refresh against the stale account list. Remove it.                                                                                                                                                   

8. pathInvalid is dead code. Declared, reset, and tested in the aggregate invalid check, but never assigned true. The PR description claims a "directory path picker/validator";  there is no validation. Empty path silently resolves to defaultVdirPath(), and a nonexistent path only produces a kLog.warn at refresh time with no UI signal. Either validate (non-empty, exists, is a directory) or drop the flag.                                                                                                                                                                                                

## Nits                                                                                                                                                                                 

- en.json ordering. The catalog is alphabetically sorted; "vdir-path-label" is wedged between "provider-label" and "save". It belongs after "username-placeholder".                  
- std::stoi + catch (...) for the order file — std::from_chars avoids exceptions-as-control-flow; a malformed order silently becomes 0.                                              
- Redundant directory scans. buildCollection re-iterates the directory hasIcsFiles just scanned, and the X-WR-CALNAME fallback opens + 4 KiB-reads files on every discovery pass.    
- Depth pruning. it.depth() > maxDepth + it.pop() descends one level past the limit before backing out; disable_recursion_pending() at depth() == maxDepth is idiomatic. Also pop()  
is the throwing overload inside an otherwise ec-based loop.                                                                                                                        
- Missing <cctype> for std::isxdigit (compiles transitively; the repo's clang-tidy set omits misc-include-cleaner, so it isn't caught).                                              
- Test hygiene. Fixed paths under temp_directory_path() (noctalia_vdir_test_nested) collide across concurrent runs and leave debris on failure. meson.build also places              
'vdir_reader' after 'wallpaper_shuffle_state', out of the list's order.                                                                                                            
- ICS → vdir provider switch on an existing account leaves the stale server_url override in place, which the new finalize turns into a diag.error. This class of bug predates the PR 
(caldav → google leaves username), but the new rule makes the adjacent-provider switch hit it. Consider clearing keys the target provider does not own.

ItsLemmy:
Thanks!

Issue #2859, opened 2026-05-31
[BUG] v5 Tray widget not rendering icons on restart of noctalia


xircon:
### Submission checklist

- [x] I have searched existing issues and confirmed this is not a duplicate.
- [x] I am using the latest available version of Noctalia.

### Bug description

After a rebuild if I restart (`killall noctalia && noctalia& disown`) some of the icons appear as three horizontal lines:

<img width="576" height="142" alt="Image" src="https://github.com/user-attachments/assets/ced4c150-d993-48aa-a17e-8c293d999bbe" />

### Steps to reproduce

Update, rebuild.
Restart by killing process.

### Expected behavior

Icons display correctly. 

### Actual behavior

As above image.

### Logs / error output

```text

```

### Noctalia version

noctalia v5.0.0 (8346d3ce5d65)

### Compositor

Mango

### Distribution

Arch-based

### Environment information

```text
Aur.
```

ItsLemmy:
What are the apps that don't recover?

xircon:
arch-update
vorta
vesktop
flameshot

Those are the ones not showing at this moment,  but it does vary and occasionally an icon will recover:

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/0b6ea8a6-4615-4782-8326-1d5281d76b2d" />

Ly-sec:
We did a lot of stuff for tray etc, can you please check if this issue still persists? @xircon

xircon:
A lot better,  only vorta misbehaves and not all of the time.

Ly-sec:
> A lot better, only vorta misbehaves and not all of the time.

Cool, sadly there isn't much more we can do given some apps just don't try to re-register if a tray is available, guess that one isn't one of them. I shall close this one as completed. 

//edit: ignore what I said there :p I shall investigate the coming days

//edit2: @xircon I just tested vorta and it behaves fine on my end after a noctalia restart

PR #4061, opened 2026-08-22
fix(nix): wrap gsettings-desktop-schemas

nicobonada:
## Summary

Prefix `gsettings-desktop-schemas` onto `XDG_DATA_DIRS` in the existing
`wrapProgram` (same pattern as `git` on `PATH`).

## Motivation

Tray `IconName` lookup uses GSettings (`org.gnome.desktop.interface`).
glib does not ship that schema. On NixOS without GNOME it is also not
on the session profile, so IconResolver only searches hicolor.

Apps that advertise a themed name and no pixmap (e.g. Solaar
`battery-090`) then miss and the tray falls back to the `menu-2` glyph.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [x] Build / packaging

## Related Issue

None open that I could find. Closest prior report is #2859 (hamburger
after a restart); that was SNI re-registration, not schema lookup.

## Testing

- `nix build .#default`
- Wrapper prefixes
  `…/gsettings-desktop-schemas-*/share/gsettings-schemas/…`
- `gsettings get org.gnome.desktop.interface icon-theme` is
  `No schemas installed` without the prefix, `'Papirus-Dark'` with it
- Replaced the running user unit with the built binary on niri: Solaar
  still published `IconName=battery-090`, tray showed the Papirus
  battery glyph instead of `menu-2`

## Manual Coverage

- [x] Tested on Niri
- [ ] Tested on Hyprland
- [ ] Tested on Sway

## Screenshots / Videos

Happy to attach a before/after of the tray (hamburger → battery).

## Checklist

## Additional Notes

nixpkgs GNOME packaging (`doc/languages-frameworks/gnome.section.md`)
would normally use `wrapGAppsNoGuiHook` so the glib setup hook can put
`GSETTINGS_SCHEMAS_PATH` on `XDG_DATA_DIRS`.

This package already `wrapProgram`s for `git`. Using that hook here
means `dontWrapGApps` plus splicing `gappsWrapperArgs` into the existing
wrapper, and it also injects `$out/share`, default `/usr/share`, and
`dconf.lib` on `GIO_EXTRA_MODULES`. One explicit `--prefix XDG_DATA_DIRS`
is the single path we need, same style as the git wrap.

### DIFF
--- a/nix/package.nix
+++ b/nix/package.nix
@@ -41,6 +41,7 @@
   jemalloc,
   makeWrapper,
   git,
+  gsettings-desktop-schemas,
   autoAddDriverRunpath,
   # DEPRECATED: no longer affects the build; kept for `.override` compat.
   cudaSupport ? config.cudaSupport,
@@ -68,7 +69,8 @@ lib.warnIf cudaSupport
 
   postFixup = ''
     wrapProgram $out/bin/noctalia \
-      --prefix PATH : ${lib.makeBinPath [ git ]}
+      --prefix PATH : ${lib.makeBinPath [ git ]} \
+      --prefix XDG_DATA_DIRS : "${gsettings-desktop-schemas}/share/gsettings-schemas/${gsettings-desktop-schemas.name}"
 
     $out/bin/noctalia completions bash | install -D /dev/stdin $out/share/bash-completion/completions/noctalia
     $out/bin/noctalia completions zsh  | install -D /dev/stdin $out/share/zsh/site-functions/_noctalia
--- a/nix/package.nix
+++ b/nix/package.nix
@@ -70,7 +70,7 @@ lib.warnIf cudaSupport
   postFixup = ''
     wrapProgram $out/bin/noctalia \
       --prefix PATH : ${lib.makeBinPath [ git ]} \
-      --prefix XDG_DATA_DIRS : "${gsettings-desktop-schemas}/share/gsettings-schemas/${gsettings-desktop-schemas.name}"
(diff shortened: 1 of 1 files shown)


nix/package.nix
```diff
@@ -68,7 +69,8 @@ lib.warnIf cudaSupport
 
   postFixup = ''
     wrapProgram $out/bin/noctalia \
-      --prefix PATH : ${lib.makeBinPath [ git ]}
+      --prefix PATH : ${lib.makeBinPath [ git ]} \
+      --prefix XDG_DATA_DIRS : "${gsettings-desktop-schemas}/share/gsettings-schemas/${gsettings-desktop-schemas.name}"
```
samiser: `glib` is already available so you can do this to get the path without re-encoding it
```suggestion
      --prefix XDG_DATA_DIRS : "${glib.getSchemaDataDirPath gsettings-desktop-schemas}"
```

samiser (commented):
lgtm! one minor nit but otherwise fine. i'd also suggest contributing this upstream to the nixpkgs noctalia package because the same issue exists there too. thanks!

nicobonada:
> i'd also suggest contributing this upstream to the nixpkgs noctalia package because the same issue exists there too. thanks!

Unfortunately, the work was mostly done through a coding agent and current nixpkgs AI policy wouldn't allow me to send the fix. I would prefer if someone with experience contributing to nixpkgs would handle that part.

samiser:
@nicobonada its fine as long as you disclose it in the pr, and i'm the maintainer so i can review and approve it :) but if you're really opposed i can make the change

ItsLemmy:
Thanks

PR #4069, opened 2026-08-22
fix(ui): clamp drag previews to overlay bounds

Radium000:
## Summary

- Keep drag previews fully inside the overlay bounds.
- Update both coordinates independently when the pointer moves beyond an edge.
- Account for the preview scale when calculating its allowed position.
- Add regression tests for edge clamping and movement parallel to an edge.

## Motivation

When a drag preview reached an overlay boundary, it could either stop moving
completely or remain partially outside the overlay.

`Node::mapFromScene()` still calculates the local coordinates when the pointer
is outside the overlay, but returns `false` because the mapped point is not
contained by the node. `DragDropController::updatePreview()` previously
discarded the entire update in that case.

This also prevented movement along the unaffected axis. For example, after
reaching the left edge, moving the pointer vertically no longer moved the
preview.

The new behavior uses the calculated coordinates and clamps each axis
independently. The preview therefore remains fully visible while continuing
to follow pointer movement parallel to the overlay edge.

## Type of Change

- [x] Bug fix

## Related Issue

N/A — reproduced and verified locally against Noctalia v5.

## Testing

- Ran `just format` with clang-format 22.1.8.
- Ran `just test debug --print-errorlogs`: 91/91 tests passed.
- Added regression tests for clamping and movement along overlay edges.
- Manually verified on Hyprland.

### DIFF
--- a/src/ui/drag_drop_controller.cpp
+++ b/src/ui/drag_drop_controller.cpp
@@ -48,6 +48,18 @@ namespace {
     return safe;
   }
 
+  float
+  clampPreviewPosition(float position, float previewSize, float previewScale, float transformOrigin, float limit) {
+    const float scaledStart = transformOrigin * (1.0F - previewScale);
+    const float scaledEnd = scaledStart + previewSize * previewScale;
+    const float minimum = -scaledStart;
+    const float maximum = limit - scaledEnd;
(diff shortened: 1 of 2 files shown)


ItsLemmy:
Thank you

PR #4079, opened 2026-08-24
feat(calendar): notify before events start

yevhen-sychov:
## Summary

Adds event reminder notifications to the calendar service. Reminder times are read from the calendar itself — `VALARM` triggers on CalDAV, ICS, and local vdir events, reminder overrides or calendar defaults on Google — with a configurable fallback lead for events that carry none. All-day events are collected into one daily digest instead of notifying individually.

The notification body counts down as the event approaches and settles on "Starts now" at the start. Reminders are kept in notification history, so one missed while you were away is still in the control center.

```toml
[calendar.reminders]
enabled              = true
use_event_reminders  = true     # honor per-event reminders; off = always use the default lead below
default_lead_minutes = 10       # used for events with no reminder of their own; 0 = at event start
all_day_digest_time  = "09:00"  # one notification listing today's all-day events; "" disables it
```

The same four options are in Settings → Services → Calendar. Reminders require `[calendar]` itself to be enabled, which is opt-in; once calendars are on they default to enabled with a 10-minute lead.

## Motivation

Noctalia syncs calendars and displays events, but never tells you one is about to start — so the calendar is useful for looking things up and useless for not missing them. Both halves of a reminder were absent: the data was discarded at parse time (`ical_parser.h`: *"VTODO/VALARM and others are ignored"*, Google's `reminders`/`defaultReminders` never read), and nothing scheduled an alert.

## Type of Change

- [x] New feature

## Testing

`just build` and `just test` — 116/116 passing.

Four new test binaries (`calendar_reminders`, `calendar_reminder_monitor`, `google_client_reminders`, `notification_internal_history`) plus new cases in `ical_parser_test` and `config_schema_roundtrip_test`.

`calendar_reminder_monitor_test` drives the real monitor against a real `NotificationManager` to show the countdown state machine terminates: a counting reminder arms exactly one poll deadline, and once it settles on "starts now" or is dismissed the monitor advertises no deadline at all. There are no threads and no timer objects in this feature, so a leak would surface there as a deadline that never goes away or never moves forward.

Cases worth calling out, since each guards a rule that is easy to undo by accident:

- `TRIGGER:PT0S` must survive as lead 0 rather than being discarded as a null trigger
- an absolute trigger must not be copied onto every instance of a recurring series
- the dedup key must not change when the default lead is reconfigured, or a fired event re-notifies
- each countdown step must land on a whole minute before the event, and the sequence must terminate

Manually verified against live CalDAV, Google, and ICS accounts: per-event lead, fallback lead, Google `useDefault`, recurring events, the all-day digest firing once per day, catch-up across a restart between fire time and event start, and already-started events staying silent.

## Manual Coverage

- [ ] Tested on Niri
- [x] Tested on Hyprland
- [ ] Tested on Sway

## Checklist

### Architecture

A pure planner plus a thin stateful monitor. The split is what makes this testable with no `NotificationManager`, `ConfigService`, or D-Bus, matching how `google_calendar_list.h` is tested today.

| Unit | Role |
|---|---|
| `calendar/calendar_reminders.{h,cpp}` | pure: `normalizeReminderLeads`, `reminderKey`, `digestInstantFor`, `countdownMinutes`, `planReminders` |
| `calendar/google_reminders.h` | header-only Google payload parsing, testable without an `HttpClient` |
| `calendar/calendar_reminder_monitor.{h,cpp}` | notifications, state persistence, poll deadlines |
| `calendar/calendar_reminder_poll_source.h` | mirrors `calendar_poll_source.h` |

`CalendarEvent` gains `std::vector<std::int32_t> reminderLeadSeconds` — lead *offsets*, not absolute times. `expandRecurrences` builds each occurrence by copying `RecurrenceData::base` and overwriting only `start`/`end`, so relative leads propagate to every recurrence instance for free. It also matches Google's native "minutes before start" model.

### Notes for review

**`src/calendar/ical_parser.cpp`**

- The VALARM walk sits in `baseEventFromComponent`, so non-recurring events get it directly and recurring ones inherit it through `RecurrenceData::base`.
- Gated on `icaltriggertype_is_bad_trigger`, **not** `is_null_trigger` — the latter is true for the perfectly valid `TRIGGER:PT0S` ("alert at start"), which must yield lead 0.
- `ACTION:EMAIL` is rejected (the server delivers it; honoring it double-notifies) as is `PROCEDURE` (script execution). `DISPLAY`, `AUDIO`, and a missing ACTION are honored.
- **Absolute triggers are skipped for recurring events.** RFC 5545 §3.8.6.3 makes an absolute trigger fire once for the whole series; copying its offset onto 300 expanded instances would be flatly wrong.
- `REPEAT`/`DURATION` inside a VALARM are ignored — a snooze ladder with no dismissal semantics to stop it.
- Drive-by: `baseEventFromComponent` was called twice per component, the first time only to read `event.id`. That call is now a direct `icalcomponent_get_uid`, so the VALARM walk isn't done twice.

**`src/calendar/google_client.cpp`** — no `CalendarMeta` plumbing and no query-string change: the `events.list` response already carries a top-level `defaultReminders` for the calendar being queried, in the same payload `parseEventsInto` already parses, and the request uses no `fields=` mask.

**`src/calendar/calendar_reminders.cpp`**

- The firing rule is the whole thing: `due <= now && event.start > now && !fired.contains(key)`. That one line gives in-session firing, startup catch-up, and silent drop of already-started events.
- Events using the configured default record a `kDefaultLeadSentinel` in their dedup key rather than the resolved lead. **This is load-bearing:** embedding the resolved value would mint a fresh key when the lead is reconfigured from 10 to 15, and re-notify for an event that already fired.
- Recurring instances share a UID and are separated by start time in the key. Two accounts subscribed to the same UID deliberately collapse to one notification.
- `countdownMinutes` rounds **up**, so the label means "starts in at most N minutes". Truncating would read "in 9 min" for a ten-minute lead, since a reminder necessarily fires a hair after its due instant. Rounding to nearest would put every step on the half minute, so "3 min" appears 3½ minutes out and disagrees with a glance at the clock for half of each minute. Rounding up anchors every step to a whole minute before the event and reaches "starts now" exactly at the start.
- The scan uses `lower_bound` on the sorted snapshot and stops at `now + kMaxLeadSeconds + 24h`, so cost is proportional to the armed window, not to calendar size.

**`src/calendar/calendar_reminder_monitor.cpp`**

- **The fired set is bounded by an invariant, not a cap:** a fired key becomes permanently irrelevant the moment its event starts, because the rule requires `event.start > now`. Pruning on start leaves residency at "fired but not yet started" — a handful in practice. A bounded-horizon watermark was considered and rejected: accept a 10:20 invite at 10:00 carrying a 30-minute reminder and its 09:50 due time is already behind the watermark, silently swallowing a reminder for an event that has not started.
- The poll source returns `-1` when disabled or unarmed, so an idle shell adds **zero** wakeups. When armed it clamps to 5 minutes — `poll()` runs on `CLOCK_MONOTONIC`, which does not advance across suspend, so a 3-hour deadline would otherwise fire 7 hours of wall time later across a 4-hour sleep. No steady_clock deadline is ever stored; the delay is recomputed from wall time on every call.
- The digest uses `to_sys(local_time, choose::earliest)`. The bare overload throws `nonexistent_local_time` on a spring-forward gap, so a 02:30 digest would throw once a year.
- The digest date is **not** recorded when there were no all-day events to show, or a sync landing at 09:05 would be suppressed for the day.
- Disabling reminders does not `clearOwner` on the persisted state — re-enabling mid-day should not replay the day.

**`src/notification/`**

- `Notification`/`NotificationRequest` gain `persistInHistory`. Internal notifications are toast-only today, which is right for every existing one — battery-low re-fires, Google connect failures resurface in Settings, config migration re-arms on a timer. A calendar reminder is the first whose entire value is a one-shot moment with no other surface. Hence an explicit opt-in field rather than a blanket `origin == External` change; filters still apply, so the reminder's `appName` stays filterable.
- `NotificationManager::updateBody` rewrites a live notification's text in place. `addOrReplace(replacesId)` was the obvious alternative and is wrong here: with a `replacesId` no longer in `m_idToIndex` it falls through and **creates a new notification**, so the first countdown tick after a dismissal would resurrect the toast. `updateBody` also leaves the expiry, the history entry's position, and its `seen` flag alone — a body that rewrites itself once a minute is not a new notification to read.
- `setInternalActionCallback`: actions on internally generated notifications are handled in-process, since there is no D-Bus client to receive `ActionInvoked`.

### Known limitations

- Snooze is out of scope. The hook points (`NotificationRequest::actions` plus `setActionInvokeCallback`) are noted in the header.
- Declined invitations still raise reminders. RSVP status is not parsed anywhere in the calendar service today — no `PARTSTAT`, `ATTENDEE`, or `responseStatus` handling and no field on `CalendarEvent` — so filtering them is a separate change rather than a line in this one.


ItsLemmy:
Thanks for the PR, made a few minor adjustments. I'll merge it and daily drive it for a few days to see how it behaves.

PR #4347, opened 2026-09-08
Live audio-reactive wallpaper

weissi1994:
## Summary

Adds an optional **live paper** wallpaper mode: a libprojectM 4 (Milkdrop) audio-reactive
visualizer that replaces the static image on every wallpaper output and on the lock surface.

- **Renderer** - `src/render/visualizer/projectm_renderer.{h,cpp}` drives one libprojectM
  instance into a shared offscreen texture. `renderFrame()` makes the surfaceless root
  context current against a private hidden Wayland surface, renders one frame, and restores
  the caller's context; wallpaper and lock surfaces (separate contexts in the same share
  group) sample it through an **EGLImage**.
- **Audio** - `src/pipewire/pipewire_pcm_tap.{h,cpp}`, a raw-PCM sibling of the existing
  spectrum tap. It follows the active sink monitor by default. Microphone fallback exists but
  is **off by default** behind `allow_mic_fallback`.
- **Service** - `src/shell/wallpaper/visualizer_service.{h,cpp}` owns preset scanning,
  rotation, and the renderer lifecycle.
- **Config** - new `[wallpaper.live_paper]` table (`enabled`, `interval_seconds`, `fps`,
  `mesh_w`/`mesh_h`, `render_width`/`render_height`, `darken`, `presets_dir`, `audio_source`,
  `allow_mic_fallback`), documented in `example.toml` with range-validated schema fields.
- **IPC** - `livepaper-enable`, `livepaper-disable`, `livepaper-toggle`, `livepaper-next`.
- **Build gate** - Meson feature option `livepaper` (default `auto`). Meson *link-tests*
  libprojectM during configure, so a broken `projectM-4.pc` is caught there instead of at
  final link. Without a usable libprojectM the feature compiles out via
  `livepaper_stub.cpp`, the binary links normally, and the settings entry explains why the
  visualizer is unavailable.
- **Nix** - a `libprojectm-gles` overlay in `nix/package.nix` builds libprojectM with
  `-DENABLE_GLES=ON` and patches the pkg-config file; `nix/home-module.nix` stages presets.
- **Docs** - a Live Paper section in `docs/user/desktop/wallpaper.mdx` (config, presets,
  resolution/cost, audio and privacy) with the four IPC commands added to the
  `Media & UI` IPC table; `BUILDING.md` and `PACKAGING.md` for the two ways a distro
  libprojectM goes wrong.
- **Tests** - `[wallpaper.live_paper]` added to the config schema round-trip probe and new
  range-clamp goldens in `tests/config_schema_roundtrip_test.cpp`.

Off by default at both build time (`auto`) and runtime (`enabled = false`).

## Motivation

Noctalia can only paint a static image as the wallpaper. Animated, audio-reactive wallpapers
are a long-standing want, and Milkdrop/projectM is the established open format for them, with
a large body of existing presets users already have.

Doing this well on the C++/Wayland/EGL stack is not just "draw into the wallpaper surface":
libprojectM needs its own current GL context, the result must be shared across per-output
contexts and the lock surface without a per-output projectM instance, and libprojectM is an
awkward dependency to require (usually built for desktop GL rather than GLES, and shipping a
`.pc` file that names a library filename that does not exist). This PR solves those and keeps
the whole thing optional, so nothing changes for users and packagers who do not opt in.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [x] Build / packaging

## Testing

Build and test runs are on NixOS via `nix develop`, which supplies the `libprojectm-gles`
overlay.

- `just format` (clang-format v22+) - clean.
- **libprojectM build** (`build-debug`, `livepaper` resolved to enabled) - builds clean. No
  new compiler warnings from any live-paper code: rebuilt
  `projectm_renderer.cpp`, `livepaper_availability.cpp`, `pipewire_pcm_tap.cpp`,
  `visualizer_service.cpp`, `wallpaper.cpp`, and `config_schema.cpp` from scratch
- **Stub build** (`build-nolivepaper`, no `NOCTALIA_HAVE_LIVEPAPER`) - builds and links, and
  the config schema tests pass there too, so `[wallpaper.live_paper]` parses identically
  whether or not the feature is compiled in.
- `meson test -C build-debug` - **110/112 pass**. The two that do not are `process`
  (`completion-only async command stdout was wrong`) and `plugin_git_export` (30s timeout);
  both are unrelated to this PR - neither `src/core/process` nor the plugin-export code is
  in this diff.
- The new checks were mutation-tested rather than just observed green: widening the
  `fps` range floor makes the clamp golden fail, and dropping a field from
  `livePaperSchema()` makes the wallpaper round-trip fail. Both pass again once reverted.

Manual: enabled live paper on the desktop and locked the session to confirm the lock surface
picks up the same texture; toggled via `livepaper-toggle` and stepped presets with
`livepaper-next`; confirmed playback-driven reactivity with `allow_mic_fallback = false`.

## Manual Coverage

- [ ] Tested on Niri
- [ ] Tested on Hyprland
- [ ] Tested on Sway
- [x] Tested on another compositor: scroll (a sway fork)

## Screenshots / Videos

https://github.com/user-attachments/assets/f1fc83b8-dade-497a-92eb-6aad46c050f8

## Checklist

## Additional Notes

**Test coverage.** `tests/config_schema_roundtrip_test.cpp` now covers `[wallpaper.live_paper]`
two ways, both following patterns already in that file:

- Every field is set off-default in the shared probe config, so the registry-driven
  read-inverse check catches a lossy codec or an asymmetric read/write key in
  `livePaperSchema()`. Without this the sub-table sat at defaults and a dropped field would
  have compared equal to itself.
- Clamp goldens pin the parse-time ranges for `fps`, `mesh_w`/`mesh_h`, `render_width`/
  `render_height`, and `darken`, plus one regression guard that `interval_seconds = 0`
  survives as "no timed rotation" rather than being floored to 1.

**Reviewer context:**

- The GL-context discipline is the subtle part. Every `projectm_*` call that touches GL -
  including `loadPreset()`, not only `initialize()` and `renderFrame()` - must run with the
  shared surfaceless root context current, because VAOs are container objects and are not
  shared across an EGL share group. Two further constraints shape the renderer, both
  explained in comments at the relevant code: libprojectM 4.1.x ignores any externally bound
  FBO and composites to draw framebuffer 0 (hence the private, never-committed `wl_surface`
  that gives the producer a real default framebuffer), and cross-context sampling goes
  through an `EGLImageKHR` alias rather than a raw shared texture name.
- `render_width`/`render_height` are a quality/GPU-cost knob, not a per-output property: one
  framebuffer feeds every output and the lock surface, scaled by each wallpaper's `fillMode`.
- Microphone fallback is a privacy-relevant path and is deliberately opt-in
  (`allow_mic_fallback = false`), documented as such in `example.toml` and in the user docs.
- **Presets.** Nothing is bundled in the tarball; users point `presets_dir` at their own, or
  drop them in `$XDG_DATA_HOME/waylivepaper/presets`. The flake's Home Manager module is the
  exception: it symlinks a pre-built pack (`presets-cream-of-the-crop` with the brightest and
  most strobe-heavy presets filtered out, from the `presets-photosensitive-filtered` flake
  input) as soon as `settings.wallpaper.live_paper.enabled` is set, so enabling the
  visualizer is a single switch. `defaultPresets = false` opts out.
- Milkdrop presets can flash and strobe. The Nix pack is filtered with that in mind, but
  user-supplied packs are not, and there is no runtime filtering - the user docs carry a
  photosensitivity caution alongside the `darken` knob.


nix/package.nix
```diff
@@ -61,77 +77,79 @@ let
...
-  };
-}
+    meta = with lib; {
+      description = "A sleek, customizable desktop shell crafted for Wayland.";
+      homepage = "https://github.com/noctalia-dev/noctalia";
+      license = licenses.mit;
+      platforms = platforms.linux;
+      mainProgram = "noctalia";
+    };
+  }
```
SollyBunny: These changes are unrelated

SollyBunny (commented):
Re-implementing ProjectM is the worst way to do this
If you want this functionality you can use the wl-paper example app from https://github.com/mahkoh/wl-proxy/ to embed any app including the existing ProjectM implementation
Noctalia plugins then can offer integration
Re-implementation leads to bugs and maintenance cost for no gain, ProjectM has a library version (libprojectm) that can be vendored/built againt if the stock app isn't suitable
There are many options for live wallpapers and ProjectM doesn't cover a good chunk of them (an example being a video loop)

PR #4443, opened 2026-09-15
feat(audio): implement sound themes

dchinmay2:
## Summary

This change removes individually configurable shell sounds and the bundled shell sounds. It adds a minimal implementation of the [Freedesktop sound theme specification](https://specifications.freedesktop.org/sound-theme/latest-single/) with a dropdown to select from installed sound themes.

## Motivation

This opens the door to adding charger plugged/unplugged and screenshot sounds, and possibly even more events, as bundling sounds for every event and having configurable paths for every event sound is not sustainable.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [x] Breaking change
- [ ] Refactoring
- [ ] Build / packaging

## Testing

Manually tested with the freedesktop, cosmic, elementary, and macos-bigsur-sound-theme themes.

## Manual Coverage

- [x] Tested on Niri
- [ ] Tested on Hyprland
- [ ] Tested on Sway

## Screenshots / Videos
image

## Checklist


NickHastings:
Great, I never felt comfortable packaging those wav files.

dchinmay2:
Thanks for the reminder, added the runtime dependency to PACKAGING.md

src/pipewire/sound_player.cpp
```diff
@@ -1,22 +1,133 @@
...
+    if (const char* dataDirsEnv = std::getenv("XDG_DATA_DIRS"); dataDirsEnv != nullptr && *dataDirsEnv != '\0') {
+      for (const auto directory : std::string_view(dataDirsEnv) | std::views::split(':')) {
+        std::string_view piece(directory);
+        if (!piece.empty()) {
+          dataDirs.emplace_back(fs::path(piece) / "sounds");
+        }
+      }
+    } else {
+      dataDirs.emplace_back("/usr/local/share/sounds");
+      dataDirs.emplace_back("/usr/share/sounds");
```
EbonJaeger, line 54: Should this use the configured build prefix instead of hardcoding `/usr` or `/usr/local`?

NickHastings: I don't think so, `dataDirs` seems to be the locations that it looks for the files from the sound theme (not provided by noctalia).

dchinmay2: correct

ItsLemmy:
### Blocking-worthy finding                        
                                                                                                           
 soundBaseDirs() omits the user data home — src/pipewire/sound_player.cpp:43-57. It searches only          
 $XDG_DATA_DIRS/sounds (or /usr/local/share/sounds + /usr/share/sounds when unset). The Freedesktop        
 sound-theme spec (mirroring the icon-theme base-dir spec) searches $XDG_DATA_HOME/sounds first, then the  
 system dirs. Consequence: a theme installed to ~/.local/share/sounds/<theme> is invisible to both         
 availableThemes() (won't appear in the dropdown) and findThemeSound() (won't play) even if selected. This 
 is the most common per-user install location.                                                             
                                                                                                           
 This codebase already establishes the correct base-dir pattern elsewhere (icon_resolver.cpp:127,          
 desktop_entry.cpp:31): resolve $XDG_DATA_HOME, defaulting to ~/.local/share, ahead of $XDG_DATA_DIRS.     
 Suggested fix, prepended in soundBaseDirs():                                                              
                                                                                                           
 ```cpp                                                                                                    
if (const char* home = std::getenv("XDG_DATA_HOME"); home != nullptr && *home != '\0') {
  dataDirs.emplace_back(fs::path(home) / "sounds");
} else if (const char* userHome = std::getenv("HOME"); userHome != nullptr && *userHome != '\0') {
  dataDirs.emplace_back(fs::path(userHome) / ".local/share/sounds");
}
 ```                                                                                                       
                                                                         
 ### Nits                                                                             
                                                                                                           
 - settings_registry.cpp:2716 hardcodes the empty-catalog fallback label "Default" (untranslated).         
   Consistent with the theme Name labels also being untranslated, so low value; no existing tr key. Left   
   it.                                                                                                     
 - Redundant default chain: AudioConfig::soundTheme defaults to "freedesktop", the schema default is       
   "freedesktop", setTheme defaults to "freedesktop", and application_services.cpp guards                  
   soundTheme.empty() ? "freedesktop" : .... Harmless, but the empty-guard only fires if a user writes     
   sound_theme = "" explicitly. Defensible; not touched.                                                   
 - Constructor SoundPlayer(loop) calls setTheme("freedesktop"), then applySoundConfig() immediately calls  
   setTheme(...) again at startup — one redundant decode of the default theme. Trivial.                    
 - A theme file present as <event>.disabled makes findThemeSound return nullopt, which setTheme then logs  
   as an error("... missing sound ...") rather than an intentional disable. Cosmetic log-level nit.

ItsLemmy:
This was the automated review, if you are confident this is the best order for theme discoveries I can merge as is. But it feels a little odd to me.

dchinmay2:
>  The Freedesktop sound-theme spec (mirroring the icon-theme base-dir spec) searches $XDG_DATA_HOME/sounds first, then the

https://specifications.freedesktop.org/sound-theme/latest-single doesn't mention XDG_DATA_HOME at all. But the base directory spec says:
> [`$XDG_DATA_DIRS` defines the preference-ordered set of base directories to search for data files in addition to the `$XDG_DATA_HOME` base directory. The directories in `$XDG_DATA_DIRS` should be separated with the separator used for `$PATH` on the platform (typically this is a colon `:`).](https://specifications.freedesktop.org/basedir/latest/#:~:text=mind%2E-,%24XDG%5FDATA%5FDIRS,colon%20%3A%29%2E)

but I wasn't sure we should apply that here - if the sound theme spec itself specifies the "in addtion to" variable, it sounds like they don't want you to look at the other thing. But now I'm coming around to the conclusion that it was an oversight, since there has to be some user-level directory for themes. So I'm implementing this.

> settings_registry.cpp:2716 hardcodes the empty-catalog fallback label "Default" (untranslated).
Consistent with the theme Name labels also being untranslated, so low value; no existing tr key. Left
it.

What should I do here? As the review says, the sound theme itself has untranslated `Name`:
```ini
[Sound Theme]
Name=Default
Directories=stereo

[stereo]
OutputProfile=stereo
```

> A theme file present as .disabled makes findThemeSound return nullopt, which setTheme then logs
as an error("... missing sound ...") rather than an intentional disable. Cosmetic log-level nit.

I don't expect that line is ever reachable since fallback to freedesktop should take care of it (there's comment above that line saying this too)

> This was the automated review, if you are confident this is the best order for theme discoveries I can merge as is. But it feels a little odd to me.

IMO theme discovery is fine, sound discovery i.e. `findThemeSound` feels a little like duplicating work (starting with a theme name and then searching all the directories for the theme, repeated for every sound - it might be better to discover the sounds and build theme structures in one pass). But $n$ is trivially small here and I feared overabstracting things. Also, once I do implement lookup in XDG_DATA_HOME, the current approach will allow simple overriding of individual sounds by just putting the overriding sound in XDG_DATA_HOME. If it is desired to implement __custom instead like the spec says then let me know

dchinmay2:
Implemented XDG_DATA_HOME lookup

ItsLemmy:
Thanks a lot, works great. Tweaked it a tiny bit.

PR #4493, opened 2026-09-19
fix(osd): restore keyboard backlight osd on upower 1.91.4

### LINKED ISSUE #4476: [BUG] Keyboard backlight indicator doesn't work on thinkpad
### Submission checklist

### Bug description

When I hit fn+space on my thinkpad t470s (toggle keyboard backlight) it doesn't show osd even tho keyboard backlight indicator turned on

### Steps to reproduce

1. Hit fn+space to turn on keyboard backlight on thinkpad laptops
2. No brightness indication even tho keyboard backlight osd turned on

### Expected behavior

Show keyboard backlight indicator on screen

### Actual behavior

No keyboard backlight indicator when turning keyboard backlight

### Logs / error output

```text

```

### Noctalia version

5.1.0

### Compositor

Niri

### Distribution

Arch-based

### Environment information

```text
Thinkpad t470s
I5 7300u
620 igpu
Keyboard backlight included
```

### Additional context

_No response_

AdityaKr015:
## Summary

This PR fixes the keyboard backlight OSD which uses upower 1.91.4, where the daemon stopped delivering `BrightnessChanged` for hardware-driven backlight changes.

`KeyboardBacklightService` now polls `GetBrightness` asynchronously every 100ms (200ms timeout) in addition to the existing signal subscription, and publishes through the existing change callback when the read value different from the tracked one. Device setup was deduplicated into `makeDevice()`.

## Motivation
upower 1.91.4 stopped delivering brightness-change events (its `brightness_hw_changed`
watch no longer works), so fn+space and other laptop shortcuts toggle the backlight without any D-Bus signal reaching clients. Verified:- A bus monitor captures zero `BrightnessChanged` emissions during keypresses on 1.91.4, while downgrading upower restores them (Lemmy verified on downgrade). Since the signal source is inside upowerd, the client now samples the value itself. The signal path remains the fast path on healthy daemons, and the tracked-value comparison prevents double OSDs.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Refactoring
- [ ] Build / packaging

## Related Issue

Closes #4476 

## Testing

- `just format && just test` all clean
- Tested in debug mode `pkill noctalia` then `./build-debug/noctalia`, the OSD now pops up correctly on fn+space, even on quick presses.
- I also tested build release and installing this branch and it works as before. 

## Manual Coverage

- [ ] Tested on Niri
- [x] Tested on Umbriel
- [ ] Tested on Hyprland
- [ ] Tested on Sway

## Screenshots / Videos

https://github.com/user-attachments/assets/9002d825-7deb-4723-bf8e-516e0d49e2ce

## Checklist

## Additional Notes

For Lemmy, can you check this PR on your 1.91.3 too?
I have the polls 100ms and 200ms timeout which seems good to me even in quick presses the backlight, it could go lower if wishes


ItsLemmy:
Thanks for the PR, but I won't merge it. I did not expect a polling based fix. This is clearly a  regression on uPower's watch, and they should fix it upstream.

This PR just mitigates the issue, and will be costly for people on battery. A 100ms per device poll means roughly ten D-Bus round trips per second, running forever on exactly the laptops that have a keyboard backlight, even when nobody touches it. The current code is purely signal driven and costs nothing at idle, so this would trade a temporary, version specific breakage for a permanent, universal wakeup source. That is not a trade I want to bake in.

The right fix is upstream: a broken hotkey signal is a first order uPower regression and others will hit it too. Could you open (or link) an issue against uPower so we can track it? Once they restore BrightnessChanged, the OSD works again with zero extra cost on our side.

Thanks again!

AdityaKr015:
No worries, Iknew this is a workaround and knew fix should be upstream. I will see in free time or if you can, then that would be better.

On another thought, its my fault i should have mention about the workaround, sorry. I will keep it running on my laptop till the fix land in upstream.

PR #4503, opened 2026-09-19
fix(niri): key workspaces by permanent id, not reorderable idx

### LINKED ISSUE #4381: [BUG] Workspace bar widget misbehaves when multiple workspaces share the same name
### Submission checklist

### Bug description

I use an utility similar to `i3-wsr` to dynamically rename workspaces based on their contents. This occasionally leads to multiple workspaces sharing the same display name. When this happens, the widget becomes confused.

Additionally, workspace ids displayed when label source is set to id are also confused and do not match ids reported by the compositor.

Image

Freshly created workspaces 2, 3, 4 all have the same name

Image

After switching from workspace 4 to workspace 2 all three are now labelled as "2".

### Steps to reproduce

1. Set multiple workspaces to the same name however that's done in your compositor
2. Enable workspace bar widget
3. ???
4. Observe the behaviour as you switch between workspaces

### Expected behavior

Workspace bar widget to correctly highlight the active workspace button regardless of it's name.

### Actual behavior

Image

The second workspace is active, yet both it and workspace three are highlighted on the widget.

Image

The third workspace is active, yet none of the workspaces are highlighted on the widget.

### Logs / error output

```text

```

### Noctalia version

noctalia v5.0.1

### Compositor

Niri

### Distribution

Arch-based

### Environment information

```text
Niri with this patch

https://github.com/niri-wm/niri/compare/niri-wm:niri:main...JohnDowson:niri:push-opwtpoxxktro
```

### Additional context

_No response_

dlucca1986:
## Summary

`NiriWorkspaceBackend::workspaceKey()` now prefers a workspace's permanent `id` over its per-output `idx` when building the key used to track the active workspace on the bar widget and to group windows by workspace.

## Motivation

`idx` is only a workspace's current position within its output; it shifts whenever a sibling workspace on the same output is created, closed, or reordered, and the same `idx` can be in use on multiple outputs at once. Keying on `idx` first meant two different (and possibly simultaneously live) workspaces could resolve to the same key, so the bar widget could highlight the wrong workspace as active or group a window's taskbar entry under the wrong workspace. `id` is niri's permanent, never-reused identifier for a workspace and is always populated by `parseWorkspace()`, so it does not have this collision.

## Type of Change

- [x] Bug fix

## Related Issue

Fixes #4381

## Testing

- `ninja -C build-debug niri_workspace_backend_test && ./build-debug/niri_workspace_backend_test`
- `meson test` (build-debug): 120/120 passed
- Added a regression test in `tests/niri_workspace_backend_test.cpp` that reproduces the collision: a `WorkspacesChanged` event assigns a new workspace (`id=300`) the `idx` that a just-closed workspace (formerly `id=1`) used to hold, and asserts `workspaceKeys()` reports it under `"300"`, not the stale `"1"`.

## Manual Coverage

- [ ] Tested on Niri
- [ ] Tested on Umbriel
- [ ] Tested on Hyprland
- [ ] Tested on Sway

I built and tested this on a remote build machine, not an interactive niri desktop session, so I could not visually confirm the bar's active-workspace highlight. The fix is instead validated by a regression test that replays the exact IPC event shape that causes the collision (`tests/niri_workspace_backend_test.cpp`) and asserts on `workspaceKeys()` directly, which is the same function the bar widget and taskbar grouping both consume.

## Checklist

## Additional Notes

This fix touches `workspaceKey()` only, which is used by both the bar widget's active-workspace tracking and by `appIdsByWorkspace()` / `workspaceWindows()` for taskbar window grouping, so both benefit from the same key stabilization.


ItsLemmy:
Thanks for this. The core change is solid: keying niri workspaces by the permanent id instead of the reorderable idx is the right call, it's self-consistent (workspaceKeys() and appIdsByWorkspace()/workspaceWindows() share the same workspaceKey()), and the added test genuinely guards it. Builds and lints clean for me.

My one concern is scope. This only feeds the taskbar's keys/grouping. The workspaces bar widget in #4381 doesn't use           
workspaceKey() at all: it reads the public Workspace.active/.id/.name (workspaces_widget.cpp:203,51-52), which come from the ext-workspace protocol (ext_workspace_backend.cpp), not this backend. So this PR doesn't touch the path behind the reported bug. I'd drop/adjust Fixes #4381 so it doesn't auto-close on merge; this is still worth landing as a taskbar-keying fix, with the real #4381 cause handled separately.

Minor nit: with id preferred first, the trailing idx branch in workspaceKey() is now unreachable and could be collapsed.

dlucca1986:
Thanks for the review and for catching the scope gap, that's a real miss on my part. Pushed: collapsed the now-unreachable idx branch, and reworded Summary/Motivation/Related Issue so this doesn't auto-close #4381 (it's a real taskbar-grouping fix on its own, just not the bar-widget bug from the original report). Will look at the real ext-workspace/widget path for #4381 separately.

